PR #2, opened 2016-04-19
"Tracker #117191625: Spike: BOSH Deployment (bosh-lite)"

cdlliuy:
All integration test passed using ./bin/getStart.sh" in my side, so ask for merge to dev branch


fraenkel:
I am not going to pull this in quite yet.

There are too many conflicts with some of the other work that has been pulled.
Much of what you did is correct, but some of it is not.
As one big commit, it's hard to separate what is right from what is wrong in this case.
I will leave this around for when we get to the actual story to do a BOSH Release. This will be a helpful start and we can focus on the remaining pieces.

cdlliuy:
Got it.  I will modify the pull request (or create a new one) to include the commit for "bosh-release" folder only, and didn't change the work flow in "/bin" . Then it will be safe for merge

fraenkel:
That would be a better split but I don't want to include things that we aren't using yet. This would belong in the autoscaler-release.
The other issues I mentioned but didn't go into detail would prevent BOSH from really working. At which point all we have is code sitting there getting stale.

The BOSH work will be coming soon. There are just a few more important things we need to hammer out before we start actively working on a BOSH release.

cdlliuy:
OK, got it, I will push this part of work to autoscaler-release only to save the efforts, and pick it up later when we are ready for bosh-release

PR #4, opened 2016-04-19
fix defect #117837333

boyang9527:
Return 500 internal server error and log with the right message when CloudFoundryManager throws exception due to login failed.


fraenkel:
@boyang9527 This doesn't seem like the right way to fix this. We are fixing the symptom and not the real issue. Why did we get the NPE instead of an Exception that we would have handle which would identify the root cause.

Also for changes like this, we need to add a test case so we don't regress on functions like this.

boyang9527:
I will close this pull request and  have a new one

PR #5, opened 2016-04-19
API naming change #116994255

boyang9527:
- change web.xml filter and servlet mapping
- change JAX-RSpath annotation
- change test case
- change test script
- change API doc


boyang9527:
@fraenkel ,  this change will  break your integration test, you may change accordingly with the new API path

fraenkel:
@boyang9527 Thanks for the heads up.  I have to run units and integration for every merge.

fraenkel:
Merged.

PR #6, opened 2016-04-20
Dependency currency #116995795

qibobo:
[#116995795]


fraenkel:
@qibobo There should only be one commit which is the dependency upgrade. The upgrade from 1.7 to 1.8 is already in so it should not be here. And I don't know why you would have a merge commit if this was rebased on top of develop.

fraenkel:
It would also be good if we pulled out common dependencies across our poms into a common parent.
See https://books.sonatype.com/mvnex-book/reference/optimizing-sect-dependencies.html

PR #9, opened 2016-04-22
Dev yingpersonal

cdlliuy:
The pull request contains fix of 2 bugs, since both of the changes are small.

The changes in readme indicates the cf version must be 6.17.0+, so that there is no "Getting OAuth Token ...  OK " in the cmd "cf oauth" output.   When running with cf 6.15.0 in my env, I always got acceptance test failure due the "Authorization" header is wrong.

### DIFF
--- a/server/src/main/java/org/cloudfoundry/autoscaler/manager/ApplicationScaleManagerImpl.java
+++ b/server/src/main/java/org/cloudfoundry/autoscaler/manager/ApplicationScaleManagerImpl.java
@@ -286,34 +286,32 @@ public class ApplicationScaleManagerImpl implements ApplicationScaleManager{
 	private boolean validateInstanceCounts(String appId,AutoScalerPolicy policy,String triggerId, Integer currentInstanceCount) throws Exception{
 		int minCount = policy.getCurrentInstanceMinCount();
 		int maxCount = policy.getCurrentInstanceMaxCount();
+		boolean validScaling = false;
 		if(null == triggerId){
-			if (currentInstanceCount >= minCount && currentInstanceCount <=maxCount) {
-				logger.info("The number of application instances is greater than minimum instances limitation and less than maximum instances limitation.");
-				return false;
-			} else if(currentInstanceCount >= maxCount){
-				logger.info("The number of application instances is greater than maximum instances limitation. Start scaling in.");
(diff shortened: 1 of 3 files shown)


fraenkel:
Please don't combine things like this since they have nothing to do with each other.
Right now I cannot pull either since I have issues with the validateInstanceCounts.
I have added comments to the commit.

server/src/main/java/org/cloudfoundry/autoscaler/manager/ApplicationScaleManagerImpl.java
```diff
@@ -286,34 +286,32 @@ private boolean validateCooldownSetting(String appId,AutoScalerPolicyTrigger pol
...
+			} 
 		}
 		else{
-			/** Check minimum count and maximum count **/
+			/** Trigger by dynamic scale. Check minimum count and maximum count **/
 			if (AutoScalerPolicyTrigger.TriggerId_LowerThreshold.equals(triggerId)
 					&& currentInstanceCount <= minCount) {
-				logger.debug("False: Min count reached. No scaling in action.");
-				return false;
+				logger.debug(String.format("Invalid scaling event %s for app %s, since the Min instance count reached",
```
fraenkel: This is definitely an improvement.
I haven't looked to see if the triggerId is something that one understands. I am not going to worry about it now but will handle the broader concern when we start improving our logging.

src/acceptance/README.md
```diff
@@ -16,7 +16,7 @@ and should not modify the CF state in such a way as to impact other tests.
 
 Set up your golang development environment, per [golang.org](http://golang.org/doc/install).
 
-See [cf CLI](https://github.com/cloudfoundry/cli) for instructions on installing the go version of `cf`.
+See [cf CLI](https://github.com/cloudfoundry/cli) for instructions on installing the go version of `cf`. Recommend version v6.17.0+
```
fraenkel: I don't want to put the version in because it will change and it does depend on the version of CF you have deployed.
The CF APIs we leverage do not require such a high release.

cdlliuy:
HI, michael.  I spitted these items into 2 separate pull request  already, please following with pull request #10 and #11. Close this one.

PR #15, opened 2016-04-29
validate token first before check user's permissions with the space

boyang9527:
[#117835821]

- New method in AuthenticationTool to get user ID from token
- refactor AuthenticationTool class
- validate token first, then validate user's permission on the space that app is in
- reuse compiled pattern in filter


fraenkel:
Merged

PR #17, opened 2016-05-03
fail the service broker if db is misconfigured

boyang9527:
[#118493209]

- throw exception when db can not be initialized
- throw runtime exception i LifecycleListener
- fix issues due to changes in couchdbstoreservice
- remove http protocol setting from service broker

### DIFF
--- a/servicebroker/src/main/java/org/cloudfoundry/autoscaler/servicebroker/LifecycleListener.java
+++ b/servicebroker/src/main/java/org/cloudfoundry/autoscaler/servicebroker/LifecycleListener.java
@@ -22,17 +22,16 @@ public class LifecycleListener implements ServletContextListener {
     }
 
     /**
-     * Load registered triggers from mongodb store
      * 
      * @see ServletContextListener#contextInitialized(ServletContextEvent)
      */
-    public void contextInitialized(ServletContextEvent event) {
-        try {
-        	DataSourceUtil.setStoreProvider(Constants.CONFIG_ENTRY_DATASTORE_PROVIDER_COUCHDB);
-        	ScalingServiceMgr.getInstance();
-        } catch (Exception e) {
-            logger.error(e.getMessage(), e);
-        }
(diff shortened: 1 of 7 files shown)


servicebroker/src/main/java/org/cloudfoundry/autoscaler/servicebroker/LifecycleListener.java
```diff
@@ -22,17 +22,16 @@ public LifecycleListener() {
...
-    public void contextInitialized(ServletContextEvent event) {
-        try {
-        	DataSourceUtil.setStoreProvider(Constants.CONFIG_ENTRY_DATASTORE_PROVIDER_COUCHDB);
-        	ScalingServiceMgr.getInstance();
-        } catch (Exception e) {
-            logger.error(e.getMessage(), e);
-        }
+    public void contextInitialized(ServletContextEvent event)  {
+        	try {
+				DataSourceUtil.setStoreProvider(Constants.CONFIG_ENTRY_DATASTORE_PROVIDER_COUCHDB);
```
fraenkel: Formatting.

servicebroker/src/main/resources/config.properties
```diff
@@ -5,7 +5,7 @@ brokerPassword=${brokerPassword}
 # Scaling server URLs and protocol settings
 serverURIList=${serverURIList}
 apiServerURI=${apiServerURI}
-httpProtocol=${httpProtocol}
+#httpProtocol=${httpProtocol}
```
fraenkel: If its being removed, why is it commented out instead?

boyang9527: This needs a complete check of all projects and further changes in other places.  It was commented to make sure it will not generate wrong setting when doing filtering. 

The complete remove is done through a new PR
https://github.com/cloudfoundry-incubator/app-autoscaler/pull/19

fraenkel:
Merged

PR #18, opened 2016-05-04
better error message from filter of API server

qibobo:
[#118614105]

### DIFF
--- a/api/src/main/java/org/cloudfoundry/autoscaler/api/filter/AuthenticationFilter.java
+++ b/api/src/main/java/org/cloudfoundry/autoscaler/api/filter/AuthenticationFilter.java
@@ -16,158 +16,148 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
 import org.apache.log4j.Logger;
-import org.cloudfoundry.autoscaler.api.util.MessageUtil;
-import org.cloudfoundry.autoscaler.api.util.RestApiResponseHandler;
 import org.cloudfoundry.autoscaler.common.AppInfoNotFoundException;
 import org.cloudfoundry.autoscaler.common.AppNotFoundException;
 import org.cloudfoundry.autoscaler.common.ClientIDLoginFailedException;
 import org.cloudfoundry.autoscaler.common.util.AuthenticationTool;
 import org.cloudfoundry.autoscaler.common.util.AuthenticationTool.SecurityCheckStatus;
 import org.cloudfoundry.autoscaler.common.util.CloudFoundryManager;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/cloudfoundry/autoscaler/api/filter/AuthenticationFilter.java
```diff
@@ -55,7 +56,7 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
             String authorization = httpReq.getHeader("Authorization");
             if ( authorization == null){
             	logger.info("No Authorization header found.");
-                httpResp.sendError(HttpServletResponse.SC_UNAUTHORIZED);
+                this.setResponse(httpResp, HttpServletResponse.SC_UNAUTHORIZED, MessageUtil.UNAUTHORIZED);
```
fraenkel: This is not how MessageUtil works.
If you are going to add keys to MessageUtil, there should be a corresponding message added to the English NLS resource bundle.
You should be using the MessageUtil.getMessageXXXX method to load the appropriate message.

api/src/main/java/org/cloudfoundry/autoscaler/api/filter/AuthenticationFilter.java
```diff
@@ -169,5 +171,14 @@ public void init(FilterConfig config) throws ServletException {
 			}
 
 	    } 
+	private void setResponse(HttpServletResponse httpResp, int status, String msg) throws IOException{
```
fraenkel: Formatting looks wrong.

api/src/main/java/org/cloudfoundry/autoscaler/api/util/MessageUtil.java
```diff
@@ -58,6 +58,8 @@
 	public final static String RestResponseErrorMsg_get_history_in_Get_History_context = "RestResponseErrorMsg_get_history_in_Get_History_context";
 	public final static String RestResponseErrorMsg_get_metric_in_Get_Metric_context = "RestResponseErrorMsg_get_metric_in_Get_Metric_context";
 	public final static String RestResponseErrorMsg_API_Server_retrieve_service_information_context = "RestResponseErrorMsg_API_Server_retrieve_service_information_context";
+	public final static String UNAUTHORIZED = "unauthorized";
+    public final static String INTERNAL_SERVER_ERROR = "internal_server_error";
```
fraenkel: Formatting is wrong.

qibobo:
@fraenkel Updated already.(1) Add message to NLS. (2) Format the code.

fraenkel:
See tracker item for my comments.

PR #19, opened 2016-05-05
remove default http protocol setting

boyang9527:
- remove httpProtocol setting from property files
- use protocol in URI and default to http if there is no protocol specified in scaling server or API server URIs
- refine README.md

[#118973159]


fraenkel:
See tracker comments.

fraenkel:
Merged

PR #22, opened 2016-05-06
Allow local logging for applications

qibobo:
[#118807555]


fraenkel:
I think I will take this and modify it to use the RollingFileAppender instead. I would rather guard against blowing up the disk limits.

fraenkel:
Merged

PR #24, opened 2016-05-10
Change  settings to speed up UAT

boyang9527:
[#117948543]

- create a new profile for UAT
- change default cf default timeout from 30 seconds to 60 seconds
- change settings in dynamic.json


fraenkel:
There are a couple of items I will fix with a follow up commit.
1. Adjust acceptance tests timings based on the reportInterval. It needs to reflect what you deployed with.
2. Add an additional item to both READMEs to cover this new uat.profile and the new integration property.

PR #29, opened 2016-05-20
story119322761: Integration tests need to inject statWindow and breachDuration

kanekoh:
Add the generator creating a policy from a template.
The below parameters also add for this generator. Because, it's convenience to test for different combinations.
- startTimeValue
- startDateValue
- endTimeValue
- maxCount

### DIFF
--- a/src/acceptance/template/template.go
+++ b/src/acceptance/template/template.go
@@ -6,8 +6,10 @@ import (
 	"strconv"
 )
 
+type TemplateValues map[string]string
+
 // Parameter names to replace them to a value.
-var (
+const (
 	EndTimeValue   = "{endTimeValue}"
 	MaxCount       = "{maxCount}"
 	StartDateValue = "{startDateValue}"
@@ -15,7 +17,7 @@ var (
 	ReportInterval = "{reportInterval}"
 )
 
-var defaultValuesForPolicy = map[string]string{
+var defaultValuesForPolicy = TemplateValues{
 	MaxCount:       "5",
 	ReportInterval: "120",
 	StartTimeValue: "00:00",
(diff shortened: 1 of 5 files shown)


src/acceptance/template/template.go
```diff
@@ -0,0 +1,74 @@
+package template
+
+import (
+	"bytes"
+	"io/ioutil"
+	"strconv"
+)
+
+// Parameter names to replace them to a value.
+var (
```
fraenkel: Why aren't these consts?

src/acceptance/template/template.go
```diff
@@ -0,0 +1,74 @@
...
+			merge_values[k] = value
+		} else {
+			merge_values[k] = default_v
+		}
+	}
+
+	return merge_values
+}
+
+func setValue(key string, value string, values map[string]string) map[string]string {
```
fraenkel: Why not just create a type called TemplateValues which is
type TemplateValues map[string]string

Then have 2 funcs on it which are SetString and SetInt.

src/acceptance/template/template.go
```diff
@@ -56,19 +58,10 @@ func mergeValuesAndDefaultValues(values map[string]string) map[string]string {
...
-	if values == nil {
-		values = map[string]string{}
-	}
-	values[key] = value
-	return values
-}
-func SetStringValue(key string, value string, values map[string]string) map[string]string {
-	new_values := setValue(key, value, values)
-	return new_values
+func (values *TemplateValues) SetString(key string, value string) {
```
fraenkel: These shouldn't be pointer types since it is a map underneath.

PR #32, opened 2016-05-27
31-fix-date-parsing Use a date format that matches the observed time.

ericpromislow:
This is needed to prevent the autoscaler server from throwing NullPointerExceptions

### DIFF
--- a/server/src/main/java/org/cloudfoundry/autoscaler/metric/bean/CFInstanceStats.java
+++ b/server/src/main/java/org/cloudfoundry/autoscaler/metric/bean/CFInstanceStats.java
@@ -87,7 +87,7 @@ public class CFInstanceStats {
 
     private static Date parseDate(String date) {
         try {
-            return new SimpleDateFormat("yyyy-MM-dd HH:mm:ss ZZZZZ").parse(date);
+            return new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS").parse(date);
         } catch (ParseException e) {
         }
         return null;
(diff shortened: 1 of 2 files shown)


fraenkel:
I am a bit confused how you are seeing a different timestamp.
If I issue the stats endpoint, I get the following back:
"usage": {
            "time": "2016-05-31 14:47:49 +0000",
            "cpu": 0.004092637867367473,
            "mem": 49086464,
            "disk": 0
         }

The code does `"time" => Time.now.to_s` which corresponds to the above and ruby docs.

ericpromislow:
It's possible that this is only an error in the cfibmers repository. I was observing timestamps of the form 
"2016-05-31T14:47:49.123456789Z..." that didn't match the format. (And I can't be more specific on the part after the "Z" -- I'm working on a different branch and had to zap the VMs with autoscaling.)  So I'll close this PR: no need to update on this repo.

Are you taking PRs for the cfibmers repository or will we need to fork it until app-autoscaling is ready for general use?

fraenkel:
This error shouldn't be isolated to that repo. It is the same code.
Are you using DEA or Diego?
This repo is ahead of cfibmers, but does not have BOSH support.

All issues should be reported here.

ericpromislow:
We're using diego.

fraenkel:
That might explain the difference. Let me check out diego.

fraenkel:
@ericpromislow Diego is serializing a time.Time which will be in RFC 3339 format with sub-second precision. To support Diego, this is needed but I would like to have tests that show both the Ruby format and the RFC 3339 format are both properly handled.

ericpromislow:
I've working on a patch to CFInstanceStats to handle both kinds of dates. There's no way to configure this -- the autoscaler is a singleton that services both diego apps and DEA apps at the same time. Is there a directory in app-autoscaler/server/src/test I should put these tests in, or would new tests be needed?

fraenkel:
I would just create a new file named server/src/test/java/org/cloudfoundry/autoscaler/metric/bean/CFInstanceStatsTest.java.

Don't worry about backfilling, just focus on the new function you are adding.

satoyan:
When will this PR be merged?

fraenkel:
I am going to close this since it no longer applies to the new code base.
If you disagree, just reopen it with comments.

PR #35, opened 2016-06-13
Introduce DB app to auto-scaler deployment

pradyutsarma:


fraenkel:
1. Can you squash the commits.
2. Why do we require java-buildpack v3.7?

pradyutsarma:
@fraenkel Will it be possible for you to use the option of "Squash and Merge" from the Github UI when you merge the PR?

fraenkel:
I can do that but I would rather folks deliver commits properly. I don't mind multiple commits if they are for separate functions but these 3 should be delivered as one.
The moment I squash, it will not reflect that it came from you.

fraenkel:
Manually merged.

pradyutsarma:
@fraenkel I see that you've squashed and merged the commit. Thanks !!

Regarding your query regarding using buildpack version 3.7, See this thread here https://github.com/spring-projects/spring-boot/issues/4897 for more details on it.
We're using spring-boot 1.4.0 for the db app and hence need buildpack 3.7.

PR #36, opened 2016-06-13
Metrics Collector: Retrieve memory metric for an application

boyang9527:


fraenkel:
Rather than force app-autoscaler under a GOPATH, metrics-collector should move under app-autoscaler/src which makes the root of autoscaler the GOPATH.

metrics-collector/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### System requirements:
+
+* Go 1.5 or above
+* Cloud Foundry relese 235 or later
+
+### Build and test
+
+1. Create a directory where you would like to store the source for Go projects and their binaries 
+2. Add this directory to your `$GOPAHT`
```
fraenkel: Spelling

metrics-collector/README.md
```diff
@@ -0,0 +1,74 @@
...
+3. create the directory `src/github.com/cloudfoundry-incubator` 
+3. change directory to `src/github.com/cloudfoundry-incubator`
+4. clone the `app-autoscaler` project: `clone git@github.com:cloudfoundry-incubator/app-autoscaler.git`
```
fraenkel: git URLs only make sense for developers, we should put in https urls.
developers should have the following in their .gitconfig to do automatic rewrite
[url "git@github.com:"]
  pushInsteadOf = https://github.com/

metrics-collector/README.md
```diff
@@ -0,0 +1,74 @@
...
+5. change directory to 'src/github.com/cloud-foundry-incubator/app-autoscaler/metrics-collector'
+6. build the project: `go build -o out/mc`
+7. test the project: `ginkgo -r`
```
fraenkel: This should also have -race. I will investigate if we can do parallel as well.

metrics-collector/README.md
```diff
@@ -0,0 +1,74 @@
...
+cf:
+  api: "https://api.bosh-lite.com"
+  grant_type: "password"
+  user: "admin"
+  pass: "admin"
+server:
+  port: 8080
+logging:
+  level: "info"
+  file: "logs/mc.log"
```
fraenkel: Normally we only log to stdout. We have no log file rotation in place.

PR #37, opened 2016-06-14
Add api application to the autoscaler repo

pradyutsarma:
1. api application added to the autoscaler repo
2. Added a manifest.yml to the repo to describe the deployment to cloudfoundry

The api application expects the DB_URI in the env (See manifest.yml) and not as a DB service instance. This enables the app to be deployed or used outside of CF (could be set via a bosh job tomorrow if and when we migrate to a bosh release) as long as the env variable is set.


fraenkel:
I have added comments in the story above.

api/lib/models/index.js
```diff
@@ -0,0 +1,32 @@
...
+var Sequelize = require('sequelize');
+var logger = require('../log/logger');
+var sequelizeURL = process.env.DB_URI;
+var sequelize = new Sequelize(sequelizeURL);
+sequelize.authenticate()
+  .then(function() {
+    logger.info('DB Connection has been established successfully');
+  })
+.catch(function(error) {
+  logger.error('DB Connection failed ',error);
```
fraenkel: Why do we continue if we failed authentication?

api/lib/models/index.js
```diff
@@ -0,0 +1,32 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var Sequelize = require('sequelize');
+var logger = require('../log/logger');
+var sequelizeURL = process.env.DB_URI;
+var sequelize = new Sequelize(sequelizeURL);
```
fraenkel: No connection pooling?

api/test/unit/index.test.js
```diff
@@ -0,0 +1,13 @@
...
+
+//var expect = require('expect.js');
+var expect = require("chai").expect;
+var logger = require('../../lib/log/logger');
+
+describe('Model Initialization', function () {
+  it('returns the Policy model', function () {
+    var models = require('../../lib/models');
+    logger.log('Policy model loaded successfully');
+    expect(models.policy_json).to.be.ok;
```
fraenkel: What about sequelize and Sequelize?

api/test/unit/index.test.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
```
fraenkel: Can this be placed in a directory that matches the source tree so we know which index.js you are testing?
E.g., /test/unit/models/index.test.js

api/test/unit/policy.test.js
```diff
@@ -0,0 +1,87 @@
+'use strict';
```
fraenkel: This doesn't test policy but policies.js

Move to /test/unit/routes/policies.test.js

api/lib/models/index.js
pradyutsarma: Sequelize uses a default connection pool which we don't override as of now. We can look into this if required.

api/test/unit/policy.test.js
```diff
@@ -0,0 +1,87 @@
...
+
+  describe('create a Policy', function () {
+    it('Success', function () {
+        return this.Policy.create({  'policy_json':dummyPolicyJson, 'app_id': '12349' }).then(function (policy) {
+        	logger.info('Policy created for app_id 12349');
+        	expect(policy.policy_json).to.equal(dummyPolicyJson);
+          
+        });
+    });
+    it('Failure', function () {
```
fraenkel: What happens if app_id or the policy_json is missing?

api/lib/models/index.js
fraenkel: It sets the pool to true but doesn't set up any values for the pool size, timeout, etc... I will add a story to track this.

fraenkel:
I will merge what you have but still have additional comments on the code/tests.

api/test/unit/routes/policies.test.js
```diff
@@ -0,0 +1,92 @@
...
+        }
+        expect(result.statusCode).to.equal(400);
+        done();
+      });
+    });
+  });
+
+  after(function() {
+    this.Policy.truncate().then(function(result) {
+      logger.warn('Truncating the test db before test run !!!');
```
fraenkel: Shouldn't this be in a before block?

api/test/unit/routes/policies.test.js
```diff
@@ -0,0 +1,92 @@
...
+      logger.info('creating the table structure');
+    });
+  });
+  describe('Policy Creation', function() {
+    it('should create a policy for app id 12345', function(done) {
+      request(app)
+      .put('/v1/apps/12345/policy')
+      .send(fakePolicy)
+      .end(function(error,result) {
+        if (error) {
```
fraenkel: If check seems unnecessary if you don't expect an error.
expect(error).to.not.be.null;

api/test/unit/routes/policies.test.js
```diff
@@ -0,0 +1,92 @@
...
+          expect(error).to.not.be.null;
+          throw error;
+        }
+        expect(result.statusCode).to.equal(200);
```
fraenkel: What about the body? If we are sending data, shouldn't we validate it?

api/test/unit/routes/policies.test.js
```diff
@@ -0,0 +1,92 @@
...
+    it('should fail to create another policy for app id 12345', function(done) {
+      request(app)
+      .put('/v1/apps/12345/policy')
+      .send(fakePolicy)
+      .end(function(error,result) {
+        if (error) {
+          expect(error).to.not.be.null;
+          throw error;
+        }
+        expect(result.statusCode).to.equal(400);
```
fraenkel: See comment regarding body validation.

api/test/unit/models/index.failure.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('Test Model Load Failures', function () {
+  before(function() {
+	  correctDBUri = process.env.DB_URI;
+	  process.env.DB_URI = 'sqlite://test:test/db.test.sqlite';
+  });
+	
+  it('Should fail for a invalid DB_URI', function () {
+    var models = require('../../../lib/models')(function (error) {
+    	logger.error ('Error while establishing connection', error);
```
fraenkel: Why are we logging an error in tests?

api/test/unit/models/index.test.js
```diff
@@ -1,12 +1,11 @@
...
 var expect = require("chai").expect;
-var logger = require('../../lib/log/logger');
+var logger = require('../../../lib/log/logger');
 
 describe('Model Initialization', function () {
   it('returns the Policy model', function () {
-    var models = require('../../lib/models');
+    var models = require('../../../lib/models')();
     logger.log('Policy model loaded successfully');
     expect(models.policy_json).to.be.ok;
```
fraenkel: This only tests what you expect, not what may actually be there.
Don't we also expect sequelize which also means there should be only 2 items in the map?

I could add more models and never change this test which would always pass.

api/test/unit/models/policy.test.js
```diff
@@ -0,0 +1,84 @@
...
+			"instance_max_count" : 40
+		} ]
+	}
+};
+
+describe('Cleaning up test database', function() {
+  before(function() {
+    this.Policy = require('../../../lib/models')().policy_json;
+    this.Policy.sequelize.sync().then(function() {
+      logger.info('creating the table structure');
```
fraenkel: There is no need to do any logging in our tests.

pradyutsarma:
Closing this PR. Will submit the code incorporating the code review comments in the next PR. Code belonging to this PR already merged via https://github.com/cloudfoundry-incubator/app-autoscaler/commit/8736e0d6b4674148b6732338272dbe511565a9a4

PR #38, opened 2016-06-16
retrieve the catalog from the broker

qibobo:
[#119505867]

### DIFF
--- a/servicebroker/lib/index.js
+++ b/servicebroker/lib/index.js
@@ -44,5 +44,8 @@ router.get('/catalog', function(req, res) {
 });
 
 app.use('/v2', router);
-var server = app.listen(port);
+var server = app.listen(port, function () {
+    var port = server.address().port;
+    console.log('Service broker app is running and listening at port %s', port);
+});
 module.exports = server;
--- a/servicebroker/lib/index.js
+++ b/servicebroker/lib/index.js
@@ -16,9 +16,8 @@ app.use(bodyParser.urlencoded({ extended: true }));
 app.use(bodyParser.json());
 
 var auth = function (req, res, next) {
-  
   function unauthorized(res) {
-    res.set('WWW-Authenticate', 'Basic realm=Authorization Required');
+    res.set('WWW-Authenticate', 'Basic realm="serviceBrokerAuth"');
     return res.sendStatus(401);
   };
 
--- a/servicebroker/test/api/api_test.js
+++ b/servicebroker/test/api/api_test.js
@@ -12,14 +12,7 @@ describe('Unit Test for API', function() {
 
   it('getCatalog() should return catalog info with JSON format', function() {
     var data = api.getCatalog();
-    data.should.have.property('services');
-    JSON.stringify(data).length.should.equal(JSON.stringify(catalog).length);
-    data.services[0].id.should.equal(catalog.services[0].id);
-    data.services[0].name.should.equal(catalog.services[0].name);
-    data.services[0].description.should.equal(catalog.services[0].description);
-    data.services[0].plans[0].id.should.equal(catalog.services[0].plans[0].id);
-    data.services[0].plans[0].name.should.equal(catalog.services[0].plans[0].name);
-    data.services[0].plans[0].description.should.equal(catalog.services[0].plans[0].description);
+    JSON.stringify(data).should.equal(JSON.stringify(catalog));
   });
 
 });
--- a/servicebroker/test/index_test.js
+++ b/servicebroker/test/index_test.js
@@ -28,9 +28,8 @@ describe("RESTful API test suite", function() {
       .get("/v2/catalog")
       .set("Authorization", "Basic " + auth)
       .expect("Content-type", /json/)
-      .expect(200) // THis is HTTP response
+      .expect(200)
       .end(function(err, res) {
-        // HTTP status should be 200
         res.status.should.equal(200);
         JSON.stringify(res.body).should.equal(JSON.stringify(catalog));
         done();
@@ -54,25 +53,19 @@ describe("RESTful API test suite", function() {
       .expect(401, done);
   });
 
-  it("should return 401 when incorrect auth format provided", function(done) {
-    supertest(server)
-      .get("/v2/catalog")
-      .set("Authorization", "invalidauth"  )
-      .expect(401, done);
-  });
-
-  it("should return 401 when auth info is not encoded with base64", function(done) {
-    supertest(server)
-      .get("/v2/catalog")
-      .set("Authorization", "Basic " + settings.user + ":" + settings.password )
-      .expect(401, done);
-  });
-
   it("should return 401 when incorrect user/password provided", function(done) {
     supertest(server)
       .get("/v2/catalog")
       .set("Authorization", "Basic " + new Buffer("incorrectuser:incorrectpassword").toString('base64') )
       .expect(401, done);
+  });
+
+  it("should return 401 when incorrect user provided", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .set("Authorization", "Basic " + new Buffer("incorrectuser" + ":" + settings.password).toString('base64') )
+      .expect(401, done);
+
   });
 
   it("should return 401 when incorrect password provided", function(done) {
@@ -81,13 +74,6 @@ describe("RESTful API test suite", function() {
       .set("Authorization", "Basic " + new Buffer(settings.user+ ":" + "incorrectpassword").toString('base64') )
       .expect(401, done);
 
-  });
-
-  it("should return 401 when incorrect basic auth format provided", function(done) {
-    supertest(server)
-      .get("/v2/catalog")
-      .set("Authorization", "Basic " + new Buffer(settings.user + settings.password).toString('base64') )
-      .expect(401, done);
   });
 
 
--- a/servicebroker/test/index_test.js
+++ b/servicebroker/test/index_test.js
@@ -32,13 +32,7 @@ describe("RESTful API test suite", function() {
       .end(function(err, res) {
         // HTTP status should be 200
         res.status.should.equal(200);
-        JSON.stringify(res.body).length.should.equal(JSON.stringify(catalog).length);
-        res.body.services[0].id.should.equal(catalog.services[0].id);
-        res.body.services[0].name.should.equal(catalog.services[0].name);
-        res.body.services[0].description.should.equal(catalog.services[0].description);
-        res.body.services[0].plans[0].id.should.equal(catalog.services[0].plans[0].id);
-        res.body.services[0].plans[0].name.should.equal(catalog.services[0].plans[0].name);
-        res.body.services[0].plans[0].description.should.equal(catalog.services[0].plans[0].description);
+        JSON.stringify(res.body).should.equal(JSON.stringify(catalog));
         done();
       });
   });
--- a/servicebroker/README.md
+++ b/servicebroker/README.md
@@ -9,9 +9,10 @@ AutoScaler service broker is built by nodejs/expressjs
 npm bootstrap
 ```
 **Service broker configuration**
+
 Change configuration for service broker in file `config/settings.json`, including the following items:
-1. `port`, the port number in which the service broker app is launched and listening to.  The default value is 8080. 
-2. `user` and `password` : HTTP basic authorization is enabled according to the [CF service broker API # authorization](http://docs.cloudfoundry.org/services/api.html#authentication).  Please specific the user/password in setting.json, and use the same user/password when issue command "cf create-service-broker"
+* `port`, the port number in which the service broker app is launched and listening to.  The default value is 8080. 
+* `user` and `password` : HTTP basic authorization is enabled according to the [CF service broker API # authorization](http://docs.cloudfoundry.org/services/api.html#authentication).  Please specific the user/password in setting.json, and use the same user/password when issue command "cf create-service-broker"
 
 
 **Run eslint**
--- a/servicebroker/README.md
+++ b/servicebroker/README.md
@@ -9,7 +9,10 @@ AutoScaler service broker is built by nodejs/expressjs
 npm bootstrap
 ```
 **Service broker configuration**
-Change configuration for service broker in file `config/settings.json`
+Change configuration for service broker in file `config/settings.json`, including the following items:
+1. `port`, the port number in which the service broker app is launched and listening to.  The default value is 8080. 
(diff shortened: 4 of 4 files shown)


servicebroker_node/config/settings.json
```diff
@@ -0,0 +1,11 @@
+{
+	"port":8080, 
+	
+	"service_id": "autoscaler-guid",
```
fraenkel: What is this stuff for?

servicebroker_node/lib/api/api.js
```diff
@@ -0,0 +1,24 @@
+module.exports = function() {
+
+  var fs = require('fs');
+  var path = require('path');
+
+  var settings = JSON.parse(
```
fraenkel: We need to assume the catalog.json and any config are provided via external files.
This needs to be injected rather than hardcoded.
We can create a separate chore if you wish.

servicebroker_node/test/ut/api/api_test.js
```diff
@@ -0,0 +1,20 @@
...
+describe('Unit Test for API', function() {
+
+it('getCatalog() should return catalog info with JSON format', function() {
+  var data = api.getCatalog();
+  data.services[0].id.should.equal(settings.service_id);
+  data.services[0].name.should.equal(settings.service_name);
+  data.services[0].description.should.equal(settings.service_description);
+  data.services[0].plans[0].id.should.equal(settings.plan_id);
+  data.services[0].plans[0].name.should.equal(settings.plan_name);
+  data.services[0].plans[0].description.should.equal(settings.plan_description);
```
fraenkel: how do we know that is all that is there?

servicebroker_node/test/fat/index_test.js
```diff
@@ -0,0 +1,34 @@
+var supertest = require("supertest");
```
fraenkel: This is still a unit test. Just move it over.

servicebroker_node/test/fat/index_test.js
```diff
@@ -0,0 +1,34 @@
...
+var should = require("should");
+var fs = require('fs');
+var path = require('path');
+
+// This agent refers to PORT where program is runninng.
+var settings = JSON.parse(fs.readFileSync( path.join (__dirname, '../../config/settings.json'), 'utf8'));
+var port = process.env.PORT || settings.port;
+var server = supertest.agent("http://localhost:"+port);
```
fraenkel: We should be starting and stopping the server as needed for our testing.
See https://glebbahmutov.com/blog/how-to-correctly-unit-test-express-server/ as an example.

servicebroker_node/test/fat/index_test.js
```diff
@@ -0,0 +1,34 @@
...
+
+// UNIT test begin
+
+describe("RESTful API test suite",function(){
+
```
fraenkel: What happens with an invalid path?

fraenkel:
You can also just put it in the servicebroker directory. Doesn't have to be servicebroker_node.

servicebroker_node/config/settings.json
cdlliuy: we extracted the configurable items in catalog here, so that it is easy to edit the key information of catalog with key-value format instead of taking care of the whole catalog JSON format

servicebroker_node/lib/api/api.js
cdlliuy: ok, we did hardcoded the service details and plan details with assumption that we only have 1 plan and 1 service id here.  Maybe back to prior JSON stream approach (save the catalog information in a JSON, and then read it directly) is better.

servicebroker_node/config/settings.json
cdlliuy: it used to configure the service information and plan information which may different per deployment. i.e. using different service name, service guid and etc.

servicebroker_node/test/ut/api/api_test.js
cdlliuy: yeah.. it is hardcoded the service details and plan details with assumption that we only have 1 plan and 1 service id here.

servicebroker_node/test/fat/index_test.js
cdlliuy: in my understanding,  if we invoke a function directly and verify the input/output , that is unit test.  If we launch the server (here, service broker app) , and invoke the RESTful API to verify the request/response, it is function test. if more apps are launched together, it is integration test.   So, here, we start service broker app , and call /v2/catalog endpoint, it should be a function test.  Is that right?

servicebroker_node/test/fat/index_test.js
cdlliuy: do you mean invalid file path in "require" or the invalid http endpoint?  if the file path for "settings" are invalid, the test case should fail.

servicebroker_node/lib/api/api.js
fraenkel: We need to pick one. If it is going to be static, we might has well hardcode it in a file which is read once and always returned. It is a bit of overkill to regenerate the map each time getCatalog is called.

servicebroker_node/test/ut/api/api_test.js
fraenkel: We should test lengths to guarantee there is no more expected data.

servicebroker_node/test/fat/index_test.js
fraenkel: You have units which test everything.
Breaking things down provides no additional value other than unnecessary setup that should be done within the tests anyhow.

Integration tests are exactly that. They test the integration between components which are outside the scope of a single project. It would be similar to where acceptance tests live which requires a live backend. Integration tests create a real setup but not within the target environment,

servicebroker_node/config/settings.json
fraenkel: I should have asked why we have catalog.json then.

servicebroker/lib/api/api.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+
+var settings = JSON.parse(
```
fraenkel: Why are we loading settings if it isn't used?

servicebroker/test/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
```
fraenkel: Is this file really necessary? We can just run mocha.

servicebroker/test/mocha.css
```diff
@@ -0,0 +1,314 @@
+@charset "utf-8";
```
fraenkel: Same question as above. Also applies to mocha.js.

servicebroker/test/ut/api/api_test.js
```diff
@@ -0,0 +1,17 @@
...
+var should = require('should');
+var API = require(path.join(__dirname, '../../../lib/api/api.js'));
+
+var api = new API();
+
+describe('Unit Test for API', function() {
+
+  it('getCatalog() should return catalog info with JSON format', function() {
+    var data = api.getCatalog();
+    data.should.have.property('services');
```
fraenkel: Shouldn't this match the contents of the catalog.json?

servicebroker/test/ut/index_test.js
```diff
@@ -0,0 +1,49 @@
...
+  it("should return catalog json", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .expect("Content-type", /json/)
+      .expect(200) // THis is HTTP response
+      .end(function(err, res) {
+        // HTTP status should be 200
+        res.status.should.equal(200);
+        JSON.stringify(res.body).length.should.equal(JSON.stringify(catalog).length);
+        res.body.services[0].id.should.equal(catalog.services[0].id);
```
fraenkel: Is this necessary since we already guaranteed it matched the catalog.json.

servicebroker/lib/api/api.js
cdlliuy: it is used to define a port when start service broker app. It is useless when cf app, but it should be useful if we want to define another port with bosh deploy. Is that right?

servicebroker/test/index.html
cdlliuy: When we do cmd "mocha init <folder>", then the index.html and others files are automatically added by mocha.

servicebroker/test/ut/api/api_test.js
cdlliuy: yes, it should match the content of catalog.json.   
Current unit test purpose is just verify the getCatalog() function return a valid JSON format, and include service definition inside. 
Do you mean we need to parse catalog.json file here, and make sure data=catalog exactly in unit test?

servicebroker/lib/api/api.js
fraenkel: where? I see it used in index.js not here.

servicebroker/test/index.html
fraenkel: mocha init is for client side usage not server side.

servicebroker/test/ut/index_test.js
cdlliuy: do you mean we can ignore the test for /v2/catalog since the implementation guaranteed the content should be same?

servicebroker/test/ut/api/api_test.js
fraenkel: yes, its more important here than in the index test which can do the same check or validate a catalog jsom document is returned.

servicebroker/test/ut/index_test.js
fraenkel: sorry I missed you only checked length.
you might as well just compare the data which some imply the same length.

servicebroker/lib/api/api.js
cdlliuy: ah, yes, sorry, this will be removed.

cdlliuy:
Refreshed again based on latest comments. Please help to review

fraenkel:
There should be no basic auth support. It isn't in the story.
We should add another story to cover the security model but it shouldn't be basic auth for the service broker.

servicebroker/lib/index.js
```diff
@@ -14,6 +15,28 @@ var app = express();
 app.use(bodyParser.urlencoded({ extended: true }));
 app.use(bodyParser.json());
 
+var auth = function (req, res, next) {
+  
+  function unauthorized(res) {
+    res.set('WWW-Authenticate', 'Basic realm=Authorization Required');
```
fraenkel: That isn't a realm. A realm must be a quoted string. (https://tools.ietf.org/html/rfc2617)

servicebroker/lib/index.js
```diff
@@ -14,6 +15,28 @@ var app = express();
...
+
+  if (user.name === settings.user && user.pass === settings.password) {
+    return next();
+  } else {
+    return unauthorized(res);
+  };
+  next();
+};
+
+app.use(auth);
```
fraenkel: Don't we want to do this first before we process the body?

servicebroker/test/index_test.js
```diff
@@ -42,8 +46,55 @@ describe("RESTful API test suite", function() {
...
+  });
+
+  it("should return 401 when incorrect auth format provided", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .set("Authorization", "invalidauth"  )
+      .expect(401, done);
+  });
+
+  it("should return 401 when auth info is not encoded with base64", function(done) {
```
fraenkel: redundant with the above test

servicebroker/test/index_test.js
```diff
@@ -42,8 +46,55 @@ describe("RESTful API test suite", function() {
...
+    supertest(server)
+      .get("/v2/catalog")
+      .set("Authorization", "Basic " + settings.user + ":" + settings.password )
+      .expect(401, done);
+  });
+
+  it("should return 401 when incorrect user/password provided", function(done) {
```
fraenkel: redundant

servicebroker/test/index_test.js
```diff
@@ -42,8 +46,55 @@ describe("RESTful API test suite", function() {
...
+
+  it("should return 401 when incorrect password provided", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .set("Authorization", "Basic " + new Buffer(settings.user+ ":" + "incorrectpassword").toString('base64') )
+      .expect(401, done);
+
+  });
+
+  it("should return 401 when incorrect basic auth format provided", function(done) {
```
fraenkel: redundant

servicebroker/test/index_test.js
```diff
@@ -42,8 +46,55 @@ describe("RESTful API test suite", function() {
...
+  });
+
+  it("should return 401 when incorrect user/password provided", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .set("Authorization", "Basic " + new Buffer("incorrectuser:incorrectpassword").toString('base64') )
+      .expect(401, done);
+  });
+
+  it("should return 401 when incorrect password provided", function(done) {
```
fraenkel: Where is the incorrect user test?

servicebroker/test/api/api_test.js
```diff
@@ -0,0 +1,25 @@
...
+var API = require(path.join(__dirname, '../../lib/api/api.js'));
+
+var catalog = JSON.parse(fs.readFileSync(path.join(__dirname, '../../config/catalog.json'), 'utf8'));
+var api = new API();
+
+describe('Unit Test for API', function() {
+
+  it('getCatalog() should return catalog info with JSON format', function() {
+    var data = api.getCatalog();
+    data.should.have.property('services');
```
fraenkel: This style of testing doesn't make sense.
Why didn't you just do

JSON.stringify(data).should.equal(JSON.stringify(catalog));

Which covers length and content. What you have doesn't guarantee extra data in the catalog.

servicebroker/test/index_test.js
```diff
@@ -0,0 +1,49 @@
...
+
+  it("should return catalog json", function(done) {
+    supertest(server)
+      .get("/v2/catalog")
+      .expect("Content-type", /json/)
+      .expect(200) // THis is HTTP response
+      .end(function(err, res) {
+        // HTTP status should be 200
+        res.status.should.equal(200);
+        JSON.stringify(res.body).length.should.equal(JSON.stringify(catalog).length);
```
fraenkel: See above.

fraenkel:
I will merge this but the items I identified need to be fixed.

servicebroker/lib/index.js
cdlliuy: Yes, using middleware of express, we do the auth first before process the body.  I added console log in code to show the sequence, and result is:
> node lib/index.js

processing the auth first
processing the body later

cdlliuy:
Add enhancement to address the final comment.

fraenkel:
Merged

PR #40, opened 2016-06-22
servicebroker: Change the order to invoke express middleware

cdlliuy:
[#121943583]

### DIFF
--- a/servicebroker/config/settings.json
+++ b/servicebroker/config/settings.json
@@ -1,5 +1,7 @@
 {
   "port": 8080,
   "user": "user",
-  "password": "password"
+  "password": "password",
+  "dbServer": "postgres://postgres@127.0.0.1:5432/",
+  "dbName": "servicebroker_db"
 }
--- a/servicebroker/README.md
+++ b/servicebroker/README.md
@@ -4,21 +4,31 @@ autoscaler-service-broker
 AutoScaler service broker is built by nodejs/expressjs
 
 ---
-**Local dependencies**
-```sh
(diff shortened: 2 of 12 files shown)


cdlliuy:
Also, the PR is updated to include the feature for "provision service"

fraenkel:
Next time split the PRs. It is confusing when you deliver multiple stories.

servicebroker/config/settings.json
```diff
@@ -1,5 +1,7 @@
 {
   "port": 8080,
   "user": "user",
```
fraenkel: For consistency, let us name this username instead of user.

fraenkel:
We need to have one style of code. I don't care which. Please look at how the API component used Sequelize. There may be a combination of both which is best but right now we have two different ways of doing similar DB operations.

servicebroker/config/settings.json
cdlliuy: Michael, how about this time? I see the branch has conflicts now. Should I raise another 2 PR instead?

fraenkel: You should pull in the latest develop and rebase your branch. The rebase will probably have a conflict which should force you to fix things. At which point once you commit the rebase you should be good with a PR.

cdlliuy: ok, will do the rebase, and also move the createService part out of the PR.   For create Service part,  I will raise another PR, and change the style for Sequelize call

fraenkel:
Merged the reorder commit.

PR #41, opened 2016-06-22
use cf-http as the http client

boyang9527:
[#121523475]

- add submodule https://github.com/cloudfoundry-incubator/cf_http
- use cf_http to replace mhttp package
- use cf_http/handlers to write JSON response


fraenkel:
merged

PR #43, opened 2016-06-23
servicebroker: provisionservice

cdlliuy:
this commit is based on previous PR#42


servicebroker/lib/servicebroker.js
```diff
@@ -0,0 +1,45 @@
...
+
+var fs = require('fs');
+var path = require('path');
+var async = require('asyncawait/async');
+var await = require('asyncawait/await');
+var model = require(path.join(__dirname, './model'))();
+
+var ServiceBroker = function ServiceBroker() {};
+ServiceBroker.prototype.getCatalog = function() {
+  var catalog = JSON.parse(
```
fraenkel: Why not just use res.sendFile via express?

servicebroker/lib/servicebroker.js
```diff
@@ -0,0 +1,45 @@
...
+    fs.readFileSync(path.join(__dirname, '../config/catalog.json'), 'utf8'));
+  return catalog;
+};
+
+ServiceBroker.prototype.provisionService = async(function(serviceId, orgId, spaceId) {
```
fraenkel: This is very confusing for many reasons.

1. We are making the whole thing async, but the actual findOrCreate which is async we made synchronous. If we plan to make this async, then don't bother wrapping.
2. We are returning status codes when we do require this to handle the http request. It is mixing ideas.

Let us follow the pattern that the API component has.

This file should go into a routes subdirectory and named serviceinstances.js
The HTTP handler and DB code should be co-mingled. There is no point in having the separation when we have no special business logic.

servicebroker/test/servicebroker_test.js
```diff
@@ -0,0 +1,48 @@
...
+var async = require('asyncawait/async');
+var await = require('asyncawait/await');
+var ServiceBroker = require(path.join(__dirname, '../lib/servicebroker.js'));
+
+var catalog = JSON.parse(fs.readFileSync(path.join(__dirname, '../config/catalog.json'), 'utf8'));
+var servicebroker = new ServiceBroker();
+
+describe('Unit Test for Service Broker', function() {
+
+  it('getCatalog() should return catalog info with JSON format', function() {
```
fraenkel: This should be structured differently.
```
  describe('getCatalog', function() {
    it('returns a catalog', function() ....
```

servicebroker/test/servicebroker_test.js
```diff
@@ -0,0 +1,48 @@
...
+    var data = servicebroker.getCatalog();
+    JSON.stringify(data).should.equal(JSON.stringify(catalog));
+  });
+  it('provisionService() should return 201 when serviceId, orgId and spaceId are new ones', async(function() {
```
fraenkel:
```
describe('provisionService', function() {
  var serviceId = uuid.v4(),
        spaceId = ....
  it('creates a new instance', function() {
  }

  context('when an instance already exists', function() {
    beforeEach(function() {
       return servicebroker.provisionService(serviceId, orgId, spaceID)
    }

    context('when orgId and spaceId are identical', function() {
      it('returns a 200', ...
    })

   context('when serviceId already exists', function() {
     it('returns a 409', ....
   })

**Comment by fraenkel** [comment]
Make sure you update the root README since it documents the unit test steps which have changed given your changes to your script definitions.

File: servicebroker/lib/servicebroker.js (hunk 2 above)
Comment (cdlliuy) [inline servicebroker/lib/servicebroker.js]: Just want to make sure #1,  the findOrCreate returns a Promise, so we need to handle it with promise way findOrCreate.then()  as well?  The await/async wrap works correctly, but we did need to add await/async wrapper for each invocation.   All the other comments will be addressed.

File: servicebroker/lib/servicebroker.js (hunk 1 above)
Comment (qibobo) [inline servicebroker/lib/servicebroker.js]: Seems that res.sendFile reads the file when each time it is called. Should we read the catalog file only once and cache it?

**Comment by fraenkel** [comment]
@qibobo I realize res.sendFile sends it every time, but how often is getCatalog really called? It is a memory vs speed tradeoff. I vote for less memory since it shouldn't be called that often. The downside is that we do need to verify the file once to make sure it is valid json.

File: servicebroker/lib/routes/serviceinstance.js [hunk 5]
```diff
@@ -0,0 +1,39 @@
...
+      spaceId: spaceId,
+      where: {
+        serviceId: serviceId,
+        orgId: orgId,
+        spaceId: spaceId
+      }
+    })
+    .then(function(result){
+      var isNew = result[1];
+      if (isNew === true) {
```
fraenkel: why not just set a status code based on isNew. The response is the same just with a different status code.

servicebroker/lib/routes/serviceinstance.js
```diff
@@ -0,0 +1,39 @@
...
+         res.status(201).send({ "dashboard_url": "" }).end();
+      } else {
+         res.status(200).send({ "dashboard_url": "" }).end();
+      }
+    }).catch(function(err){
+      if (err.name.indexOf('SequelizeUniqueConstraintError') >= 0) {
```
fraenkel: Why don't we do this the right way? 
if (err instanceof UniqueConstraintError) ...

servicebroker/lib/routes/serviceinstance.js
```diff
@@ -0,0 +1,39 @@
...
+        res.status(409).end();
```
fraenkel: Same as above, just set a status code and a common response.

servicebroker/test/index_test.js
```diff
@@ -7,13 +7,11 @@ var path = require('path');
...
 
 // This agent refers to PORT where program is runninng.
-var catalog = JSON.parse(fs.readFileSync(path.join(__dirname, '../config/catalog.json'), 'utf8'));
 var settings = JSON.parse(
   fs.readFileSync(path.join(__dirname, '../config/settings.json'), 'utf8'));
 var auth = new Buffer(settings.username + ":" + settings.password).toString('base64')
 
-// UNIT test begin
-describe("RESTful API test suite", function() {
+describe("Invalid path for RESTful API", function() {
```
fraenkel: normally you only have 1 top level describe that everything nests under.
describe('routing', function() {
  before(function() {
   start server
  })
  after(function() {
  stop server
 })

 context('with an invalid path', ....)
 describe('authorization', function() ....)
})

We don't need to start and stop the server for each test since we don't change the configuration.

servicebroker/test/routes/serviceinsance_test.js
```diff
@@ -0,0 +1,73 @@
...
+  fs.readFileSync(path.join(__dirname, '../../config/settings.json'), 'utf8'));
+var auth = new Buffer(settings.username + ":" + settings.password).toString('base64');
+describe('provisionService', function() {
+  var server;
+  var serviceId = uuid.v4();
+  var orgId = uuid.v4();
+  var spaceId = uuid.v4();
+  var orgIdAgain = uuid.v4();
+  var spaceIdAgain = uuid.v4();
+  beforeEach(function() {
```
fraenkel: We can make this a before()

servicebroker/test/routes/serviceinsance_test.js
```diff
@@ -0,0 +1,73 @@
...
+    delete require.cache[require.resolve('../../lib/index.js')];
+    server = require(path.join(__dirname, '../../lib/index.js'));
+  });
+
+  afterEach(function(done) {
```
fraenkel: Make this an after()

servicebroker/test/routes/serviceinsance_test.js
```diff
@@ -0,0 +1,73 @@
...
+var fs = require('fs');
+var path = require('path');
+var supertest = require("supertest");
+var should = require('should');
+var uuid = require('uuid');
+
+var settings = JSON.parse(
+  fs.readFileSync(path.join(__dirname, '../../config/settings.json'), 'utf8'));
+var auth = new Buffer(settings.username + ":" + settings.password).toString('base64');
+describe('provisionService', function() {
```
fraenkel: You need to truncate the table before each test.
Right now you are getting lucky based on execution order of this test and this test in sequence of the others.
Mocha is coming with a randomize support which we will run with eventually.

README.md
```diff
@@ -66,7 +66,7 @@ npm test
 popd
 
 pushd servicebroker
-npm run bootstrap
+npm install
```
fraenkel: What about initializing the db?

fraenkel:
Merging the current code but please address the new issues.

PR #45, opened 2016-06-27
service broker code enhancement

cdlliuy:
1.  refactor service instance routes & models test
2. pretty code
3. add unit test for db uri parse
4. add support for liquidbase

### DIFF
--- a/servicebroker/config/servicebroker.db.changelog.yaml
+++ b/servicebroker/config/servicebroker.db.changelog.yaml
@@ -1,0 +1,25 @@
+--- 
+databaseChangeLog:
+   - changeSet:
+      id: 1
+      author: pradyutsarma
+      changes:
+        - createTable:
+            tableName: service_instance
+            columns:
+              - column:
+                  name: service_id
+                  type: varchar(40)
+                  constraints:
+                    primaryKey: true
+                    nullable: false
+              - column:
+                  name: org_id
+                  type: varchar(40)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: space_id
+                  type: varchar(40)
+                  constraints:
+                    nullable: false
--- a/servicebroker/test/model/serviceinstance_test.js
+++ b/servicebroker/test/model/serviceinstance_test.js
@@ -1,35 +1,71 @@
 'use strict';
 var expect = require('chai').expect;
+var serviceInstance = require('../../lib/models')().service_instance;
 
-var fake_serviceinstance = { 'serviceId':'test_service', 'orgId':'test_org', 'spaceId':'test_space' };
+var fake_serviceId = "test_service";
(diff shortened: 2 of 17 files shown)


servicebroker/test/model/serviceinstance_test.js
```diff
@@ -1,35 +1,71 @@
...
+            done();
+          });
+      });
+  });
+
+  it('Should throw UniqueConstraintError with conflict record', function(done) {
+    serviceInstance.findOrCreate(fake_condition)
+      .then(function(result) {
+        serviceInstance.findOrCreate(fake_condition2)
+          .then(function(result) {})
```
fraenkel: shouldn't this be calling done with an error since we should never get here?!?

servicebroker/test/model/serviceinstance_test.js
```diff
@@ -38,7 +38,11 @@ describe('Service_instance Model Definition Test Suite', function() {
         expect(result[0].serviceId).to.equal(fake_condition.serviceId);
         expect(result[1]).to.equal(true);
         done();
-      });
+      }) 
+      .catch (function(error) {
+          console.log ("error in findOrCreate: " + error.message);
+          done();
```
fraenkel: Shouldn't this be done(error)? So we fail.

servicebroker/config/servicebroker.db.changelog.yaml
```diff
@@ -0,0 +1,25 @@
+--- 
+databaseChangeLog:
+   - changeSet:
+      id: 1
+      author: pradyutsarma
```
fraenkel: interesting author

fraenkel:
Caught some issues with misusing done.

servicebroker/test/model/serviceinstance_test.js
cdlliuy: we should return promise here instead of calling done. Mocha documents "Alternately, instead of using the done() callback, you may return a Promise. This is useful if the APIs you are testing return promises instead of taking callbacks".   
We are blocked here for about 1 hour by missing "done" or "return" here.  The cmd "mocha test/models/serviceinstance_test" passed, and the cmd " mocha test/routes/serviceinstance_test" passed as well, but the cmd "mocha test --recuisive" sometimes failed in "test/routes/serviceinstance_test"  with error "done() callback is not handled correctly".  It did take time to figure out the done is missing in models/serviceinstance_test ...  

I would like to follow the "return" approach to handle promise as suggested by the document

fraenkel: SGTM, consistency is best.

cdlliuy:
updated with latest commit

fraenkel:
Merged

PR #48, opened 2016-07-08
As a user, I can bind with a policy

qibobo:
[#119504807]


servicebroker/config/settings.json
```diff
@@ -2,5 +2,6 @@
   "port": 8080,
   "username": "username",
   "password": "password",
-  "dbUri": "postgres://postgres@127.0.0.1:5432/autoscaler"
+  "dbUri": "postgres://postgres@127.0.0.1:5432/servicebroker_db",
```
fraenkel: this does not match the README that is available for the app-autoscaler.
while this dbUri is for the servicebroker_db, right now for development, we have 1 DB, autoscaler.

servicebroker/lib/routes/binding.js
```diff
@@ -0,0 +1,82 @@
...
+module.exports = function(app) {
+  var path = require('path');
+  var fs = require('fs');
+  var request = require('request');
+  var logger = require(path.join(__dirname, '../logger/logger.js'));
+  var models = require(path.join(__dirname, '../models'))();
+
+  var settings = JSON.parse(
+    fs.readFileSync(path.join(__dirname, '../../config/settings.json'), 'utf8'));
+  var apiServerUtil = require(path.join(__dirname, '../util/apiServerUtil.js'))(settings.apiServerUri);
```
fraenkel: Couple of comments.
1. This is the second time we have read the settings and we don't do validation anywhere
2. More importantly, we should just inject the apiServerUri to this module. This allows for easier testing.

servicebroker/lib/routes/binding.js
```diff
@@ -0,0 +1,82 @@
...
+    var bindingId = req.params.binding_id;
+    var appId = req.body.app_guid;
+    var policyJSON = req.body.parameters;
+    if (typeof(policyJSON) === "undefined") {
+      res.status(400).json({ "description": messageUtil.getMessage("POLICY_REQUIRED") });
+    } else {
+      models.binding.findAll({ where: { appId: appId } }).then(function(result) {
+        var length = result.length;
+        if (length === 0) {//no binding , so create one
+          models.sequelize.transaction().then(function(t) {
```
fraenkel: Shouldn't the findAll be part of the transaction? Otherwise I can have 2 requests that both fail the findAll.

servicebroker/lib/routes/binding.js
```diff
@@ -0,0 +1,82 @@
...
+
+            });
+          }).catch(function(error2) { //catch transaction
+            logger.error("Fail to bind service when handle transaction, error=>" + error2);
+            t.rollback();
+            res.status(500).json({});
+          });
+        } else if (length > 1) { // an app has been bound to more than one service instance, this error should not exist
+          logger.error("");
+          res.status(499).json({ "description": messageUtil.getMessage("DUPLICATED_BIND", { "applicationId": appId }) });
```
fraenkel: 499 is not a valid response code.

servicebroker/lib/routes/binding.js
```diff
@@ -0,0 +1,82 @@
...
+        } else if (length == 1) { // an app has been bound to a service instance
+          var bindingRecord = result[0];
+          if (bindingRecord.serviceInstanceId === serviceInstanceId) {
+            res.status(409).json({});
+          } else {
+            res.status(499).json({ "description": messageUtil.getMessage("DUPLICATED_BIND", { "applicationId": appId }) });
```
fraenkel: See above.

servicebroker/lib/util/apiServerUtil.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+var request = require('request');
+
+module.exports = function(apiServerUri){
+  var apiUtil = {};
+  apiServerUri = apiServerUri.replace(/\/$/g, "").toLowerCase();
```
fraenkel: What are we trying to do here?

servicebroker/lib/util/apiServerUtil.js
```diff
@@ -0,0 +1,20 @@
...
+  apiUtil.attachPolicy= function(appId, policy, callback) {
+    var options = {
+      uri: apiServerUri + '/v1/apps/' + appId + '/policy',
+      method: 'PUT',
+      json: policy
+    };
+    request(options, function(error, response, body) {
+    callback(error, response);
```
fraenkel: indentation

servicebroker/lib/util/apiServerUtil.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = function(apiServerUri){
+  var apiUtil = {};
+  apiServerUri = apiServerUri.replace(/\/$/g, "").toLowerCase();
+  apiUtil.attachPolicy= function(appId, policy, callback) {
+    var options = {
+      uri: apiServerUri + '/v1/apps/' + appId + '/policy',
+      method: 'PUT',
+      json: policy
+    };
+    request(options, function(error, response, body) {
```
fraenkel: is this better than request.put({uri: foo, json: policy}, function(err, rsp, body) {
})?

servicebroker/lib/util/messageUtil.js
```diff
@@ -0,0 +1,23 @@
...
+var path = require('path');
+var fs = require('fs');
+var catalog = JSON.parse(
+  fs.readFileSync(path.join(__dirname, '../../config/catalog.json'), 'utf8'));
+module.exports = function(){
+  var messageUtil = {};
+  var message = {
+    POLICY_REQUIRED: "Policy is required as a parameter for binding request.",
+    INVALID_POLICY_FORMAT: "Invalid policy format.",
+    DUPLICATED_BIND: "Another " + catalog.services[0].name + " service instance already bound to current application ${applicationId}. The multiple binding is not supported.",
```
fraenkel: DUPLICATE_BIND: "A" + catalog.services[0].name + " service instance is already bound to application ${applicationId}. Multiple bindings are not supported.",

servicebroker/lib/util/messageUtil.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+var path = require('path');
+var fs = require('fs');
+var catalog = JSON.parse(
+  fs.readFileSync(path.join(__dirname, '../../config/catalog.json'), 'utf8'));
+module.exports = function(){
+  var messageUtil = {};
+  var message = {
+    POLICY_REQUIRED: "Policy is required as a parameter for binding request.",
```
fraenkel: POLICY_REQUIRED: "Policy is required as a parameter.",

servicebroker/lib/util/messageUtil.js
```diff
@@ -0,0 +1,23 @@
...
+    INVALID_POLICY_FORMAT: "Invalid policy format.",
```
fraenkel: Where is this used?

servicebroker/package.json
```diff
@@ -14,15 +14,19 @@
     "basic-auth": "^1.0.4",
     "body-parser": "^1.15.1",
     "express": "^4.13.4",
-    "pg": "^6.0.1",
-    "sequelize": "^3.23.4"
+    "pg": "^6.0.0",
```
fraenkel: Why backlevel?

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+  'appId': fake_appId2,
+  'bindingId': fake_bindingId,
+  'timestamp': new Date().getTime() 
+};
+
+describe('Binding Model Definition Test Suite', function() {
+
+  before(function(done) {
+    serviceInstance.sequelize.sync().then(function(result) {
+      binding.sequelize.sync().then(function(result) {
```
fraenkel: We need to remove sequelize.sync() since it recreates the tables and doesn't validate that your liquibase table definitions are correct.

The right approach is to just call sequelize.truncate() which will truncate all tables.

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+        done(error1);
+      });
+    }).catch(function(error) {
+      console.log("Failed to sync model serviceInstance, error: " + error);
+      done(error);
+    });
+  });
+
+  beforeEach(function(done) {
+    return binding.truncate({ cascade: true }).then(function(result) {
```
fraenkel: With cascade: true, you should be able to truncate the service_instance table and have it wipe the binding table.
We should not be doing a findOrCreate here but in the contexts that require an item in the DB.

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
+'use strict';
+var expect = require('chai').expect;
+var serviceInstance = require('../../lib/models')().service_instance;
+var binding = require('../../lib/models')().binding;
+
+var fake_serviceInstanceId = "test_serviceinstance";
```
fraenkel: The naming is wrong.
These are not fakes but actual names.
Fakes are when you aren't using a real instance.

servicebroker/package.json
```diff
@@ -14,15 +14,19 @@
...
+    "sequelize": "^3.23.3",
+    "uuid": "^2.0.2"
   },
   "devDependencies": {
     "uuid": "^2.0.2",
     "eslint": "^2.11.1",
     "chai": "^3.5.0",
     "mocha": "^2.5.3",
+    "nock": "^8.0.0",
     "should": "^9.0.0",
```
fraenkel: I thought we were removing should?

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+
+  context(' create and find', function() {
+    context('when there is no record ', function() {
+      it('should create a new binding record successfully with the 1st create', function() {
+        return binding.create(fake_binding_condition)
+          .then(function(result) {
+            expect(result[0].bindingId).to.equal(fake_binding_condition.bindingId);
+            expect(result[1]).to.equal(true);
+          }).catch(function(err) {
+
```
fraenkel: will we fail?

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+          });
+      });
+    });
+
+    context('when an binding already exists ', function() {
+      beforeEach(function(done) {
+        return binding.create(fake_binding_condition).then(function(result) {
```
fraenkel: why use done? Use the promise from create.

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+          .then(function(instance) {
+            expect(instance).to.be.null;
+          })
+      });
+    });
+
+    context('when an instance already exists ', function() {
+      beforeEach(function(done) {
+        return binding.create(fake_binding_condition).then(function(result) {
+          done();
```
fraenkel: see above

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+        return binding.findById(fake_binding_condition.bindingId)
+          .then(function(instance) {
+            expect(instance.bindingId).to.equal(fake_binding_condition.bindingId);
+          })
+      });
+
+    });
+  });
+
+  context(' destory ', function() {
```
fraenkel: spelling

servicebroker/test/models/binding_test.js
```diff
@@ -0,0 +1,183 @@
...
+    context('when there is no record ', function() {
+      it('should return 0 with nonexist binding', function() {
```
fraenkel: it('should return 0')

servicebroker/test/models/serviceinstance_test.js
```diff
@@ -22,8 +22,10 @@ var fake_condition2 = {
 
 describe('Service_instance Model Definition Test Suite', function() {
 
-  before(function() {
-    serviceInstance.sequelize.sync();
+  before(function(done) {
+    serviceInstance.sequelize.sync().then(function(result){done();}).catch(function(error){
```
fraenkel: see above

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var supertest = require("supertest");
+var expect = require('chai').expect;
+var nock = require('nock');
```
fraenkel: Did we need nock over sinon and its fake server?

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+  var fake_service_condition = {
+    'serviceInstanceId': serviceInstanceId,
+    'orgId': orgId,
+    'spaceId': spaceId,
+    where: { 'serviceInstanceId': serviceInstanceId, 'orgId': orgId, 'spaceId': spaceId },
+  };
+  var policy = { "policy": "testPolicy" };
+
+  before(function(done) {
+    serviceInstance.sequelize.sync().then(function(result) {
```
fraenkel: no sync

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+    }).catch(function(error2) {
+      console.log("Failed to sync model serviceInstance, error: " + error2);
+      done(error2);
+    });
+  });
+
+  beforeEach(function(done) {
+    delete require.cache[require.resolve('../../lib/index.js')];
+    server = require(path.join(__dirname, '../../lib/index.js'));
+    binding.truncate({ cascade: true }).then(function(result) {
```
fraenkel: see above

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+      }).catch(function(error2) {
+        done(error2);
+      });
+    }).catch(function(error3) {
+      done(error3);
+    });
+  });
+
+  afterEach(function(done) {
+    server.close(done);
```
fraenkel: Why are we starting and stopping a server for every test?
Start it once.

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+        supertest(server)
+          .put("/v2/service_instances/" + serviceInstanceId + "/service_bindings/" + bindingId)
+          .set("Authorization", "Basic " + auth)
+          .send({ "app_guid": appId, "parameters": policy })
+          .expect(201)
+          .expect('Content-Type', /json/)
+          .expect({}, done);
+      });
+      context("when the api server return error", function() {
+        it("return a 400", function(done) {
```
fraenkel: these tests highlight an issue in the code.
What happens when the api server returns some unexpected non-2xxx code?
What is returned?

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+  scope = nock(settings.apiServerUri)
+    .put(/\/v1\/apps\/.*\/policy/)
+    .reply(201, {
+      'success': true,
+      'error': null,
+      'result': "created"
+    });
+}
+
+function initNockBind400() {
```
fraenkel: Why not just have a initNockBindError(statusCode)?

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+        supertest(server)
+          .put("/v2/service_instances/" + serviceInstanceId + "/service_bindings/" + bindingId)
+          .set("Authorization", "Basic " + auth)
+          .send({ "app_guid": appId, "parameters": policy })
+          .expect(201)
+          .expect('Content-Type', /json/)
+          .expect({}, done);
+      });
+
+      context('when erviceInstanceId and appId are conflict with an existed one', function() {
```
fraenkel: typo

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+            .put("/v2/service_instances/" + serviceInstanceId2 + "/service_bindings/" + bindingId)
+            .set("Authorization", "Basic " + auth)
+            .set('Accept', 'application/json')
+            .send({ "app_guid": appId, "parameters": policy })
+            .expect(499)
+            .expect({"description":messageUtil.getMessage("DUPLICATED_BIND",{"applicationId":appId})}, done);
+        });
+      });
+    });
+
```
fraenkel: How do we test when the binding create fails?

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+            .put("/v2/service_instances/" + serviceInstanceId + "/service_bindings/" + bindingId)
+            .set("Authorization", "Basic " + auth)
+            .set('Accept', 'application/json')
+            .send({ "app_guid": appId, "parameters": policy })
+            .expect(409)
+            .expect('Content-Type', /json/)
+            .expect({}, done);
+        });
+      });
+      context('when bindingId are conflict with an existed one', function() {
```
fraenkel: this doesn't test anything new since it is all about the app id.

servicebroker/test/routes/binding_test.js
```diff
@@ -0,0 +1,204 @@
...
+            .set('Accept', 'application/json')
+            .send({ "app_guid": appId, "parameters": policy })
+            .expect(499)
+            .expect({"description":messageUtil.getMessage("DUPLICATED_BIND",{"applicationId":appId})}, done);
+        });
+      });
+    });
+
+  });
+});
```
fraenkel: Where is a test for trying to bind to a non-existent service instance?

fraenkel:
When you make the changes above, just put them in a separate commit so I can just review the updates. I will squash them before I merge.

servicebroker/lib/routes/binding.js
cdlliuy: Here， we want to deliver a specific response to end-user. If using predefined error code in service broker API, the customized response can't be displayed with cf cli.   So, we used a customized response code here.  499 is not a valid response code,  so we won't worry about to overwrite any pre-defined HTTP response code.   At least, 499 works with current cf cli.  we can check again whether the specific msg could be delivered as well with other 4xx error, i.e. 400.

servicebroker/lib/routes/binding.js
qibobo: findById is used to find by primary key while here we find by appId and it is not the primary key.
Seems managed transaction only rollback when there is error, while we may need to roll back when the api server return some results that we do not expect.

servicebroker/lib/routes/binding.js
qibobo: @fraenkel  Do you have a good way for object injection？

servicebroker/lib/util/apiServerUtil.js
qibobo: Familiar with dbUri , the provided api server uri may be "http://autoscaler.boshlite.com" or "http://autoscaler.boshlite.com/". So we need to remove the last "/".

servicebroker/test/models/binding_test.js
qibobo: For the foreign key, we do not use "cascade" but "restrict", it means when try to delete a service instance that has been referenced by a binding, postgres will alert exceptions. We use this to guarantee that user must do unbind and then delete service instance.

servicebroker/test/routes/binding_test.js
qibobo: Liu Ying and I do some investigation about sinon and did not find the way it work with supertest. Then we found "nock" is easy to use and fulfill our request, so we use "nock".

servicebroker/test/routes/binding_test.js
qibobo: It returns the code just api returns to broker, there is a test case for code "500" .

servicebroker/test/routes/binding_test.js
qibobo: There are two scenarios that will return 409. One  is an app has bound to another service instance and the other is trying to insert a binding id which already in the db.

servicebroker/lib/routes/binding.js
fraenkel: I care less about how the CLI behaves because it can change. If we want a message displayed for 400 then we get it fixed.
Show me where 499 is defined as an expected HTTP response.

servicebroker/lib/util/apiServerUtil.js
fraenkel: There are things like url-join but I get it. Might want to add a comment for the next person.

servicebroker/test/routes/binding_test.js
fraenkel: You have a test that says when 500 is returned a 500 is returned. My point is that when a non-200 and non-500 is returned, what happens? We have a gap in the tests.

servicebroker/test/routes/binding_test.js
fraenkel: So name the tests that way,
1. When an app is already bound
2. When the binding id already exists

I don't understand what when a serviceInstanceId and appId are in conflict.  In conflict with what?

servicebroker/lib/routes/binding.js
fraenkel: Are we going to create an index on the appId then?

servicebroker/lib/routes/binding.js
fraenkel: Just add additional arguments to the function.

servicebroker/lib/routes/binding.js
```diff
@@ -20,57 +20,64 @@ module.exports = function(app) {
...
-              apiServerUtil.attachPolicy(appId, policyJSON, function(error, response) {
-                var statusCode = response.statusCode;
-                if (!error && (statusCode == 200 || statusCode == 201)) {
-                  t.commit();
-                  res.status(statusCode).json({});
+      models.sequelize.transaction().then(function(t) {
 
-                } else {
-                  logger.error("Fail to bind service when call api server, error=>" + error);
+        models.service_instance.findById(serviceInstanceId).then(function(result) {
```
fraenkel: This seems a bit unnecessary since the create should fail if the service instance doesn't exist. Just trying to keep the number of queries to a minimum here.

servicebroker/lib/routes/binding.js
```diff
@@ -20,57 +20,64 @@ module.exports = function(app) {
...
+                  apiServerUtil.attachPolicy(appId, policyJSON, function(error, response) {
+                    var statusCode = response.statusCode;
+                    if (!error && (statusCode == 200 || statusCode == 201)) {
+                      t.commit();
+                      res.status(statusCode).json({});
+
+                    } else {
+                      logger.error("Fail to bind service when call api server, error=>" + error);
+                      t.rollback();
+                      res.status(statusCode).json({});
```
fraenkel: We should log the status code too.
We should not return the API server status code. It is a 500.

servicebroker/lib/routes/binding.js
```diff
@@ -20,57 +20,64 @@ module.exports = function(app) {
...
-              t.rollback();
-              if (error1 instanceof models.sequelize.UniqueConstraintError) {
-                res.status(409).json({});
-              } else {
-                logger.error("Fail to handle request: " + JSON.stringify(req) + " with ERROR: " + error1);
-                res.status(500).json({});
+                });
+              } else if (length > 1) { // an app has been bound to more than one service instance, this error should not exist
+                logger.error("");
+                res.status(499).json({ "description": messageUtil.getMessage("DUPLICATE_BIND", { "applicationId": appId }) });
```
fraenkel: 499 is not a valid HTTP status code.

servicebroker/test/routes/binding_test.js
```diff
@@ -176,6 +150,11 @@ describe('binding RESTful API', function() {
         });
       });
       context('when bindingId are conflict with an existed one', function() {
+        beforeEach(function(done) {
+          serviceInstance.findOrCreate(service_condition2).then(function(result) {
```
fraenkel: using findOrCreate bothers me. It would seem we don't know if it should already be there or not.
If we expect it not to be there, we should use create()
We also don't need done here.

servicebroker/test/routes/binding_test.js
```diff
@@ -188,14 +167,19 @@ describe('binding RESTful API', function() {
         });
       });
       context('when an service instance has already bound to an application', function() {
+        beforeEach(function(done) {
+          serviceInstance.findOrCreate(service_condition2).then(function(result) {
```
fraenkel: See above.

servicebroker/lib/routes/binding.js
qibobo: Here， we want to deliver a specific response to end-user. If using predefined error code in service broker API, the customized response can't be displayed with cf cli. So, we used a customized response code here. 499 is not a valid response code, so we won't worry about to overwrite any pre-defined HTTP response code. At least, 499 works with current cf cli. we can check again whether the specific msg could be delivered as well with other 4xx error, i.e. 400.

fraenkel: My point is that 400 is the right status code. If the CLI isn't displaying the message, we should open an issue against them to have it fixed.
CF_TRACE=true will always display it as a last resort.

servicebroker/lib/config/setting.js
```diff
@@ -0,0 +1,29 @@
+//'use strict';
+
+module.exports = function(settings) {
```
fraenkel: where or when are we validating the settings?

servicebroker/lib/models/index.js
```diff
@@ -1,12 +1,11 @@
...
+module.exports = function(settings, callback) {
 
   var fs = require('fs');
   var path = require('path');
   var Sequelize = require('sequelize');
   var logger = require(path.join(__dirname, '../logger/logger.js'));
-  var settings = JSON.parse(fs.readFileSync(path.join(__dirname, '../../config/settings.json'), 'utf8'));
-  var dbConnectionInfo = require(path.join(__dirname, '../../lib/util/dbConnectionInfo.js'))(settings.dbUri);
-  var sequelize = new Sequelize(dbConnectionInfo.dbUri, { logging: false });
+  console.log ("index.js : " + settings.db.uri);
```
fraenkel: Why are we logging this?

servicebroker/lib/models/index.js
```diff
@@ -1,12 +1,11 @@
...
+  var sequelize = new Sequelize(settings.db.uri, { logging: false });
```
fraenkel: If all we need is the URI why are we passing in the entire settings?

servicebroker/lib/routes/binding.js
```diff
@@ -1,15 +1,13 @@
...
   var path = require('path');
   var fs = require('fs');
   var request = require('request');
   var logger = require(path.join(__dirname, '../logger/logger.js'));
-  var models = require(path.join(__dirname, '../models'))();
+  var models = require(path.join(__dirname, '../models'))(settings);
 
-  var settings = JSON.parse(
-    fs.readFileSync(path.join(__dirname, '../../config/settings.json'), 'utf8'));
   var apiServerUtil = require(path.join(__dirname, '../util/apiServerUtil.js'))(settings.apiServerUri);
```
fraenkel: If all we need is the apiServerUri why are we passing in settings?

servicebroker/lib/routes/catalog.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-module.exports = function(app) {
+module.exports = function(app, settings) {
```
fraenkel: where do we use settings?

servicebroker/lib/routes/index.js
```diff
@@ -1,14 +1,14 @@
 var fs = require('fs');
 var path = require('path');
 
-module.exports = function(app) {
+module.exports = function(app, settings) {
```
fraenkel: ditto

servicebroker/test/config/setting_test.js
```diff
@@ -0,0 +1,91 @@
...
+  context('Parse dbUri', function() {
+    it('Should parse dbUri correctly', function() {
+      var dbSetting = require(path.join(__dirname, '../../lib/config/setting.js'))(setting).db;
+      expect(dbSetting.uri).to.equal(setting.dbUri);
+      expect(dbSetting.server).to.equal("postgress://postgres@server:port");
+      expect(dbSetting.name).to.equal("dbname");
+    });
+
+    it('Should filter the last slash when parse dbUri', function() {
+      var settingTmp = {};
```
fraenkel: why not create setting in a beforeEach? This would just be changing the dbUri before invoking the require

servicebroker/lib/config/setting.js
```diff
@@ -1,5 +1,6 @@
-//'use strict';
-
+'use strict';
```
fraenkel: The more I look at this, the more confused I get.

I would have just had the validation here as a separate function added to the object returned.
Let the user call validate and decide how to handle the success or failure.

I don't see why parse() is even there since it is called as part of the return.
You might as well just inline it and be done with it.

servicebroker/lib/util/validateUtil.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+module.exports = {
```
fraenkel: This is too generic. It would probably be quicker just to write what we expect which is why it is best placed in the settings.validate method.

if (x != null) will catch both undefined and null so doing isUndefined and isNull isn't needed

servicebroker/test/config/setting_test.js
```diff
@@ -2,90 +2,166 @@
...
   "password": "password",
   "dbUri": "postgress://postgres@server:port/dbname",
   "apiServerUri": "http://apiserveruri"
 }
+var settingTmp = {};
+var settingObj, dbSetting;
+var apiServerUriParsed;
+var settingUndefined;
 describe('config setting Test Suite', function() {
+  beforeEach(function(done) {
```
fraenkel: we don't need done

servicebroker/test/config/setting_test.js
```diff
@@ -2,90 +2,166 @@
...
   it('Should get username and password correctly', function() {
-    var settingObj = require(path.join(__dirname, '../../lib/config/setting.js'))(setting);
     expect(settingObj.username).to.equal(setting.username);
     expect(settingObj.password).to.equal(setting.password);
   });
+  context('Get setting', function(){
+    context('when setting is null',function(){
+      beforeEach(function(done){
+        settingTmp = null;
+        done();
```
fraenkel: no done

servicebroker/test/util/validateUtil_test.js
```diff
@@ -0,0 +1,102 @@
...
+
+var path = require('path');
+var expect = require('chai').expect;
+var validateUtil = require(path.join(__dirname, '../../lib/util/validateUtil.js'));
+var testNullObject, testUndefinedObject;
+var testObject, keys;
+describe('validateUtil Test Suite', function() {
+  context('isNull()', function() {
+    context('when testObject is null', function() {
+      beforeEach(function(done) {
```
fraenkel: no done

qibobo:
@fraenkel @cdlliuy  updated again.

servicebroker/lib/index.js
```diff
@@ -7,13 +7,16 @@ var bodyParser = require('body-parser');
 var fs = require('fs');
 var path = require('path');
 var logger = require(path.join(__dirname, './logger/logger.js'));
+var validateUtil = require(path.join(__dirname, './util/validateUtil.js'));
 var settings = require(path.join(__dirname, './config/setting.js'))((JSON.parse(
   fs.readFileSync(path.join(__dirname, '../config/settings.json'), 'utf8'))));
-
+if (validateUtil.validate(settings, ['port', 'username', 'password', 'db', 'apiServerUri']) === false) {
```
fraenkel: The user has no way of knowing what is wrong/missing with the config.

servicebroker/lib/util/validateUtil.js
```diff
@@ -1,29 +1,21 @@
...
   validate: function(object, keys) {
-    if (this.isNull(object) === false && this.isUndefined(object) === false) {
-      if (this.isNull(keys) === false && this.isUndefined(keys) === false) {
+    if (object != null) {
+      if (keys != null) {
         if (keys instanceof Array) {
           for (var key in keys) {
-            if (this.isUndefined(object[keys[key]]) === true || this.isNull(object[keys[key]]) === true) {
-              return { valid: false, msg: 'The value of ' + key + ' should not be null or undefined' };
+            if (typeof (object[keys[key]]) === "undefined" || object[keys[key]] === null) {
```
fraenkel: isn't this if (object[keys[key] == null) ?!?

servicebroker/lib/util/validateUtil.js
```diff
@@ -5,17 +5,17 @@ module.exports = {
       if (keys != null) {
         if (keys instanceof Array) {
           for (var key in keys) {
-            if (typeof (object[keys[key]]) === "undefined" || object[keys[key]] === null) {
-              return false;
+            if (object[keys[key]] == null) {
+              return { valid: false, message: "The value of " + keys[key] + " should not be undefined or null" };
```
fraenkel: Rather than go round and round, I have created a chore to clean up the validation.

servicebroker/lib/index.js
```diff
@@ -10,7 +10,9 @@ var logger = require(path.join(__dirname, './logger/logger.js'));
 var validateUtil = require(path.join(__dirname, './util/validateUtil.js'));
 var settings = require(path.join(__dirname, './config/setting.js'))((JSON.parse(
   fs.readFileSync(path.join(__dirname, '../config/settings.json'), 'utf8'))));
-if (validateUtil.validate(settings, ['port', 'username', 'password', 'db', 'apiServerUri']) === false) {
+var validateResult = validateUtil.validate(settings, ['port', 'username', 'password', 'db', 'apiServerUri']);
+if (validateResult.valid === false) {
+  logger.error("Setting validate error," + validateResult.message);
   throw new TypeError('setting.json is invalid');
 }
```
fraenkel: Where is the test for invalid settings?

fraenkel:
Please take a look at the changes I made.
It is now merged.

qibobo:
Thank you!

PR #50, opened 2016-07-12
Validate Autoscaler Policy

pradyutsarma:
1. Validate Autoscale policy request
2. Remove unnecessary logs from tests
3. Refactor tests

Pending items (from last PR's comments):
1. Tests still execute with sqlite. Will create a separate PR to support pg for tests too.
2. Validation of App id (whether the ID passed is a valid app id) is still not done.


api/app.js
```diff
@@ -2,12 +2,25 @@ var express = require('express');
...
+app.use(function(err, req, res, next) {
+  var errorResponse = {};
+  if (err) {
+    errorResponse = {
+      'success': false,
+      'error': err,
+      'result': null
+    };
+  }
+  res.status(HttpStatus.BAD_REQUEST).send(errorResponse);
```
fraenkel: How do you know that a BAD_REQUEST is good for all types of errors?

pradyutsarma: @fraenkel We are currently considering all validation failures as BAD REQUEST. What is your proposal? We could use 403 or 422 too. Do we differentiate between structural and data errors?

api/lib/validation/attributeValidator.js
```diff
@@ -0,0 +1,230 @@
...
+        if(intersect(inputRecurringSchedules[i].days_of_week,
+         inputRecurringSchedules[j].days_of_week).length > 0) {
+          if(inputRecurringSchedules[i].start_time <=
+          inputRecurringSchedules[j].end_time && 
+          inputRecurringSchedules[i].end_time >=
+          inputRecurringSchedules[j].start_time) {
+            let error = createErrorResponse('recurring_schedule.start_time',
+            'recurring_schedule.start_time and recurring_schedule.end_time ' + 
+            'ranges are overlapping',inputRecurringSchedules[j],
+            'Days_of_week based time range of recurring_schedule[' + j + ']' +
```
fraenkel: I will need to see how the error appears in the log to determine if this will work. Need to determine if there will be enough contextual information.

api/lib/validation/attributeValidator.js
```diff
@@ -0,0 +1,230 @@
...
+    .isAfter(moment(endDateTime,'YYYY-MM-DDTHH:mm'))) {
+    result = false;
+  }
+  return result;
+}
+
+var validateDaysOfWeekErrors = function(inputRecurringSchedules) {
+  var daysOfWeekErrors = [];
+  var errorCount = 0;
+  for(let j = 0; j < inputRecurringSchedules.length; j++) {
```
fraenkel: Shouldn't this be iRS.length-1?

api/lib/validation/attributeValidator.js
```diff
@@ -0,0 +1,230 @@
...
+          }
+        }
+      }
+    }
+  }
+  return daysOfWeekErrors;
+}
+
+
+var validateDaysOfMonthErrors = function(inputRecurringSchedules) {
```
fraenkel: Would be nice if we could DRY this up with the above.

api/lib/validation/attributeValidator.js
pradyutsarma: That's a good point. We also wanted to get your feedback on the error response that we create. An example of a validation error is this (in case the metric_type is misspelt). The response is loosely based on the response created by the schema validator that we use https://www.npmjs.com/package/jsonschema
```
{
    "success": false,
    "error": [
        {
            "property": "instance.scaling_rules[0].metric_type",
            "message": "is not one of enum values: MemoryUsage",
            "schema": {
                "type": "string",
                "enum": [
                    "MemoryUsage"
                ]
            },
            "instance": "MemoryUsagde",
            "name": "enum",
            "argument": [
                "MemoryUsage"
            ],
            "stack": "instance.scaling_rules[0].metric_type is not one of enum values: MemoryUsage"
        }
    ],
    "result": null
}
```

api/lib/validation/attributeValidator.js
```diff
@@ -0,0 +1,230 @@
...
+  }
+  
+  return errors;
+}
+
+
+
+exports.validatePolicy = function(inputPolicy,callback) {
+  if(callback) {
+    var result = validatePolicyJSONValues(inputPolicy);
```
fraenkel: isn't this really errors?

pradyutsarma: 'result' could either be an empty array (means everything's fine) or an array of errors to be passed back via the callback.

api/lib/validation/schemaValidator.js
```diff
@@ -0,0 +1,183 @@
+'use-strict'
```
fraenkel: We forgot the initial_instance_count which is optional and allows you to select something in the range between min and max when a schedule triggers.

api/lib/validation/validationMiddleware.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+var logger = require('../log/logger');
+var _ = require('underscore');
+var schemaStructureValidator = require('./schemaValidator');
+var attributeValidator = require('./attributeValidator');
+module.exports = function(req, res, next) {  
+  schemaStructureValidator.validatePolicy(req.body, function(validationResult) {
+    if(!_.isEmpty(validationResult)) {
+      logger.error('Input policy json schema is not valid', validationResult);
```
fraenkel: Need more contextual info so we can correlate it to the request, like the app id?

api/lib/validation/validationMiddleware.js
```diff
@@ -0,0 +1,27 @@
...
+      next (validationResult);
+    }
+    else{
+      logger.log('Input Policy JSON schema structure is valid');
```
fraenkel: Is this useful? If we keep it, then we need the app id.

api/lib/validation/schemaValidator.js
pradyutsarma: Why do we need initial_instance_count? Wouldn't apps already provide this via their deployment manifest?

api/lib/validation/validationMiddleware.js
pradyutsarma: Yes. Will handle this in the next PR.

api/lib/validation/validationMiddleware.js
pradyutsarma: Agreed.

api/test/unit/routes/policies.test.js
```diff
@@ -2,91 +2,51 @@
...
-  after(function() {
-    this.Policy.truncate().then(function(result) {
-      logger.warn('Truncating the test db before test run !!!');
-    });  
+  it('should fail to create another policy for app id 12345', function(done) {
+    request(app)
+    .put('/v1/apps/12345/policy')
+    .send(fakePolicy)
+    .end(function(error,result) {
+      expect(result.statusCode).to.equal(400);
```
fraenkel: Should this be a 400 or 409?

pradyutsarma: Yes. It could be 409 but then, once we have "update-policy" in place, this use case would be redundant. An existing app id would be updated in that case.

api/test/unit/validation/attributeValidator.test.js
```diff
@@ -0,0 +1,101 @@
...
+    it('Should validate the policy JSON successfully',function(){
+      attributeValidator.validatePolicy(fakePolicy,function(result){
+        expect(result).to.be.empty;
+      });
+    });
+    it('Should fail to validate the policy due to wrong Instance count',function(){
+      fakePolicy.instance_min_count = 10;
+      fakePolicy.instance_max_count = 5;
+      attributeValidator.validatePolicy(fakePolicy,function(result){
+        expect(result).to.not.be.empty;
```
fraenkel: We can combine the 2 lines with
expect(result).to.have.members('instance....')

api/test/unit/validation/attributeValidator.test.js
```diff
@@ -0,0 +1,101 @@
...
+    beforeEach(function(){
+      fakePolicy = JSON.parse(fs.readFileSync(__dirname+'/../fakePolicy.json', 'utf8'));
+    });
+    
+    it('Should validate the policy JSON successfully',function(){
+      attributeValidator.validatePolicy(fakePolicy,function(result){
+        expect(result).to.be.empty;
+      });
+    });
+    it('Should fail to validate the policy due to wrong Instance count',function(){
```
fraenkel: this is really min > max, not wrong instance count.

api/test/unit/validation/attributeValidator.test.js
pradyutsarma: Agreed.

api/test/unit/validation/attributeValidator.test.js
pradyutsarma: That's what the 'expect check' says but the 'it' doesn't. Will fix in the next PR.

api/test/unit/validation/schemaValidator.test.js
```diff
@@ -0,0 +1,106 @@
+'use strict';
+
+var expect = require("chai").expect;
```
fraenkel: This doesn't really test anything since it is more a recoding of the schema.
We should be testing the schema validation not reaching through to see how it is defined.

pradyutsarma: That's right. We debated before we decided that we need this test to guard against anyone changing anything in the methods which may impact the policy validation indirectly.

api/app.js
fraenkel: It is just weird to do a get /foo/bar and receive a 400 instead of a 404 or something more common.

pradyutsarma: Good point. I will modify this to handle GET better by checking the req type.

api/lib/validation/attributeValidator.js
fraenkel: We will go with what is there and see how it plays out.

api/lib/validation/attributeValidator.js
fraenkel: yes but they represent errors if they are there.

api/lib/validation/attributeValidator.js
pradyutsarma: Yes. Will fix in next PR.

api/lib/validation/schemaValidator.js
fraenkel: This was to handle the case where black friday started and you wanted a range from 1 - 10 but you knew you needed to start at nothing less than 5.
Its really initial_min_instance_count.

api/test/unit/routes/policies.test.js
fraenkel: Ok, won't worry about it now since it will go away with update.

api/test/unit/validation/attributeValidator.test.js
```diff
@@ -0,0 +1,101 @@
...
+      fakePolicy.schedules.recurring_schedule[1].end_time = '23:15';
+      attributeValidator.validatePolicy(fakePolicy,function(result){
+        expect(result).to.be.empty;
+      });
+    });
+    it('should fail to validate the policy if start_time is after end_time in recurring schedule',function(){
+      fakePolicy.schedules.recurring_schedule[1].start_time = '23:00'
+      fakePolicy.schedules.recurring_schedule[1].end_time = '13:15';
+      attributeValidator.validatePolicy(fakePolicy,function(result){
+        expect(result).to.not.be.null;
```
pradyutsarma: Why are we checking for NOT NULL in this test and not in the other tests? Anyways, do we need this check?

api/test/unit/validation/schemaValidator.test.js
pradyutsarma: @fraenkel As of now, we will keep this test.

pradyutsarma:
@fraenkel The updated PR contains,
1. Review comments incorporated.
2. DB Connection pool options being read from env or otherwise defaulted
3. Tests run on Postgresql (README updated)

fraenkel:
I have merged this will some updates of my own. See the commits that follow.

pradyutsarma:
@fraenkel I see that you've modified the tests to assume that the policy_json table also exists before the tests are run. Wouldn't it be easier to create the table via the tests using `sequelize#sync` and then execute the tests on top. In that case, the amount of work required to setup and run the tests is restricted to just the user providing the DB_URI.

fraenkel:
No.
It validates the liquibase db definitions. 
Some of the tests need the other defined. This is why I have kept saying to truncate the dbs. It's created once for all and reused.

Read the top-level readme.

pradyutsarma:
@fraenkel  I've read the README and I get your point. However, I've always imagined unit tests to be self-contained. Somehow, in the above manner, makes me feel we are breaking into the boundaries of integration tests.

fraenkel:
It is no different than requiring the database to be started and created. There tables are not part of your code but used by your code. If sequelize could do migrations I would have had a separate command top create and migrate the tables.

PR #55, opened 2016-07-25
auto-refresh access token for metrics-collector

boyang9527:
[#121995347]

- refactor cf-client
- add RefreshAuthToken in cf-client
- set RefreshTokenFrom when initiating noaa
- add token refresh grant in config
- regenerate fake_cf_client due to interface change
- unit test for cf_client token refresh
- modified unit test for main.go

### DIFF
--- a/src/metricscollector/cf/cf_client_test.go
+++ b/src/metricscollector/cf/cf_client_test.go
@@ -47,37 +47,50 @@ var authBody = []byte(`
 }
 `)
 
+var refreshBody = []byte(`
+{
+	"access_token":"test-access-token-refreshed",
+	"token_type":"bearer",
+	"refresh_token":"test-refresh-token-refreshed",
+	"expires_in":24000,
+	"scope":"openid cloud_controller.read password.write cloud_controller.write",
+	"jti":"a735f90f-0b49-447d-8f9d-ae2fbc1491dd"
+}
+`)
+
 var _ = Describe("CfClient", func() {
+	var (
+		fakeCC          *ghttp.Server
+		fakeLoginServer *ghttp.Server
+		cfc             CfClient
+		conf            *config.Config
+		authToken       string
+		err             error
+	)
+
+	BeforeEach(func() {
+		fakeCC = ghttp.NewServer()
+		fakeLoginServer = ghttp.NewServer()
+		conf = &config.Config{}
+
+		conf.Cf.Api = fakeCC.URL()
+		err = nil
+	})
+
+	AfterEach(func() {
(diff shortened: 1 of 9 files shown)


src/metricscollector/cf/cf_client_test.go
```diff
@@ -228,4 +241,163 @@ var _ = Describe("CfClient", func() {
...
+					Expect(cfc.GetTokens().RefreshToken).To(Equal("test-refresh-token"))
+					Expect(cfc.GetTokens().ExpiresIn).To(Equal(int64(12000)))
+				})
+
+			})
+
+		})
+
+		Context("when already logged in", func() {
+			BeforeEach(func() {
```
fraenkel: The setup is wrong.
If we are already logged in, we should not see any request made to PathCFInfo.
It is always best to structure the good path first and then do all the error cases.

src/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
```diff
@@ -64,6 +65,12 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	ccNOAAUAA.RouteToHandler("GET", "/apps/an-app-id/containermetrics",
 		func(rw http.ResponseWriter, r *http.Request) {
+			if isTokenExpired {
+				isTokenExpired = false
```
fraenkel: This seems unnecessary and a bit weird to reset during a test.

src/metricscollector/cf/cf_client_test.go
boyang9527: this is used for cfc.login(), as this invocation needs requests to PathCFInfo

boyang9527: I will adjust the order to have a good path first

src/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
boyang9527: if token is expired,  NOAA will firstly refresh the token, and then call loggregator again with the refreshed token. If isTokenExpired is not reset to false, it will return 401 for second call to loggregator.

src/metricscollector/cf/cf_client_test.go
fraenkel: I know why you have it but it is not too beer invoked in all cases which you are not guaranteeing

boyang9527:
I changed the order to go with good path first. 
Also please see my explanation of your comments.

fraenkel:
Merged

PR #56, opened 2016-07-25
126002173 : I can schedule based on a policy - Recurring Schedules - Modifications to support recurring schedule.

ghaih:
Removed the schedule type column from table since there will be separate tables for specific date and recurring schedule. Separated the entity and dao for specific date schedule which resulted modifications in dependent classes.


fraenkel:
Please add the story number to your commit next time.

PR #57, opened 2016-07-27
[#126384395] Recurring schedules are persisted

kanekoh:
Added a new table for recurring schedule, new entity and DAO class and modified other classes to support recurring schedule persistence.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
@@ -446,11 +446,8 @@ public class ScheduleManager {
 		if (recurringScheduls != null) {
 			for (RecurringScheduleEntity recurringScheduleEntity : recurringScheduls) {
 				// Persist the schedule in database
-				RecurringScheduleEntity savedScheduleEntity = saveNewRecurringSchedule(recurringScheduleEntity);
+				saveNewRecurringSchedule(recurringScheduleEntity);
 
-				// Ask ScalingJobManager to create scaling job
-				if (savedScheduleEntity != null) {
-				}
 			}
 		}
 	}
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
@@ -7,7 +7,9 @@ import java.util.TimeZone;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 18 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -141,10 +159,19 @@ public void validateSchedules(String appId, ApplicationScalingSchedules applicat
...
+			// Validate specific date schedules.
+			if (specificDateSchedules != null) {
+				validateSpecificDateSchedules(specificDateSchedules, isValidTimeZone);
+			}
 
-			validateSpecificDateSchedules(applicationScalingSchedules.getSpecific_date(), isValidTimeZone);
+			List<RecurringScheduleEntity> recurringSchedules = applicationScalingSchedules.getRecurring_schedule();
+			// Validate recurring schedules.
+			if (recurringSchedules != null) {
+				//TODO validate recurring schedule.
```
fraenkel: I don't see any story for validation therefore this needs to be done.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -408,13 +429,28 @@ private SpecificDateScheduleDateTime validateStartEndDateTime(String scheduleBei
...
+		}
+
+		List<RecurringScheduleEntity> recurringScheduls = applicationScalingSchedules.getRecurring_schedule();
+		if (recurringScheduls != null) {
+			for (RecurringScheduleEntity recurringScheduleEntity : recurringScheduls) {
+				// Persist the schedule in database
+				RecurringScheduleEntity savedScheduleEntity = saveNewRecurringSchedule(recurringScheduleEntity);
+
+				// Ask ScalingJobManager to create scaling job
+				if (savedScheduleEntity != null) {
```
fraenkel: What is supposed to be here?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
kanekoh: Will create a new story for validation of recurring schedule.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
kanekoh: Be supposed to implement to create cron job with Quartz on the story 'When recurring rules are created, associated jobs are created'.
Will remove the 'if' statement.

fraenkel:
Merged

PR #59, opened 2016-07-28
Policy CRUD operations in api server

pradyutsarma:
1. Support for PUT, GET and DELETE policy
2. Requisite tests


api/lib/routes/policies.js
```diff
@@ -6,26 +6,98 @@ var HttpStatus = require('http-status-codes');
...
+          'error': null,
+          'result': result
+        };
+        res.set('Location', '/v1/apps/' + req.params.app_id + '/policy');
+        res.status(HttpStatus.CREATED).json(successResponse);
+      }
+      else{
+        logger.info('Updating the existing policy for the application with app id : ' +
+            req.params.app_id);
+        models.policy_json.update({
```
fraenkel: Shouldn't we be doing this with the above in a transaction?

fraenkel:
I have been playing around with a different style of logging to make things a bit easier to parse.
Something to consider is the following:

logger.error ('Failed to delete policy for the application with app id : ' + 
        req.params.app_id, error);
becomes
logger.error ('Failed to delete policy", { "app id": req.params.app_id, "err": err});

fraenkel:
I did not merge the add npm-shrinkwrap commit since it was unnecessary. Please make sure you are rebasing your branch before you commit.

pradyutsarma:
@fraenkel It would be great to have this JSON logging for all our app. This is what I did discuss with you in one of the initial sync meetings. Libs like winston and sl4j already have support for this format. I will create a story to add this support for api server and see how it goes. That's ok?

fraenkel:
It works with console.log. But feel free to investigate options. I only want to pull in a logging package if there is some real value add.

PR #60, opened 2016-07-29
watch app changes in policy db and start/stop collecting metrics of t…

boyang9527:
…he apps

[#119505033]

- changed the db.GetAppIds to return a map  instead of array
- add configs for collector
- add apppoller to poll app metrics periodically
- add collector to refresh apps, and start/stop app polling
- test cases for apppoller and collector

The following is not included, will be addressed in next user story

- share auto-generated fakes( cf_client, noaa, and db)


src/metricscollector/collector/apppoller.go
```diff
@@ -0,0 +1,83 @@
...
+	ap.ticker = time.NewTicker(ap.interval)
+	go ap.startPollMetrics()
+
+	ap.logger.Info("app-poller-started", lager.Data{"appid": ap.appId})
+}
+
+func (ap *AppPoller) Stop() {
+	if ap.ticker != nil {
+		ap.ticker.Stop()
+		ap.doneChan <- true
```
fraenkel: Normally, you would close the channel to signify its done rather than send a value.
The other issue is that you made it a non-buffered channel which means the go routine will wait here until something pulls it off. While it will work, it does cause a delay.
Either switch to close(app.doneChan) or make the channel buffered with 1 element.

src/metricscollector/collector/apppoller.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	ap.logger.Info("app-poller-stopped", lager.Data{"appid": ap.appId})
+}
+
+func (ap *AppPoller) startPollMetrics() {
+	ap.pollMetric()
```
fraenkel: Move this into the for-loop. Remove the one called on the tick event.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+		database = &fakes.FakeDB{}
+		logger := lager.NewLogger("apppoller-test")
+
+		poller = NewAppPoller(logger, "test-app-id", TestPollInterval*time.Second, cfc, noaa, database)
+	})
+
+	Describe("Start", func() {
+
+		JustBeforeEach(func() {
+			poller.Start()
```
fraenkel: This is wrong.
You start the poller here.
You stop the poller in the AfterEach.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+								ApplicationId: proto.String("test-app-id"),
+								InstanceIndex: proto.Int32(0),
+								MemoryBytes:   proto.Uint64(1234),
+							},
+						}, nil
+					}
+				})
+
+				It("saves all the metrics to database", func() {
+					Expect(database.SaveMetricCallCount()).To(Equal(3))
```
fraenkel: Switch to a
Eventually(database.SaveMetricCallCount).Should(Equal(3))

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+					noaa.ContainerMetricsStub = func(appid string, token string) ([]*events.ContainerMetric, error) {
+						Expect(appid).To(Equal("test-app-id"))
+						Expect(token).To(Equal("bearer test-access-token"))
+
+						return []*events.ContainerMetric{}, nil
+					}
+				})
+
+				It("saves nothings to database", func() {
+					Expect(database.SaveMetricCallCount()).To(BeZero())
```
fraenkel: This would be a
Consistently(database.SaveMetricCallCount).Should(BeZero())

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+									MemoryBytes:   proto.Uint64(1234),
+								},
+							}, nil
+						}
+
+					}
+				})
+
+				It("saves non-empty metrics to database", func() {
+					Expect(database.SaveMetricCallCount()).To(Equal(2))
```
fraenkel: Eventually

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+		})
+
+		Context("when retrieving container metrics all fails", func() {
+
+			BeforeEach(func() {
+				noaa.ContainerMetricsReturns(nil, errors.New("an error"))
+			})
+
+			It("saves nothing to database", func() {
+				Expect(database.SaveMetricCallCount()).To(Equal(0))
```
fraenkel: Consistently

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	Describe("Stop", func() {
+		JustBeforeEach(func() {
+			poller.Stop()
+		})
+
+		Context("when apppoller is started", func() {
+			BeforeEach(func() {
+				poller.Start()
+				time.Sleep(2500 * time.Millisecond)
```
fraenkel: No sleep.

src/metricscollector/collector/collector_test.go
```diff
@@ -0,0 +1,212 @@
...
+		database = &fakes.FakeDB{}
+		logger := lager.NewLogger("collector-test")
+		conf := &config.CollectorConfig{RefreshInterval: TestRefreshInterval, PollInterval: TestPollInterval}
+		coll = NewCollector(logger, conf, cfc, noaa, database)
+	})
+
+	Describe("Start", func() {
+		JustBeforeEach(func() {
+			coll.Start()
+			time.Sleep(duration)
```
fraenkel: Same as above.

src/metricscollector/collector/apppoller_test.go
boyang9527: if we want to inject, we will need to make the ticker as a argument when creating apppoller.  This looks a little bit wired since it is an internal thing. Did I misunderstand something?

"Otherwise you need to just add a Consistently(noaa.ContainerMetrics) is not invoked"
you mean stop poller just after start?   once the poller started,  ap.pollMetric() will be invoked before getting the stop signal

src/metricscollector/collector/apppoller_test.go
```diff
@@ -0,0 +1,191 @@
...
+				num := noaa.ContainerMetricsCallCount()
+				time.Sleep(1500 * time.Millisecond)
+				Expect(noaa.ContainerMetricsCallCount()).To(Equal(num))
+			})
+		})
+
+		Context("when apppoller is not started", func() {
+			It("does nothing", func() {
+				Expect(noaa.ContainerMetricsCallCount()).To(BeZero())
+			})
```
boyang9527: removed

src/metricscollector/collector/apppoller.go
boyang9527: changed to close(app.doneChan)

src/metricscollector/collector/apppoller.go
boyang9527: Here is the change. Move the beginning of the loop to make sure we don't need a cycle to get the metrics polled.

```
	for {
		ap.pollMetric()
		select {
		case <-ap.doneChan:
			return
		case <-ap.ticker.C:
		}
	}

```

src/metricscollector/collector/apppoller_test.go
boyang9527: I changed to the following:
```
	Describe("Stop", func() {
		JustBeforeEach(func() {
			poller.Stop()
		})

		Context("when apppoller is started", func() {
			BeforeEach(func() {
				poller.Start()
			})

			It("stops the polling", func() {
				num := noaa.ContainerMetricsCallCount()
				Consistently(noaa.ContainerMetricsCallCount(), 1500*time.Millisecond, 500*time.Millisecond).Should(Equal(num))
			})
		})

```

fraenkel: Still wrong. It makes little sense. You have no guarantee the loop ever ran or will run.
Normally, you would start, wait for a metricscallcount == 1 to verify the poller is running and then stop it and make sure it never happens. Again it is a weak test since it is time dependent which means you should also adjust the poll interval to make it happen very often like 5ms or 10ms.

fraenkel: With a FakeClock from code.cloudfoundry.org/fakeclock, you create a faketicker from it. The FakeClock gives you precise control of time so you can cause a tick whenever you want. You can easily prove that when you trigger a tick, the CallCount goes up and after a Stop, you can also prove that it doesn't.

src/metricscollector/collector/apppoller_test.go
boyang9527: changed to use Eventually and Constently

src/metricscollector/collector/apppoller_test.go
boyang9527: the problem of this test case is it only tests stop without ticker triggered polling. 
I will change the collector and app poller impl to be able to inject ticker.

boyang9527:
@fraenkel  
I addressed your comments in the new commit.  Please review.

src/metricscollector/collector/apppoller.go
```diff
@@ -5,56 +5,59 @@ import (
...
 
 func (ap *AppPoller) startPollMetrics() {
-	ap.pollMetric()
 	for {
 		select {
 		case <-ap.doneChan:
 			return
-		case <-ap.ticker.C:
+		case <-ap.ticker.C():
 			ap.pollMetric()
```
fraenkel: This needs to move above in the for loop otherwise you aren't polling the first time.

src/metricscollector/collector/apppoller.go
```diff
@@ -5,56 +5,59 @@ import (
...
+	cfc          cf.CfClient
+	noaa         NoaaConsumer
+	database     db.DB
+	pclock       clock.Clock
+	ticker       clock.Ticker
+	doneChan     chan bool
 }
 
-func NewAppPoller(logger lager.Logger, appId string, interval time.Duration, cfc cf.CfClient, noaa NoaaConsumer, database db.DB) *AppPoller {
+func NewAppPoller(appId string, pollInterval int, logger lager.Logger, cfc cf.CfClient, noaa NoaaConsumer, database db.DB, pclcok clock.Clock) *AppPoller {
```
fraenkel: The interval should be a duration.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -118,19 +141,33 @@ var _ = Describe("Apppoller", func() {
 				})
 
 				It("saves non-empty metrics to database", func() {
-					Expect(database.SaveMetricCallCount()).To(Equal(2))
+					fclock.Increment(TestPollInterval * time.Second)
```
fraenkel: TestPollInterval should be a duration.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -118,19 +141,33 @@ var _ = Describe("Apppoller", func() {
...
-				noaa.ContainerMetricsReturns(nil, errors.New("an error"))
+				noaa.ContainerMetricsReturns(nil, errors.New("test apppoller error"))
 			})
 
-			It("saves nothing to database", func() {
-				Expect(database.SaveMetricCallCount()).To(Equal(0))
+			It("saves nothing to database and logs the errors", func() {
+				fclock.Increment(TestPollInterval * time.Second)
+				Consistently(database.SaveMetricCallCount).Should(BeZero())
+				Eventually(buffer).Should(gbytes.Say("test apppoller error"))
```
fraenkel: Checking for the error is insufficient since you have 3 places that could log it. You need to look for poll-metric-from-noaa.
At which point the consistently could be an Expect instead.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -154,38 +191,39 @@ var _ = Describe("Apppoller", func() {
...
 
-			It("saves successful results to database", func() {
-				Expect(database.SaveMetricCallCount()).To(Equal(2))
+			It("saves successful results to database and logs the errors", func() {
+				fclock.Increment(TestPollInterval * time.Second)
+				Eventually(database.SaveMetricCallCount).Should(Equal(1))
+
+				fclock.Increment(TestPollInterval * time.Second)
+				Consistently(database.SaveMetricCallCount).Should(Equal(1))
+				Eventually(buffer).Should(gbytes.Say("apppoller test error"))
```
fraenkel: This is unneeded since you are tracking save call counts and we have proven that it saves on regular intervals.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -154,38 +191,39 @@ var _ = Describe("Apppoller", func() {
...
+			fclock.Increment(TestPollInterval * time.Second)
+			Eventually(noaa.ContainerMetricsCallCount).Should(Equal(1))
 
-			It("stops the polling", func() {
-				num := noaa.ContainerMetricsCallCount()
-				time.Sleep(1500 * time.Millisecond)
-				Expect(noaa.ContainerMetricsCallCount()).To(Equal(num))
-			})
-		})
+			fclock.Increment(TestPollInterval * time.Second)
```
fraenkel: Don't need to do it again.

src/metricscollector/collector/collector.go
```diff
@@ -17,64 +18,64 @@ type NoaaConsumer interface {
...
+		cclock:   cclock,
 		doneChan: make(chan bool),
 		pollers:  make(map[string]*AppPoller),
 		lock:     &sync.Mutex{},
 	}
 }
 
 func (c *Collector) Start() {
-	c.tick = time.NewTicker(time.Duration(c.conf.RefreshInterval) * time.Second)
+	c.ticker = c.cclock.NewTicker(time.Duration(c.conf.RefreshInterval) * time.Second)
```
fraenkel: The config should make the RefreshInterval a duration. We can expose it as a duration or number of seconds in BOSH later.

src/metricscollector/collector/collector.go
```diff
@@ -17,64 +18,64 @@ type NoaaConsumer interface {
...
 
 func (c *Collector) startAppRefresh() {
-	c.refreshApps()
 	for {
 		select {
 		case <-c.doneChan:
 			return
-		case <-c.tick.C:
+		case <-c.ticker.C():
 			c.refreshApps()
```
fraenkel: Same issue as before. We don't refresh immediately.

src/metricscollector/collector/collector_suite_test.go
```diff
@@ -11,9 +11,10 @@ import (
 //go:generate counterfeiter -o ./fakes/fake_noaa_consumer.go . NoaaConsumer
 //go:generate counterfeiter -o ./fakes/fake_DB.go ../db DB
 
+// make sure TestPollInterval is less than TestRefreshInterval
 const (
-	TestRefreshInterval = 2
 	TestPollInterval    = 1
+	TestRefreshInterval = 3
```
fraenkel: Make these durations

src/metricscollector/collector/collector_test.go
```diff
@@ -18,67 +19,104 @@ import (
...
 			Context("when no apps in policy database", func() {
 				BeforeEach(func() {
-					duration = 2500 * time.Millisecond
 					database.GetAppIdsReturns(make(map[string]bool), nil)
 				})
 
 				It("does not poll anything", func() {
-					Expect(noaa.ContainerMetricsCallCount()).To(BeZero())
+					fclock.Increment(TestRefreshInterval * time.Second)
+					Eventually(database.GetAppIdsCallCount).Should(Equal(1))
```
fraenkel: We are confusing 2 things here, polling and refreshing.
The collector only knows about one thing, refreshing the set of pollers.
If you want to test whether pollers are created, started and stopped, which would be a good thing, then I would allow a function to be passed in that did the creation of a AppPoller interface.

fraenkel:
Also
```
[1470662447] Collector Suite - 13/13 specs ••==================
WARNING: DATA RACE
Write by goroutine 8:
  metricscollector/collector_test.glob.func2.2.3.4.2()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/collector_test.go:145 +0xc7b
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/leafnodes/runner.go:109 +0xbb
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/leafnodes/runner.go:63 +0xdb
  github.com/onsi/ginkgo/internal/leafnodes.(*ItNode).Run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/leafnodes/it_node.go:25 +0x77
  github.com/onsi/ginkgo/internal/spec.(*Spec).runSample()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/spec/spec.go:167 +0x7b1
  github.com/onsi/ginkgo/internal/spec.(*Spec).Run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/spec/spec.go:118 +0xf4
  github.com/onsi/ginkgo/internal/specrunner.(*SpecRunner).runSpecs()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/specrunner/spec_runner.go:144 +0x2f2
  github.com/onsi/ginkgo/internal/specrunner.(*SpecRunner).Run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/specrunner/spec_runner.go:61 +0xbd
  github.com/onsi/ginkgo/internal/suite.(*Suite).Run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/suite/suite.go:59 +0x34f
  github.com/onsi/ginkgo.RunSpecsWithCustomReporters()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/ginkgo_dsl.go:207 +0x3c7
  github.com/onsi/ginkgo.RunSpecs()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/ginkgo_dsl.go:188 +0x41a
  metricscollector/collector_test.TestCollector()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/collector_suite_test.go:22 +0xa3
  testing.tRunner()
      /usr/local/Cellar/go/1.6.3/libexec/src/testing/testing.go:473 +0xdc

Previous read by goroutine 22:
  metricscollector/collector_test.glob.func2.2.3.1.1()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/collector_test.go:63 +0xb6
  metricscollector/collector/fakes.(*FakeNoaaConsumer).ContainerMetrics()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/fakes/fake_noaa_consumer.go:35 +0x465
  metricscollector/collector.(*AppPoller).pollMetric()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/apppoller.go:69 +0x353
  metricscollector/collector.(*AppPoller).startPollMetrics()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/collector/apppoller.go:61 +0x140
```

src/metricscollector/collector/apppoller.go
boyang9527: I moved back to ticker so that there is a late polling so that the test cases are more clearer. If we care immediate polling, then I will change back.

src/metricscollector/collector/apppoller.go
boyang9527: sure, will change to Duration.

src/metricscollector/collector/apppoller_test.go
boyang9527: sure, will change accordingly.

src/metricscollector/collector/apppoller_test.go
boyang9527: Right... will change

src/metricscollector/collector/apppoller_test.go
boyang9527: ok.. will remove it.

boyang9527:
The data races are introduced when different go routines(ginkgo and poller) are reading/writing slice pollings []bool.  I ran the test cases with --race. Though there are data races, it returns correct result, it should not be harmful. 

If we change the apppoller to an interface, and only test refresh, then we don't need the slice to record whether the pollings happen or not, the data races will be better. But there are still some, especially when we are using eventually/consistently to check the count of the method invocation.

fraenkel:
In the end, you should never have data races. They will prove to be flakey and fail every now and then. Regardless whether it is code or tests, data races should never exist.

src/metricscollector/collector/apppoller.go
boyang9527: moved back to beginning of the loop to ensure immediate poll

src/metricscollector/collector/apppoller_test.go
boyang9527: All intervals changed to be a Duration

src/metricscollector/collector/apppoller_test.go
boyang9527: Consistently needs to be after checking log, and will make sure the pollMetric()  returns without db operation if retrieving container metrics fail

src/metricscollector/collector/apppoller_test.go
boyang9527: added test for polling with given interval, and removed polling count in other test cases

src/metricscollector/collector/apppoller_test.go
boyang9527: removed the second ticker triggered polling before stop

src/metricscollector/collector/collector.go
boyang9527: changed

src/metricscollector/collector/collector.go
boyang9527: changed.

src/metricscollector/collector/collector_suite_test.go
boyang9527: done

src/metricscollector/collector/collector_test.go
boyang9527: added AppPoller interface, and passed in poller creation function as argument so that we can inject fake app poller

boyang9527:
removed the races.

src/metricscollector/collector/apppoller_test.go
```diff
@@ -52,21 +52,31 @@ var _ = Describe("Apppoller", func() {
...
+			fclock.Increment(TestPollInterval)
+			Eventually(noaa.ContainerMetricsCallCount).Should(Equal(3))
 		})
 
 		Context("when retrieving container metrics all succeeds", func() {
 
+			BeforeEach(func() {
+				cfc.GetTokensReturns(cf.Tokens{AccessToken: "test-access-token"})
+
+				database.SaveMetricStub = func(metric *metrics.Metric) error {
```
fraenkel: Normally you wouldn't do it this way since if SaveMetrics is never called these checks don't happen.
You would use the SaveMetricArgsForCall to retrieve the parameters and check in the test.

src/metricscollector/collector/collector.go
```diff
@@ -18,53 +17,60 @@ type NoaaConsumer interface {
...
+	database     db.DB
+	cclock       clock.Clock
+	createPoller func(string, time.Duration, lager.Logger, cf.CfClient, NoaaConsumer, db.DB, clock.Clock) AppPoller
+	doneChan     chan bool
+	pollers      map[string]AppPoller
+	ticker       clock.Ticker
+	lock         *sync.Mutex
+}
+
+var createAppPollerFunc = func(appId string, pollInterval time.Duration, logger lager.Logger, cfc cf.CfClient, noaa NoaaConsumer, database db.DB, pclcok clock.Clock) AppPoller {
```
fraenkel: Would have just called it createAppPoller.
You can always pass around a function by its name.

fraenkel:
Squashed & Merged

PR #63, opened 2016-08-02
[#127046799] - Specific Date schedule should have the start date and time merged into a single property, also the end date and time 

ghaih:
Schema changes and property specific changes in Service, Test classes etc.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Calendar;
 import java.util.Collections;
+import java.util.Date;
 import java.util.List;
 import java.util.TimeZone;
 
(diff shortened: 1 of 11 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
```diff
@@ -55,32 +56,37 @@ public static boolean isValidTimeZone(String timeZoneId) {
...
-	 * 
 	 * @return
 	 */
-	public static boolean isLaterThanNow(Long timeInMillis, TimeZone timeZone) {
-		if (timeInMillis != null && timeZone != null) {
-			Calendar calendar = Calendar.getInstance(timeZone);
-			Long nowInMillis = calendar.getTimeInMillis();
-
-			return timeInMillis > nowInMillis;
+	public static boolean isLaterThanNow(Date dateTime, TimeZone timeZone) {
```
fraenkel: Just a general comment for future reference.

Your naming is a bit mixed. Here you use isLaterThanNow and below you use isAfter. Shouldn't this be isAfterNow?

Why don't you just do: return calToCompare.after(calNow); instead of the if block?

ghaih: Thanks for the heads up Michael.
Yes you are right. In the process of refactoring missed these points , will fix them in next refactoring.

Regards,

Harpreet Ghai
[cid:image001.jpg@01D1EE23.7E9A9120]
[cid:image002.jpg@01D1EE23.7E9A9120]

PR #65, opened 2016-08-04
Policies are periodically retrieved by data aggregator

qibobo:
[#127058171]


src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -0,0 +1,77 @@
...
+	logger     lager.Logger
+	interval   int
+	database   db.DB
+	tick       *time.Ticker
+	doneChan   chan bool
+	PolicyMap  map[string]*policy.PolicyJson
+	TriggerMap map[string]*policy.Trigger
+}
+
+func NewPolicyPoller(logger lager.Logger, interval int, database db.DB) *PolicyPoller {
```
fraenkel: interval should be a time.Duration

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -0,0 +1,77 @@
...
+	return &PolicyPoller{
+		logger:     logger,
```
fraenkel: I would have created a new logger.WithSession("policy-poller") so we know what this is logging.

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -0,0 +1,77 @@
...
+			return
+		case <-p.tick.C:
+		}
+	}
+}
+
+func (p *PolicyPoller) retrievePolicy() {
+	policies, err := p.database.RetrievePolicies()
+	if err != nil {
+		p.logger.Error("refresh-app", err)
```
fraenkel: Shouldn't this be retrieve-policies?

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -0,0 +1,77 @@
...
+		return
+	}
+	p.logger.Info("policy count", lager.Data{"count": len(policies)})
+	for _, policyRow := range policies {
+		appId := policyRow.AppId
+		storedPolicyJson, ok := p.PolicyMap[policyRow.AppId]
+		if ok {
```
fraenkel: This is highly redundant. Just determine if things need to be updated and have one block.

src/dataaggregator/aggregator/policyPoller_test.go
```diff
@@ -0,0 +1,96 @@
...
+			poller.Stop()
+		})
+		Context("when the poller is started", func() {
+			BeforeEach(func() {
+				database.RetrievePoliciesStub = func() ([]*PolicyJson, error) {
+					return []*PolicyJson{&PolicyJson{AppId: testAppId1, PolicyStr: policyStr1}}, nil
+				}
+			})
+			It("should retrieve policies for every interval", func() {
+				Eventually(database.RetrievePoliciesCallCount, 2500*time.Millisecond, 500*time.Millisecond).Should(Equal(2))
```
fraenkel: Once you switch to use a Duration, this can be simplified to use the defaults values.

src/dataaggregator/aggregator/policyPoller_test.go
```diff
@@ -0,0 +1,96 @@
...
+			})
+		})
+		Context("when there is no policy", func() {
+			BeforeEach(func() {
+				database.RetrievePoliciesStub = func() ([]*PolicyJson, error) {
+					return []*PolicyJson{}, nil
+				}
+			})
+			It("no trigger generated", func() {
+				Consistently(func() int { return len(poller.PolicyMap) }, 2500*time.Millisecond, 500*time.Millisecond).Should(Equal(0))
```
fraenkel: I am going to let this slide for now, but I currently don't see how the poller is going to be used.
I would have expected that something is triggered after a poll which isn't the case.

src/dataaggregator/config/config.go
```diff
@@ -0,0 +1,66 @@
...
+type LoggingConfig struct {
+	Level string `yaml:"level"`
+}
+
+var defaultLoggingConfig = LoggingConfig{
+	Level: DefaultLoggingLevel,
+}
+
+type PolicyPollerConfig struct {
+	PollInterval int `yaml:"poll_interval"`
```
fraenkel: time.Duration

src/dataaggregator/config/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+var defaultPolicyPollerConfig = PolicyPollerConfig{
+	PollInterval: DefaultPollInterval,
+}
+
+type DbConfig struct {
+	PolicyDbUrl string `yaml:"policy_db_url"`
+}
+
+type Config struct {
```
fraenkel: Do we need it to be nested? If we only have 1 config item per, just pull them up. Do the grouping once we have more than a few.

src/dataaggregator/config/config.go
```diff
@@ -0,0 +1,66 @@
...
+	Poller  PolicyPollerConfig `yaml:"poller"`
+}
+
+func LoadConfig(reader io.Reader) (*Config, error) {
+	conf := &Config{
+		Logging: defaultLoggingConfig,
+		Poller:  defaultPolicyPollerConfig,
+	}
+
+	decoder := candiedyaml.NewDecoder(reader)
```
fraenkel: Please use goyaml. Long story.

src/dataaggregator/config/config.go
```diff
@@ -0,0 +1,66 @@
...
+
+	return conf, nil
+}
+
+func (c *Config) Validate() error {
+
+	if c.Db.PolicyDbUrl == "" {
+		return fmt.Errorf("Configuration error: Policy DB url is empty")
+	}
+
```
fraenkel: What should happen if the interval is 0?

src/dataaggregator/config/config_test.go
```diff
@@ -0,0 +1,121 @@
...
+  poll_interval: "NotInteger"
+`)
+			})
+
+			It("should error", func() {
+				Expect(err).To(MatchError(MatchRegexp("Invalid integer:.*")))
+			})
+		})
+
+		Context("with valid yaml", func() {
```
fraenkel: This should be first... It is hard to find the good case.

src/dataaggregator/db/db.go
```diff
@@ -0,0 +1,10 @@
+package db
+
+import (
+	"dataaggregator/policy"
+)
+
+type DB interface {
```
fraenkel: If this is from metricscollector, then we should pull this out of both to src/db and share across them.

src/dataaggregator/policy/policy.go
```diff
@@ -0,0 +1,53 @@
+package policy
+
+import (
+	"encoding/json"
+)
+
+type IPolicy interface {
```
fraenkel: We are not doing IPolicy. Call it Policy or ScalingPolicy.

src/dataaggregator/policy/policy.go
```diff
@@ -0,0 +1,53 @@
...
+			return true
+		} else {
+			return false
+		}
+	} else {
+		return false
+	}
+}
+func (p *PolicyJson) GetTrigger() *Trigger {
+	appId := p.AppId
```
fraenkel: Why do we have this variable?

src/dataaggregator/policy/policy_test.go
```diff
@@ -0,0 +1,120 @@
...
+		Context("when only one of p1 and p2 are is nil", func() {
+			BeforeEach(func() {
+				p1 = nil
+				p2 = &PolicyJson{AppId: testAppId, PolicyStr: policyStr}
+			})
+			It("should return false", func() {
+				Expect(p1.Equals(p2)).To(Equal(false))
+			})
+		})
+		Context("when p1 and p2 are both not nil", func() {
```
fraenkel: Remove this context.

src/dataaggregator/policy/policy_test.go
```diff
@@ -0,0 +1,120 @@
...
+			Context("when the AppId and PolicyStr are both the same", func() {
```
fraenkel: Split this into Context("when the appIds are the same")

src/dataaggregator/policy/policy_test.go
```diff
@@ -0,0 +1,120 @@
...
+		})
+	})
+	Context("GetTriggerFromPolicy", func() {
+
+		BeforeEach(func() {
+			policyJson = &PolicyJson{AppId: testAppId, PolicyStr: policyStr}
+			trigger = policyJson.GetTrigger()
+		})
+		It("should return a trigger", func() {
+			Expect(trigger.AppId).To(Equal(testAppId))
```
fraenkel: You could just do a Equals(&Trigger {
....
})
rather that each line which could overlook new fields.

src/dataaggregator/aggregator/policyPoller_test.go
qibobo: For this story "Policies are periodically retrieved by data aggregator", we just retrieve policies and parse triggers. We will go to next step to aggregate the memory datas.

Now when I use ginkgo -r -race , it will show some warning msg as below. Seems Ben has the same issue. Could you give some advice?
Warning msg:
WARNING: DATA RACE
Read by goroutine 10:
  dataaggregator/aggregator/fakes.(*FakeDB).RetrievePolicies()
      /Users/qiyang/dev/openjavaspace/app-autoscaler/src/dataaggregator/aggregator/fakes/fake_db.go:33 +0x22d
  dataaggregator/aggregator.(*PolicyPoller).retrievePolicies()
      /Users/qiyang/dev/openjavaspace/app-autoscaler/src/dataaggregator/aggregator/policyPoller.go:59 +0x7c
  dataaggregator/aggregator.(*PolicyPoller).startPolicyRetrieve()
      /Users/qiyang/dev/openjavaspace/app-autoscaler/src/dataaggregator/aggregator/pol

fraenkel: You need to examine the stacks between the WRITE and READ. In most cases it is more a matter of adding sync blocks. I had you add -race because I could see the issue in how you are testing which is why I asked how the Poller is going to be used.

I would personally break this apart into 2 pieces.
1. A poller that calls some operation on a periodic basis that can be stopped.
2. The ability to retrieve policies
3. There is something that is going to be called after retrieving policies which is passed the new set of policies

It becomes easier to test and by having these additional entry points, you can do the testing on the right goroutines.

src/dataaggregator/aggregator/policyPoller_test.go
```diff
@@ -51,46 +53,47 @@ var _ = Describe("PolicyPoller", func() {
...
 	})
 	Context("Stop", func() {
-		JustBeforeEach(func() {
-			poller.Stop()
-		})
-
 		BeforeEach(func() {
 			poller.Start()
+			poller.Stop()
+			clock.Increment(time.Duration(5 * TestPolicyPollerInterval * time.Second))
```
fraenkel: This should be in the test since you are first waiting to see a 1.

src/dataaggregator/config/config.go
```diff
@@ -2,65 +2,44 @@ package config
...
+	"gopkg.in/yaml.v2"
 	"strings"
-
-	"github.com/cloudfoundry-incubator/candiedyaml"
+	"time"
 )
 
 const (
 	DefaultLoggingLevel = "info"
 	DefaultPollInterval = 30
```
fraenkel: This should be a duration too

src/dataaggregator/config/config.go
```diff
@@ -2,65 +2,44 @@ package config
...
+	PollInterval time.Duration `yaml:"poll_interval"`
 }
 
-func LoadConfig(reader io.Reader) (*Config, error) {
+func LoadConfig(bytes []byte) (*Config, error) {
 	conf := &Config{
-		Logging: defaultLoggingConfig,
-		Poller:  defaultPolicyPollerConfig,
+		LogLevel:     DefaultLoggingLevel,
+		PollInterval: DefaultPollInterval,
```
fraenkel: This will be 30ns instead of 30s given your current default.

fraenkel:
Running with ginkgo -r -p -race I received multiple DATA RACE issues.

fraenkel:
Seems I may have confused you on how this would work.
There shouldn't be a GetTriggers func, it is actually the reverse and closer to what you had. The func is what is called after the new Triggers are built and then passed along to the func for further processing.
There should be no need for any locks since the data is only ever created and passed on the same Go routine.

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -43,7 +45,12 @@ func (p *PolicyPoller) Stop() {
 }
 func (p *PolicyPoller) startPolicyRetrieve() {
 	for {
-		p.retrievePolicies()
+		policies, error := p.retrievePolicies()
```
fraenkel: Don't name the variable error since that is a type.

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -52,12 +59,21 @@ func (p *PolicyPoller) startPolicyRetrieve() {
 	}
 }
 
-func (p *PolicyPoller) retrievePolicies() {
+func (p *PolicyPoller) retrievePolicies() ([]*policy.PolicyJson, error) {
 	policies, err := p.database.RetrievePolicies()
 	if err != nil {
 		p.logger.Error("refresh-app", err)
```
fraenkel: Wrong long message

src/dataaggregator/aggregator/policyPoller.go
```diff
@@ -43,7 +45,12 @@ func (p *PolicyPoller) Stop() {
 }
 func (p *PolicyPoller) startPolicyRetrieve() {
 	for {
-		p.retrievePolicies()
+		policies, error := p.retrievePolicies()
+		if error != nil {
+			continue
```
fraenkel: Where is the test for this path?

src/dataaggregator/aggregator/policyPoller_test.go
```diff
@@ -74,48 +66,49 @@ var _ = Describe("PolicyPoller", func() {
...
 	})
 	Context("Stop", func() {
 		BeforeEach(func() {
+			poller = NewPolicyPoller(logger, clock, TestPolicyPollerInterval, database, consumer)
 			poller.Start()
 			poller.Stop()
 		})
 
 		It("stops the polling", func() {
-			clock.Increment(5 * TestPolicyPollerInterval * time.Second)
```
fraenkel: Why did you remove this? Without adjusting time, you haven't proven that Stop really works.

fraenkel:
Looks good.
Just squash these into 1 commit.

qibobo:
One commit now. Thank you very much!

fraenkel:
Merged

PR #66, opened 2016-08-05
127333301 : Add validations for recurring schedules

kanekoh:
The overlap check for the optional properties start_date and end_date has not been done in this story. I will create a separate story for checking the overlap in the start_date and end_date for recurring schedule.


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -226,6 +227,186 @@ private void validateSpecificDateSchedules(List<SpecificDateScheduleEntity> spec
...
+			// it should be after current date.
+			if (!DataValidationHelper.isDateAfterOrEqualsNow(endDate, timeZone)) {
+				isValid = false;
+				validationErrorResult.addFieldError(recurringSchedule, "schedule.date.invalid.before.current",
+						scheduleBeingProcessed, "end_date", DateHelper.convertDateToString(endDate));
+			}
+		}
+
+		if (startDate != null && endDate != null && isValid) {
+			// Is it ok to have startDate= endDate? Does not make sense in case of recurring schedules
```
fraenkel: How/where are we tracking this question?
Why doesn't it make sense if I only need something to occur on a specific date once a year?

kanekoh: Thanks. I understand it. Will remove this comment next PR.

PR #67, opened 2016-08-08
Support for start_date, end_date and initial_min_instance_count

pradyutsarma:
1. Support for end_date and start_date in schedules
2. Support for initial_min_instance_count in schedules
3. Requisite tests

### DIFF
--- a/api/lib/validation/attributeValidator.js
+++ b/api/lib/validation/attributeValidator.js
@@ -18,6 +18,13 @@ var compareTimes = function(startTime,endTime) {
   }
   return result;
 }
+var compareDates = function(startDate,endDate) {
+  var result = true;
+  if(moment(startDate,'YYYY-MM-DD').isAfter(moment(endDate,'YYYY-MM-DD'))) {
+    result = false;
+  }
+  return result;
+}
 
 var compareDateTimes = function(startDateTime,endDateTime) {
   var result = true;
@@ -56,14 +63,14 @@ var validateOverlappingDaysInRecurringSchedules = function(inputRecurringSchedul
 
 var validateDaysOfWeekErrors = function(inputRecurringSchedules) {
   var daysOfWeekErrors = validateOverlappingDaysInRecurringSchedules(inputRecurringSchedules,
-                         'days_of_week');
+  'days_of_week');
   return daysOfWeekErrors;
 }
 
 
 var validateDaysOfMonthErrors = function(inputRecurringSchedules) {
   var daysOfMonthErrors = validateOverlappingDaysInRecurringSchedules(inputRecurringSchedules,
-                          'days_of_month')
+  'days_of_month')
   return daysOfMonthErrors;
 }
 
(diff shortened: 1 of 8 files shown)


api/lib/validation/attributeValidator.js
```diff
@@ -76,36 +83,81 @@ var validateRecurringScheduleAttributeValue = function(inputRecurringSchedules)
...
-       'instance_max_count ' + inputRecurringSchedules[i].instance_max_count +
-       ' in recurring_schedule :[' + i + ']');
+          'recurring_schedule.instance_min_count and recurring_schedule.instance_max_count ' + 
+          'values are not compatible',inputRecurringSchedules[i],'instance_min_count ' + 
+          inputRecurringSchedules[i].instance_min_count + ' is higher or equal to ' + 
+          'instance_max_count ' + inputRecurringSchedules[i].instance_max_count +
+          ' in recurring_schedule :[' + i + ']');
       attributeValueErrors[errorCount++] = error;
     }
     if(!compareTimes(inputRecurringSchedules[i].start_time,
```
fraenkel: This seems broken.
Recurring has many possibilities since we have day of month, day of week and start/end date. We seem to allow all three in a given rule which may be ok, a bit confusing.

I we use start/end date, the check for start/end time is incorrect. If the start/end date are the same, as long as the times are after one another, it should be valid, no?

api/lib/validation/attributeValidator.js
```diff
@@ -18,6 +18,13 @@ var compareTimes = function(startTime,endTime) {
   }
   return result;
 }
+var compareDates = function(startDate,endDate) {
+  var result = true;
+  if(moment(startDate,'YYYY-MM-DD').isAfter(moment(endDate,'YYYY-MM-DD'))) {
+    result = false;
```
fraenkel: Why isn't this returning the result of isAfter?

api/test/unit/validation/validationMiddleware.test.js
```diff
@@ -365,4 +365,209 @@ describe('Validate Policy JSON Schema structure', function() {
       done();
     });
   });
+  it('should successfully validate policy schema without start_date and end_date in recurring_schedule', function(done) {
```
fraenkel: There is a problem with all these tests.
If I change any of the expects which should cause a failure, it passes.

I was also expecting that I could have 0 recurring or specific dates and pass which doesn't seem to be the case.

api/lib/validation/attributeValidator.js
pradyutsarma: @fraenkel We don't allow days_of_week and days_of_month as part of the same schedule. The PUT will fail in that case since the schema validation only allows for one of them in a schedule. However, we allow start_date, end_date to be specified with either of the other two.

Yes, It will be valid. It first checks the time and then the dates. So if the end_time > start_time and the dates are the same, the check will pass. So, shouldn't be a problem.

api/lib/validation/attributeValidator.js
pradyutsarma: Fair point. Will change.

api/test/unit/validation/validationMiddleware.test.js
pradyutsarma: @fraenkel If you have the "schedules" tag specified, then you've to specify at least one of either specific_date or recurring schedule. Otherwise, why would you have the schedules section at all in the policy?

Can you please point me to one of the tests where your "expect"ations weren't met?

fraenkel: Change any of them. I changed a 400 to a 200, as well as false to true, and npm test just passed.

api/lib/validation/attributeValidator.js
fraenkel: How? Your compareDates does a strict isAfter. If startDate == endDate, you fail.

api/test/unit/validation/validationMiddleware.test.js
pradyutsarma: Thanks @fraenkel . Will check thoroughly again.

fraenkel:
Merged

PR #68, opened 2016-08-09
119511497: I can delete a policy's schedule - Specific Date schedules

ghaih:
This story covers only the specific date schedules. Few place holders for recurring schedules have been created.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
@@ -21,7 +21,6 @@ import org.cloudfoundry.autoscaler.scheduler.util.SpecificDateScheduleDateTime;
 import org.cloudfoundry.autoscaler.scheduler.util.error.DatabaseValidationException;
 import org.cloudfoundry.autoscaler.scheduler.util.error.SchedulerInternalException;
 import org.cloudfoundry.autoscaler.scheduler.util.error.ValidationErrorResult;
-import org.quartz.SchedulerException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
@@ -38,7 +37,7 @@ public class ScheduleManager {
 	@Autowired
 	private RecurringScheduleDao recurringScheduleDao;
 	@Autowired
-	private ScheduleJobManager scalingJobManager;
+	private ScheduleJobManager scheduleJobManager;
 	@Autowired
 	private ValidationErrorResult validationErrorResult;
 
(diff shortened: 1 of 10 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -639,4 +636,50 @@ private RecurringScheduleEntity saveNewRecurringSchedule(RecurringScheduleEntity
...
+			specificDateScheduleDao.delete(specificDateScheduleEntity);
+		} catch (DatabaseValidationException dve) {
+			validationErrorResult.addErrorForDatabaseValidationException(dve, "database.error.delete.failed",
+					"app_id=" + specificDateScheduleEntity.getAppId());
+			throw new SchedulerInternalException("Database error", dve);
+		}
+	}
+
+	private void deleteRecurringSchedule(RecurringScheduleEntity recurringScheduleEntity) {
+
```
fraenkel: If you are going to put empty methods for future use, at least have it throw an exception with some catch phrase so it is easy to determine what needs to be completed and have it fail if we attempt to use it.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -639,4 +636,50 @@ private RecurringScheduleEntity saveNewRecurringSchedule(RecurringScheduleEntity
...
+			// Ask ScalingJobManager to delete scaling job 
+			scheduleJobManager.deleteSimpleJob(specificDateScheduleEntity);
+			
+		}
+
+		// Get all the recurring schedules for the specifies application id and delete them.
+		List<RecurringScheduleEntity> recurringSchedules = recurringScheduleDao.findAllRecurringSchedulesByAppId(appId);
+		for (RecurringScheduleEntity recurringScheduleEntity : recurringSchedules) {
+			// Delete the specific date schedule from database
```
fraenkel: I would just remove the comment.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
```diff
@@ -331,6 +330,41 @@ public void testCreateSchedule_multiple_error() throws Exception {
...
 
+	@Test
+	@Transactional
+	public void testDeleteSchedules() throws Exception {
+
+		// Test multiple applications each having multiple specific date schedules, no recurring schedule
+		String[] multipleAppIds = TestDataSetupHelper.generateAppIds(5);
+		assertDeleteSchedules(multipleAppIds, 5, 0);
+
+		// Note: Once recurring schedule delete in place following can be used
```
fraenkel: Should really be 3 tests rather than 3 in 1 but we seem to be paying a heavy cost in sprint startup for each test so it is fine.

PR #69, opened 2016-08-10
127871921 : Add recurring schedule validation for start/end date.

kanekoh:
Contains additional commit for refactored code.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
@@ -17,71 +17,59 @@ public class DataValidationHelper {
 
 	/**
 	 * Checks if the specified object is null.
-	 * 
+	 *
 	 * @param object
 	 * @return true if not null otherwise false
 	 */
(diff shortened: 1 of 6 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
```diff
@@ -168,6 +157,52 @@ public static boolean isElementUnique(int[] array) {
 		return overlapDateTimeValidationErrorMsgList;
 	}
 
+	private static boolean isStartEndDateOverlapping(RecurringScheduleTime current, RecurringScheduleTime next) {
+		boolean isOverlapping;
+
+		// NOTE: The Start and End Dates in the schedules are not sorted, so we need to compare dates
```
fraenkel: Would it be easier to just sort them first and then do the check?

ghaih: We had thought about sorting the dates.Sorting current and next date is a bit complex since these dates are optional and can have same start date and different end dates and vice versa...plus many more combinations. We will try again to find some way out.

PR #70, opened 2016-08-10
retrive metrics for an app over an interval

boyang9527:
[#127841745]

- add another endpoint to get metrics history
- add hanlder to retrieve metrics from database with start and end time
- change db to return metrics ordered by timestamp
- test cases for server, handler, db  and main


src/metricscollector/cmd/metricscollector/main.go
```diff
@@ -60,13 +62,18 @@ func main() {
...
 	logger.Info("create-noaa-client", map[string]interface{}{"dopplerUrl": dopplerUrl})
 	tlsConfig := &tls.Config{InsecureSkipVerify: true}
 	noaa := consumer.New(dopplerUrl, tlsConfig, nil)
 	noaa.RefreshTokenFrom(cfClient)
 
-	httpServer := server.NewServer(logger, conf.Server, cfClient, noaa)
+	var database db.DB
+	database, err = sqldb.NewSQLDB(&conf.Db, logger.Session("db"))
+	if err != nil {
+		logger.Error("failed to connect database", err, lager.Data{"conf": conf.Db})
```
fraenkel: Shouldn't we exit?

src/metricscollector/server/memory_metric_handler.go
```diff
@@ -44,20 +48,20 @@ func (h *MemoryMetricHandler) GetMemoryMetric(w http.ResponseWriter, r *http.Req
...
 
-	h.logger.Debug("get-container-metrics-from-doppler", lager.Data{"container-metrics": containerMetrics})
 	metric := metrics.GetMemoryMetricFromContainerMetrics(appId, containerMetrics)
-
-	body, err := json.Marshal(metric)
+	var body []byte
+	body, err = json.Marshal(metric)
 	if err != nil {
-		h.logger.Error("marshal-memory-metrics", err, lager.Data{"metric": metric})
+		h.logger.Error("Get-memory-metrics-marshal", err, lager.Data{"metric": metric})
```
fraenkel: would it be useful to know the appId?

src/metricscollector/server/memory_metric_handler.go
```diff
@@ -67,3 +71,75 @@ func (h *MemoryMetricHandler) GetMemoryMetric(w http.ResponseWriter, r *http.Req
...
+	appId := vars["appid"]
+	startParam := r.URL.Query()["start"]
+	endParam := r.URL.Query()["end"]
+	h.logger.Debug("get-memeory-metric-history", lager.Data{"appId": appId, "start": startParam, "end": endParam})
+
+	var err error
+	start := int64(0)
+	end := int64(-1)
+
+	if startParam != nil {
```
fraenkel: not needed.
len(nil) == 0

src/metricscollector/server/memory_metric_handler_test.go
```diff
@@ -106,4 +115,194 @@ var _ = Describe("MemoryMetricHandler", func() {
...
+					req, err = http.NewRequest(http.MethodGet, "http://localhost/v1/apps/an-app-id/metrics_history/memory?start=123&start=231", nil)
+					Expect(err).ToNot(HaveOccurred())
+
+				})
+
+				It("returns 400", func() {
+					Expect(resp.Code).To(Equal(http.StatusBadRequest))
+
+					errJson := &ErrorResponse{}
+					d := json.NewDecoder(resp.Body)
```
fraenkel: Might as well use json.Unmarshal() to reduce lines.

src/metricscollector/server/memory_metric_handler_test.go
```diff
@@ -106,4 +115,194 @@ var _ = Describe("MemoryMetricHandler", func() {
...
+					err = d.Decode(errJson)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(errJson.Code).To(Equal("Bad-Request"))
```
fraenkel: Why not just do a Equal(&ErrorResponse{
Code: "Bad-Request",
Message: ".....",
})

src/metricscollector/server/memory_metric_handler_test.go
```diff
@@ -106,4 +115,194 @@ var _ = Describe("MemoryMetricHandler", func() {
...
+					Expect(errJson.Message).To(Equal("Error parsing end time"))
+				})
+			})
+
+		})
+
+		Context("when request query string is valid", func() {
```
fraenkel: If you reordered your tests and got the base cases out of the way, you need to check less later on.
While it is good practice to make sure everything is valid, you can also get away with just checking for the difference/new behavior.

src/metricscollector/cmd/metricscollector/main.go
boyang9527: added exit and close the database

src/metricscollector/server/memory_metric_handler.go
boyang9527: added in log data

src/metricscollector/server/memory_metric_handler.go
boyang9527: still need to check whether "startParam == nil" or "len(startParam) == 0". In this case, start will remain the default value 0. 
Anyway changed to use len() for all cases

src/metricscollector/server/memory_metric_handler_test.go
boyang9527: all changed

src/metricscollector/server/memory_metric_handler_test.go
boyang9527: all changed.

src/metricscollector/server/memory_metric_handler_test.go
boyang9527: changed the order to reduce the checks.

PR #71, opened 2016-08-11
119511761: A schedule can be updated/replaced - Specific Date Schedule

ghaih:
Contains addition commit for refactored code.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
@@ -35,7 +35,12 @@ public class ScheduleRestController {
 	public ResponseEntity<ApplicationScalingSchedules> getAllSchedules(@PathVariable String app_id) {
 		logger.info("Get All schedules for application: " + app_id);
 		ApplicationScalingSchedules savedApplicationSchedules = scheduleManager.getAllSchedules(app_id);
-
+		
+		// Any exception in getting the schedules
+		if (validationErrorResult.hasErrors()) {
+			throw new InvalidDataException();
+		}
+		
 		// No schedules found for the specified application return status code NOT_FOUND
 		if (!savedApplicationSchedules.hasSchedules()) {
 			return new ResponseEntity<>(null, null, HttpStatus.NOT_FOUND);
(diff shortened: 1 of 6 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
```diff
@@ -35,7 +35,12 @@
 	public ResponseEntity<ApplicationScalingSchedules> getAllSchedules(@PathVariable String app_id) {
 		logger.info("Get All schedules for application: " + app_id);
 		ApplicationScalingSchedules savedApplicationSchedules = scheduleManager.getAllSchedules(app_id);
-
+		
+		// Any exception in getting the schedules
```
fraenkel: How does this ever happen?
getAllSchedules throws an Exception.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
```diff
@@ -61,11 +65,25 @@
 
 		// If there are no validation errors then proceed with persisting the schedules
 		if (!validationErrorResult.hasErrors()) {
+			ApplicationScalingSchedules existingSchedules = scheduleManager.getAllSchedules(app_id);
+
```
fraenkel: And now you show why using the validationErrorResult is so dangerous.
Here you didn't check like you did above.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
kanekoh: We get an exception in that case. For instance, database connection error.
We will remove check for the ValidationErrorResult in case of get if you think it is not justified.

fraenkel: My point is that it is all confused. It seems to be a guessing game when to catch an exception vs check the validationErrorResult. It isn't obvious since nothing is passed in or explicitly thrown.

ghaih:
Removed the unnecessary checks on validationErrorResult , since the database exceptions are being handled by ControllerExceptionHandler.
Minor change in TestDataSetupHelper to correct the cast.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
```diff
@@ -63,36 +58,29 @@
 		logger.info("Validate schedules for application: " + app_id);
 		scheduleManager.validateSchedules(app_id, rawApplicationSchedules);
 
-		// If there are no validation errors then proceed with persisting the schedules
-		if (!validationErrorResult.hasErrors()) {
+		if (validationErrorResult.hasErrors()) {
+			throw new InvalidDataException();
+
+		} else { // If there are no validation errors then proceed with persisting the schedules
+
```
fraenkel: Why have an else block?

fraenkel:
Merged, removed the else block

PR #73, opened 2016-08-16
share db among components in Go

boyang9527:
[#127841745]

- move db module out of metircscollector
- separate metrics and policy db interface
- changes in other modules accordingly
- add collector start in main.go of metricscollector


fraenkel:
```
==================
WARNING: DATA RACE
Read at 0x00000091526a by goroutine 22:
  metricscollector/cmd/metricscollector_test.glob..func2.1()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/cmd/metricscollector/metricscollector_suite_test.go:70 +0x6b
  github.com/onsi/gomega/ghttp.(*Server).ServeHTTP()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/gomega/ghttp/test_server.go:253 +0x488
  net/http.serverHandler.ServeHTTP()
      /usr/local/Cellar/go/1.7/libexec/src/net/http/server.go:2202 +0xbb
  net/http.(*conn).serve()
      /usr/local/Cellar/go/1.7/libexec/src/net/http/server.go:1579 +0x5f6

Previous write at 0x00000091526a by goroutine 8:
  metricscollector/cmd/metricscollector_test.glob..func5.9.2.1()
      /Users/fraenkel/workspace/app-autoscaler/src/metricscollector/cmd/metricscollector/metricscollector_test.go:112 +0x3a
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/leafnodes/runner.go:109 +0xbb
  github.com/onsi/ginkgo/internal/leafnodes.(*runner).run()
      /Users/fraenkel/workspace/app-autoscaler/src/github.com/onsi/ginkgo/internal/leafnodes/runner.go:63 +0x141
```

fraenkel:
Running ginkgo on the db package 

```
FAIL! -- 13 Passed | 6 Failed | 0 Pending | 0 Skipped
```

fraenkel:
Ok, so I think I know what is going on with the db package. I am running them in parallel which causes all sorts of failures. It is something we should support. But can create a separate story if it is too much work to fix.

The data race however is real.

src/db/sqldb/metrics_sqldb.go
```diff
@@ -0,0 +1,130 @@
...
+	metricsDB.sqldb, err = sql.Open(db.PostgresDriverName, url)
+	if err != nil {
+		logger.Error("open-metrics-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	err = metricsDB.sqldb.Ping()
+	if err != nil {
+		metricsDB.sqldb.Close()
+		metricsDB.sqldb = nil
```
fraenkel: why?

src/db/sqldb/metrics_sqldb.go
```diff
@@ -0,0 +1,130 @@
...
+		logger.Error("ping-metrics-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	return metricsDB, nil
+}
+
+func (mdb *MetricsSQLDB) Close() error {
+	if mdb.sqldb != nil {
```
fraenkel: why? who sets it to nil?

src/db/sqldb/metrics_sqldb.go
```diff
@@ -0,0 +1,130 @@
...
+	if end < 0 {
+		end = time.Now().UnixNano()
+	}
+
+	mtrcs := []*metrics.Metric{}
+	rows, err := mdb.sqldb.Query(query, appid, name, start, end)
+	if err != nil {
+		mdb.logger.Error("retrive-metrics-from-applicationmetrics-table", err,
+			lager.Data{"query": query, "appid": appid, "metricName": name, "start": start, "end": end})
+		return mtrcs, err
```
fraenkel: why would we return an empty slice and an error?

src/db/sqldb/policy_sqldb.go
```diff
@@ -0,0 +1,76 @@
...
+	policyDB.sqldb, err = sql.Open(db.PostgresDriverName, url)
+	if err != nil {
+		logger.Error("open-policy-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	err = policyDB.sqldb.Ping()
+	if err != nil {
+		policyDB.sqldb.Close()
+		policyDB.sqldb = nil
```
fraenkel: Same as above

src/db/sqldb/policy_sqldb.go
```diff
@@ -0,0 +1,76 @@
...
+		logger.Error("ping-policy-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	return policyDB, nil
+}
+
+func (pdb *PolicySQLDB) Close() error {
+	if pdb.sqldb != nil {
```
fraenkel: Same question

src/metricscollector/cmd/metricscollector/main.go
```diff
@@ -67,15 +75,33 @@ func main() {
...
-	defer database.Close()
+	defer policyDB.Close()
+
+	mcClock := clock.NewClock()
+	createPoller := func(appId string) collector.AppPoller {
+		return collector.NewAppPoller(appId, conf.Collector.PollInterval, logger.Session("app-poller"), cfClient, noaa, metricsDB, mcClock)
+	}
+
+	collectServer := collector.NewCollector(conf.Collector.RefreshInterval, logger.Session("collector"), policyDB, mcClock, createPoller)
+	collectServer.Start()
```
fraenkel: If we are using ifrit, then this should be wrapped in an ifritFunc that is ready and waits for a signal to call stop.

src/db/sqldb/metrics_sqldb.go
boyang9527: your are right.  We will never get a  metricsDB with a nil metricsDB.sqldb as long as it is created from NewMetricsSQLDB(), not initialized directly from struct MetricsSQLDB

src/db/sqldb/metrics_sqldb.go
boyang9527: changed to return nil, err

src/metricscollector/cmd/metricscollector/main.go
boyang9527: @fraenkel 
do you mean we remove the start / stop function of collector, and get all the logic in  Run function ?  instead of just having a new Run func to wrap the existing Start/Stop func

fraenkel:
Here is an example of an ifrit RunFunc
https://github.com/cf-furnace/furnace-release/blob/master/src/github.com/cf-furnace/route-emitter/cmd/route-emitter/main.go#L144

boyang9527:
Michael,  my questions was do you want to change the current collector implementation to remove Start/Stop func , and replace with just Run() function.  or use either ifrit RunFunc or have a Run method in collecror to wrap Start/Stop method, like following:
```
func (c *Collector) Run(signals <-chan os.Signal, ready chan<- struct{}) error {
        c.Start()
        close(ready)

        <-signals
        c.Stop()

        return nil
}

```

fraenkel:
Just offering an alternative. Its one way to just inline the stuff but its up to you.  If you already had a struct to hang Run off of it didn't matter.

boyang9527:
I added an inline  ifrit RunFunc in main.go.  If we revise collector code to merge several functions together into one Run() function as the implementation of  the runner interface, we will need to revise the test case though we can reuse most of the test cases.  Adding RunFunc in main.go will be able to reuse the current test code both in collector and main. 

I updated the commit, please review.

PR #77, opened 2016-08-22
126384779: As a user, delete removes recurring rules as well

kanekoh:


scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManagerTest.java
```diff
@@ -199,6 +200,17 @@ public void testDeleteSpecificDateSchedules_throw_DatabaseValidationException()
...
 
+	@Test
+	@Rollback
+	public void testDeleteRecurringSchedules_throw_DatabaseValidationException() {
+		String appId = TestDataSetupHelper.generateAppIds(1)[0];
+
+		// Create 4 recurring schedules.
+		createScheduleNotThrowAnyException(appId, 0, 4);
+
+		assertDatabaseExceptionOnDelete(appId, ScheduleTypeEnum.RECURRING);
```
fraenkel: Shouldn't we guarantee the DB hasn't removed anything?
We should also have a test which has both schedules and any exception does not remove anything for a given app.

fraenkel:
Merged

PR #79, opened 2016-08-24
126384815: As a user, I can update a schedule w/ recurring rules

kanekoh:


scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
```diff
@@ -114,15 +114,15 @@ public void testCreateAndGetSchedules() throws Exception {
...
 		String appId = TestDataSetupHelper.generateAppIds(1)[0];
-		// Create one specific date schedule for the application.
-		callCreateSchedules(appId, 1, 0);
+		// Create one specific date schedule and two recurring schedules for the application.
+		callCreateSchedules(appId, 1, 2);
 
-		// Create two specific date schedule for the same application.
-		ResultActions resultActions = callCreateSchedules(appId, 2, 0);
+		// Create two specific date schedules and one recurring schedule for the same application.
+		ResultActions resultActions = callCreateSchedules(appId, 2, 1);
```
fraenkel: The naming is a bit poor since it is doing more than just Create, its doing an Upsert.

kanekoh: The callCreateSchedules method here is calling the the RestController's createSchedules method using mock, so named like that.

fraenkel: You are missing the point. If I create(A), and then I create(A) again, in most worlds that would fail. In a world where the second A succeeds it is either doing an update or replace. In either case, the create() becomes a misnomer and doesn't reflect what it really does. Since you chose to hid the fact that it is a PUT vs a POST, the naming vs the behavior is just a bit odd to anyone looking at the code.

ghaih:
Thought you were talking about the test class method. Yes in the controller should have renamed it after implementing the update functionality. Will make the change in next commit.Thanks.

PR #80, opened 2016-08-25
Metrics are periodically retrieved

qibobo:
[#119505277]

### DIFF
--- a/src/dataaggregator/aggregator/aggregator.go
+++ b/src/dataaggregator/aggregator/aggregator.go
@@ -1,0 +1,79 @@
+package aggregator
+
+import (
+	"code.cloudfoundry.org/cfhttp"
+	"code.cloudfoundry.org/clock"
+	"code.cloudfoundry.org/lager"
+	"dataaggregator/appmetric"
+	"dataaggregator/db"
+	"dataaggregator/policy"
+	"strconv"
+	"time"
+)
+
+type Aggregator struct {
+	logger             lager.Logger
+	metricCollectorUrl string
+	doneChan           chan bool
+	appChan            chan *appmetric.AppMonitor
+	metricChan         chan *appmetric.AppMetric
+	policyPoller       *PolicyPoller
+	metricPollerCount  int64
+	metricPollerArray  []*MetricPoller
+	database           db.DB
+}
+
+func NewAggregator(logger lager.Logger, clock clock.Clock, policyPollerInterval time.Duration, database db.DB, metricCollectorUrl string, metricPollerCount int64) *Aggregator {
+	aggregator := &Aggregator{
+		logger:             logger.Session("aggregator"),
+		metricCollectorUrl: metricCollectorUrl,
+		doneChan:           make(chan bool),
+		appChan:            make(chan *appmetric.AppMonitor, 10),
+		metricChan:         make(chan *appmetric.AppMetric, 10),
+		metricPollerCount:  metricPollerCount,
+		metricPollerArray:  []*MetricPoller{},
+		database:           database,
+	}
+	aggregator.policyPoller = NewPolicyPoller(logger, clock, policyPollerInterval, database, aggregator.consumeTrigger)
+	var i int64
+	for i = 0; i < metricPollerCount; i++ {
+		pollerLogger := lager.NewLogger("Metric-Poller-" + strconv.FormatInt(i, 10))
+		poller := NewMetricPoller(metricCollectorUrl, pollerLogger, aggregator.appChan, aggregator.consumeAppMetric, cfhttp.NewClient())
+		aggregator.metricPollerArray = append(aggregator.metricPollerArray, poller)
+	}
+	return aggregator
+}
+func (a *Aggregator) consumeTrigger(policyList []*policy.PolicyJson, triggerMap map[string]*policy.Trigger) {
+	for appId, trigger := range triggerMap {
+		for _, rule := range trigger.TriggerRecord.ScalingRules {
+			a.appChan <- &appmetric.AppMonitor{
+				AppId:          appId,
+				MetricType:     rule.MetricType,
+				StatWindowSecs: rule.StatWindowSecs,
+			}
+		}
+	}
+}
+func (a *Aggregator) consumeAppMetric(appMetric *appmetric.AppMetric) {
+	a.metricChan <- appMetric
+	err := a.database.SaveAppMetric(appMetric)
+	if err != nil {
+		a.logger.Error("save appmetric to database failed", err, lager.Data{"appmetric": appMetric})
+	}
+}
+func (a *Aggregator) Start() {
+	a.policyPoller.Start()
+	for _, metricPoller := range a.metricPollerArray {
+		metricPoller.Start()
+	}
+	a.logger.Info("policy-poller-started")
+}
+func (a *Aggregator) Stop() {
+	a.policyPoller.Stop()
+	for _, metricPoller := range a.metricPollerArray {
+		metricPoller.Stop()
+	}
+	// close(a.appChan)
+	// close(a.metricChan)
+	a.logger.Info("aggregator-stopped")
+}
--- a/src/dataaggregator/aggregator/aggregator.go
+++ b/src/dataaggregator/aggregator/aggregator.go
@@ -7,7 +7,6 @@ import (
 	"dataaggregator/appmetric"
 	"dataaggregator/db"
 	"dataaggregator/policy"
-	"errors"
 	"net/http"
 	"time"
 )
@@ -18,30 +17,30 @@ type Aggregator struct {
 	doneChan           chan bool
 	appChan            chan *appmetric.AppMonitor
 	policyPoller       *PolicyPoller
-	metricPollerCount  int64
-	MetricPollerArray  []*MetricPoller
+	metricPollerCount  int
+	metricPollerArray  []*MetricPoller
 	database           db.DB
 }
 
-func NewAggregator(logger lager.Logger, clock clock.Clock, policyPollerInterval time.Duration, database db.DB, metricCollectorUrl string, metricPollerCount int64) *Aggregator {
+func NewAggregator(logger lager.Logger, clock clock.Clock, policyPollerInterval time.Duration, database db.DB, metricCollectorUrl string, metricPollerCount int) *Aggregator {
 	aggregator := &Aggregator{
 		logger:             logger.Session("Aggregator"),
 		metricCollectorUrl: metricCollectorUrl,
 		doneChan:           make(chan bool),
 		appChan:            make(chan *appmetric.AppMonitor, 10),
 		metricPollerCount:  metricPollerCount,
-		MetricPollerArray:  []*MetricPoller{},
+		metricPollerArray:  []*MetricPoller{},
 		database:           database,
 	}
 	aggregator.policyPoller = NewPolicyPoller(logger, clock, policyPollerInterval, database, aggregator.ConsumeTrigger, aggregator.appChan)
 
 	client := cfhttp.NewClient()
-	client.Transport.(*http.Transport).MaxIdleConnsPerHost = (int)(metricPollerCount)
+	client.Transport.(*http.Transport).MaxIdleConnsPerHost = metricPollerCount
 
-	var i int64
+	var i int
 	for i = 0; i < metricPollerCount; i++ {
 		poller := NewMetricPoller(metricCollectorUrl, logger, aggregator.appChan, aggregator.ConsumeAppMetric, client)
-		aggregator.MetricPollerArray = append(aggregator.MetricPollerArray, poller)
+		aggregator.metricPollerArray = append(aggregator.metricPollerArray, poller)
 	}
 	return aggregator
 }
@@ -61,7 +60,6 @@ func (a *Aggregator) ConsumeTrigger(triggerMap map[string]*policy.Trigger, appCh
 }
 func (a *Aggregator) ConsumeAppMetric(appMetric *appmetric.AppMetric) {
 	if appMetric == nil {
-		a.logger.Error("appmetric is nil", errors.New("Should not save a nil appmetric to database"))
 		return
 	}
 	err := a.database.SaveAppMetric(appMetric)
@@ -71,14 +69,14 @@ func (a *Aggregator) ConsumeAppMetric(appMetric *appmetric.AppMetric) {
 }
 func (a *Aggregator) Start() {
 	a.policyPoller.Start()
-	for _, metricPoller := range a.MetricPollerArray {
+	for _, metricPoller := range a.metricPollerArray {
 		metricPoller.Start()
 	}
 	a.logger.Info("started")
 }
 func (a *Aggregator) Stop() {
 	a.policyPoller.Stop()
-	for _, metricPoller := range a.MetricPollerArray {
+	for _, metricPoller := range a.metricPollerArray {
 		metricPoller.Stop()
 	}
 	a.logger.Info("stopped")
--- a/src/dataaggregator/aggregator/aggregator.go
+++ b/src/dataaggregator/aggregator/aggregator.go
@@ -7,6 +7,7 @@ import (
 	"dataaggregator/appmetric"
 	"dataaggregator/db"
 	"dataaggregator/policy"
+	"net/http"
 	"strconv"
 	"time"
 )
@@ -16,7 +17,6 @@ type Aggregator struct {
 	metricCollectorUrl string
 	doneChan           chan bool
 	appChan            chan *appmetric.AppMonitor
-	metricChan         chan *appmetric.AppMetric
 	policyPoller       *PolicyPoller
 	metricPollerCount  int64
 	metricPollerArray  []*MetricPoller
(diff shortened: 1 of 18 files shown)


src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -0,0 +1,128 @@
...
+			Expect(appMetric.Value).To(BeNumerically("==", 250))
+			return nil
+		}
+		clock = fakeclock.NewFakeClock(time.Now())
+		logger = lager.NewLogger("Aggregator-test")
+		metricServer = ghttp.NewServer()
+		metricServer.RouteToHandler("GET", "/v1/apps/"+testAppId+"/metrics_history/memory", ghttp.RespondWithJSONEncoded(http.StatusOK,
+			&metrics))
+	})
+	Context("Start", func() {
```
fraenkel: This is insufficient testing. The problem you have here is how you decided to link the Pollers with the Aggregator. You aren't testing enough of consumeTrigger or consumeAppMetric.
Expose them and you can test it directly.
The wiring of the pollers and aggregator should be done elsewhere so you can just test the wiring and validate that you have created all the pollers on start and that they stop.

src/dataaggregator/aggregator/aggregator.go
```diff
@@ -0,0 +1,79 @@
+package aggregator
+
+import (
+	"code.cloudfoundry.org/cfhttp"
+	"code.cloudfoundry.org/clock"
+	"code.cloudfoundry.org/lager"
+	"dataaggregator/appmetric"
+	"dataaggregator/db"
```
fraenkel: please start using the shared packages.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -0,0 +1,122 @@
...
+		logger:             logger.Session("metric-poller"),
+		appChan:            appChan,
+		doneChan:           make(chan bool),
+		metricConsumer:     metricConsumer,
+		httpClient:         httpClient,
+	}
+}
+func (m *MetricPoller) Start() {
+	go m.startMetricRetrieve()
+	m.logger.Info("metric-poller-started")
```
fraenkel: This will create a metric-poller-metric-poller-started message. You have this problem all over.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -0,0 +1,122 @@
...
+	appId := app.AppId
+	metricType := app.MetricType
+	endTime := time.Now().UnixNano()
+	startTime := endTime - app.StatWindowSecs*1000*1000
+	var url string
+	switch metricType {
+	case "MemoryUsage":
+		url = m.metricCollectorUrl + "/v1/apps/" + app.AppId + "/metrics_history/memory?start=" + strconv.FormatInt(startTime, 10) + "&end=" + strconv.FormatInt(endTime, 10)
+	default:
+		url = m.metricCollectorUrl + "/v1/apps/" + app.AppId + "/metrics_history/memory?start=" + strconv.FormatInt(startTime, 10) + "&end=" + strconv.FormatInt(endTime, 10)
```
fraenkel: Shouldn't this be an error?

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	resp, err := m.httpClient.Get(url)
+	if err != nil {
+		m.logger.Error("Retrieve metric failed", err, lager.Data{"appId": appId, "metricType": metricType, "err": err})
```
fraenkel: Just return, no else body required.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -0,0 +1,122 @@
...
+	} else {
+		defer resp.Body.Close()
+		var metrics []*metrics.Metric
+		if resp.StatusCode == http.StatusOK {
+			data, _ := ioutil.ReadAll(resp.Body)
```
fraenkel: Either check the error or use the JsonDecoder and pass resp.Body in.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -0,0 +1,122 @@
...
+		timestamp = metric.TimeStamp
+		for _, instanceMetric := range metric.Instances {
+			count++
+			intValue, _ := strconv.ParseInt(instanceMetric.Value, 10, 64)
+			sum += intValue
+		}
+	}
+
+	if count == 0 {
+		return &appmetric.AppMetric{
```
fraenkel: Just define the return once. Just reset the values when count == 0.

src/dataaggregator/aggregator/metricPoller_test.go
```diff
@@ -0,0 +1,162 @@
...
+			var consumed chan *AppMetric
+			BeforeEach(func() {
+				consumed = make(chan *AppMetric, 1)
+				metricConsumer = func(appMetric *AppMetric) {
+					consumed <- appMetric
+				}
+			})
+			Context("when retrieve metrics from metric-collector successfully", func() {
+				BeforeEach(func() {
+					metricServer = ghttp.NewServer()
```
fraenkel: When is this stopped?

src/dataaggregator/aggregator/metricPoller_test.go
```diff
@@ -0,0 +1,162 @@
...
+					Expect(appMetric).To(Equal(&AppMetric{
+						AppId:      testAppId,
+						MetricType: metricType,
+						Value:      0,
+						Unit:       "",
+						Timestamp:  0,
+					}))
+				})
+			})
+			Context("when metric-collector returns error", func() {
```
fraenkel: Where is the test when the server isn't running?

src/dataaggregator/aggregator/metricPoller_test.go
```diff
@@ -0,0 +1,162 @@
...
+			metricPoller.Stop()
+			appChan <- &AppMonitor{
+				AppId:          testAppId,
+				MetricType:     metricType,
+				StatWindowSecs: 10,
+			}
+
+		})
+		It("stops the aggregating", func() {
+			Eventually(consumed).ShouldNot(Receive())
```
fraenkel: Consistently

src/dataaggregator/policy/policy.go
```diff
@@ -33,16 +33,16 @@ func (p *PolicyJson) GetTrigger() *Trigger {
 }
 
 type TriggerRecord struct {
-	InstanceMaxCount int            `json:"instance_max_count"`
-	InstanceMinCount int            `json:"instance_min_count"`
+	InstanceMaxCount int64          `json:"instance_max_count"`
```
fraenkel: why do these need to be int64? Can't imagine us having that many instances.

src/dataaggregator/aggregator/aggregator.go
```diff
@@ -25,37 +25,39 @@ type Aggregator struct {
 
 func NewAggregator(logger lager.Logger, clock clock.Clock, policyPollerInterval time.Duration, database db.DB, metricCollectorUrl string, metricPollerCount int64) *Aggregator {
 	aggregator := &Aggregator{
-		logger:             logger.Session("aggregator"),
+		logger:             logger,
```
fraenkel: This should be logger.Session("aggregator").

src/dataaggregator/aggregator/aggregator.go
```diff
@@ -66,7 +68,7 @@ func (a *Aggregator) Start() {
 	for _, metricPoller := range a.metricPollerArray {
 		metricPoller.Start()
 	}
-	a.logger.Info("policy-poller-started")
+	a.logger.Info("aggregator-started")
```
fraenkel: Just "started".

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -93,6 +93,55 @@ var _ = Describe("Aggregator", func() {
...
+						StatWindowSecs:     300,
+						BreachDurationSecs: 300,
+						CoolDownSecs:       300,
+						Threshold:          30,
+						Operator:           "<",
+						Adjustment:         "-1",
+					}}},
+			}}
+		})
+		It("should parse the triggers to appmonitor and put them in appChan", func() {
```
fraenkel: Insufficient testing.

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -93,6 +93,55 @@ var _ = Describe("Aggregator", func() {
...
+		JustBeforeEach(func() {
+			aggregator = NewAggregator(logger, clock, TestPolicyPollerInterval, database, metricServer.URL(), TestMetricPollerCount)
+			appmetric = &AppMetric{
+				AppId:      testAppId,
+				MetricType: metricType,
+				Value:      250,
+				Unit:       "bytes",
+				Timestamp:  timestamp}
+		})
+		It("should save the appmetric to database", func() {
```
fraenkel: Insufficient testing.

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -93,6 +93,55 @@ var _ = Describe("Aggregator", func() {
...
+			aggregator.ConsumeAppMetric(appmetric)
+			Eventually(database.SaveAppMetricCallCount).Should(BeNumerically("==", 1))
+		})
+	})
 	Context("Start", func() {
```
fraenkel: Where are the tests to guarantee N metric pollers are created?

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -26,7 +27,7 @@ type MetricPoller struct {
 func NewMetricPoller(metricCollectorUrl string, logger lager.Logger, appChan chan *appmetric.AppMonitor, metricConsumer MetricConsumer, httpClient *http.Client) *MetricPoller {
 	return &MetricPoller{
 		metricCollectorUrl: metricCollectorUrl,
-		logger:             logger.Session("metric-poller"),
+		logger:             logger,
```
fraenkel: Let's be consistent. Just pass the logger through and let the function set the name, so what you had was correct. This will prevent nesting, like aggregator-metric-poller while following a common pattern.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -56,22 +57,27 @@ func (m *MetricPoller) retrieveMetric(app *appmetric.AppMonitor) {
 	appId := app.AppId
 	metricType := app.MetricType
 	endTime := time.Now().UnixNano()
-	startTime := endTime - app.StatWindowSecs*1000*1000
+	startTime := endTime - int64(app.StatWindowSecs)*1000*1000
```
fraenkel: make StatWindowSecs just StatWindow as a time.Duration

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -56,22 +57,27 @@ func (m *MetricPoller) retrieveMetric(app *appmetric.AppMonitor) {
...
 	case "MemoryUsage":
 		url = m.metricCollectorUrl + "/v1/apps/" + app.AppId + "/metrics_history/memory?start=" + strconv.FormatInt(startTime, 10) + "&end=" + strconv.FormatInt(endTime, 10)
 	default:
 		url = m.metricCollectorUrl + "/v1/apps/" + app.AppId + "/metrics_history/memory?start=" + strconv.FormatInt(startTime, 10) + "&end=" + strconv.FormatInt(endTime, 10)
 	}
-	resp, err := m.httpClient.Get(url)
-	if err != nil {
-		m.logger.Error("Retrieve metric failed", err, lager.Data{"appId": appId, "metricType": metricType, "err": err})
-	} else {
+	if metricType == "MemoryUsage" {
```
fraenkel: flip this
if metricType != "MemoryUsage" {
  log.Error
  return
}
....

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -56,22 +57,27 @@ func (m *MetricPoller) retrieveMetric(app *appmetric.AppMonitor) {
...
+			m.logger.Error("Retrieve metric failed", err, lager.Data{"appId": appId, "metricType": metricType, "err": err})
+			return
+		}
 		defer resp.Body.Close()
 		var metrics []*metrics.Metric
 		if resp.StatusCode == http.StatusOK {
-			data, _ := ioutil.ReadAll(resp.Body)
+			data, readError := ioutil.ReadAll(resp.Body)
+			if readError != nil {
+				m.logger.Error("Can not read data from response", readError)
```
fraenkel: You can't continue after this error.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -101,22 +109,23 @@ func (m *MetricPoller) doAggregate(appId string, metricType string, metrics []*m
 			sum += intValue
 		}
 	}
-
+	var avgAppMetric *appmetric.AppMetric
 	if count == 0 {
-		return &appmetric.AppMetric{
+		avgAppMetric = &appmetric.AppMetric{
```
fraenkel: The AppMetric should only be constructed in one spot. just set the values in the if block.

src/dataaggregator/aggregator/metricPoller_test.go
```diff
@@ -132,6 +133,20 @@ var _ = Describe("MetricPoller", func() {
 					Consistently(consumed).ShouldNot(Receive())
 				})
 			})
+			Context("when metric-collector is not running", func() {
+				JustBeforeEach(func() {
+					metricServer.Close()
+				})
+				BeforeEach(func() {
+					metricServer = ghttp.NewServer()
+					metricServer.RouteToHandler("GET", "/v1/apps/"+testAppId+"/metrics_history/memory", ghttp.RespondWithJSONEncoded(http.StatusOK,
```
fraenkel: Why do we have a route?

src/dataaggregator/aggregator/metricPoller.go
qibobo: @fraenkel When there is no metric from metric-collector, the aggregator will construct an zero-value AppMetric instance and persist it.

src/dataaggregator/aggregator/aggregator.go
```diff
@@ -19,63 +19,67 @@ type Aggregator struct {
 	appChan            chan *appmetric.AppMonitor
 	policyPoller       *PolicyPoller
 	metricPollerCount  int64
-	metricPollerArray  []*MetricPoller
+	MetricPollerArray  []*MetricPoller
```
fraenkel: No. This should remain private. We don't expose things just so it makes testing easier. There are other ways.

src/dataaggregator/aggregator/aggregator.go
```diff
@@ -19,63 +19,67 @@ type Aggregator struct {
...
+				AppId:      appId,
+				MetricType: rule.MetricType,
+				StatWindow: rule.StatWindow,
 			}
 		}
 	}
 }
 func (a *Aggregator) ConsumeAppMetric(appMetric *appmetric.AppMetric) {
+	if appMetric == nil {
+		a.logger.Error("appmetric is nil", errors.New("Should not save a nil appmetric to database"))
```
fraenkel: Do we expect nil or not? I am trying to understand what I would do if I received this error.
If it's just that we shouldn't save it, then don't log an error. If it is unexpected then log an error.

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -97,55 +97,95 @@ var _ = Describe("Aggregator", func() {
...
-						Threshold:          30,
-						Operator:           "<",
-						Adjustment:         "-1",
-					}}},
-			}}
+
+		})
+		Context("when there are data in triggerMap", func() {
+			JustBeforeEach(func() {
+				triggerMap = map[string]*Trigger{testAppId: &Trigger{
```
fraenkel: Why isn't this now in JustBeforeEach? Seems it should be in the BeforeEach and any test can override the triggerMap in their BeforeEach.
The JustBeforeEach can then just be the ConsumeTrigger call.

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -97,55 +97,95 @@ var _ = Describe("Aggregator", func() {
...
+				appmetric = &AppMetric{
+					AppId:      testAppId,
+					MetricType: metricType,
+					Value:      250,
+					Unit:       "bytes",
+					Timestamp:  timestamp}
+			})
+			It("should call database.SaveAppmetric to save the appmetric to database", func() {
+				aggregator.ConsumeAppMetric(appmetric)
+				Eventually(database.SaveAppMetricCallCount).Should(BeNumerically("==", 1))
```
fraenkel: Should(Equal(1))
BeNumerically should be used when you need > or >= like tests.

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -97,55 +97,95 @@ var _ = Describe("Aggregator", func() {
...
+			})
 		})
+		Context("when the appmetric is nil", func() {
+			JustBeforeEach(func() {
+				appmetric = nil
+			})
+			It("should call database.SaveAppmetric to save the appmetric to database", func() {
+				aggregator.ConsumeAppMetric(appmetric)
+				Consistently(database.SaveAppMetricCallCount).Should(BeNumerically("==", 0))
```
fraenkel: Equal(0) or BeZero()

src/dataaggregator/aggregator/aggregator_test.go
```diff
@@ -97,55 +97,95 @@ var _ = Describe("Aggregator", func() {
...
+			})
+		})
+
 	})
 	Context("Start", func() {
 		JustBeforeEach(func() {
 			aggregator = NewAggregator(logger, clock, TestPolicyPollerInterval, database, metricServer.URL(), TestMetricPollerCount)
 			aggregator.Start()
+			Eventually(len(aggregator.MetricPollerArray)).Should(BeNumerically("==", TestMetricPollerCount))
```
fraenkel: This should be an actual test but you shouldn't be testing it this way.

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -57,41 +57,36 @@ func (m *MetricPoller) retrieveMetric(app *appmetric.AppMonitor) {
 	appId := app.AppId
 	metricType := app.MetricType
 	endTime := time.Now().UnixNano()
-	startTime := endTime - int64(app.StatWindowSecs)*1000*1000
+	startTime := endTime - int64(app.StatWindow)
```
fraenkel: Why are we casting rather than using Duration everywhere until we actually need to cast?

src/dataaggregator/aggregator/metricPoller.go
```diff
@@ -57,41 +57,36 @@ func (m *MetricPoller) retrieveMetric(app *appmetric.AppMonitor) {
...
+		json.Unmarshal(data, &metrics)
+		avgMetric := m.doAggregate(appId, metricType, metrics)
+		if avgMetric != nil {
+			m.metricConsumer(avgMetric)
 		}
-
 	} else {
-		m.logger.Error("Unsupported metric type", fmt.Errorf("%s is not supported", metricType))
+		var errorResponse server.ErrorResponse
+		errBody, _ := ioutil.ReadAll(resp.Body)
```
fraenkel: What happens if this fails?
Also means you are missing tests.

src/dataaggregator/policy/policy_test.go
```diff
@@ -19,11 +19,11 @@ var _ = Describe("Policy", func() {
    "scaling_rules":[
       {
          "metric_type":"MemoryUsage",
-         "stat_window_secs":300,
-         "breach_duration_secs":300,
+         "stat_window":300,
```
fraenkel: This means 300ns not 300s.

src/dataaggregator/aggregator/metricPoller.go
qibobo: @fraenkel View the ioutil.go source code and found that when the input is too large, the ReadAll function will panic. So I add a test case with very long response body. Not sure whether I use it correctly.

fraenkel:
Fixed some of your tests and updated the README

PR #85, opened 2016-09-02
data aggregator is a sub_component of eventgenerator

qibobo:
1 rename module name from data aggregator to event generator
2 update the import expression to adapt #1
3 update readme file


fraenkel:
You forgot to update the app-autoscaler/README.md

PR #86, opened 2016-09-02
take scaling action when alarm is received

boyang9527:
[#119505369]

- server to handle triggers from eventgen
- add get/set instance number in cf client
- actively refresh token in cf client
- add GetAppPolicy in shared db module
- change test case of RetrievePolicys in db module
- share models package
- change eventgen to use type metric/instanceMetrics  in shared models

not in scope

- scaling engine configuration
- scaling engine main
- handling cool down


src/cf/app.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+const (
+	TokenTypeBearer = "bearer"
+	PathAppSummary  = "/v2/apps/{appid}/summary"
+	PathAppUpdate   = "/v2/apps/{appid}"
+)
+
+func (c *cfClient) GetAppInstances(appId string) (int, error) {
+	url := c.conf.Api + strings.Replace(PathAppSummary, "{appid}", appId, 1)
```
fraenkel: Let us use the PathAppUpdate instead. Summary is more expensive and unnecessary to get the instance count.

src/cf/client.go
```diff
@@ -186,6 +199,21 @@ func (c *cfClient) GetTokens() Tokens {
 	return c.tokens
 }
 
+func (c *cfClient) GetTokensWithRefresh() Tokens {
+	if c.isTokenToBeExpired() {
```
fraenkel: Please don't do double-lock checking. Just lock and check.

src/cf/config.go
```diff
@@ -18,6 +20,18 @@ func (conf *CfConfig) Validate() error {
 		return fmt.Errorf("Configuration error: cf api is empty")
 	}
 
+	apiUrl, err := url.Parse(conf.Api)
+	if err != nil {
+		return fmt.Errorf("Configuration error: cf api is invalid")
+	}
+	if apiUrl.Scheme == "" {
+		apiUrl.Scheme = "https"
```
fraenkel: We should require the scheme. Having them guess our default isn't great.

src/cf/config_test.go
```diff
@@ -33,7 +33,28 @@ var _ = Describe("Config", func() {
...
+			BeforeEach(func() {
+				conf.Api = "a.com%"
+			})
+
+			It("should error", func() {
+				Expect(err).To(MatchError("Configuration error: cf api is invalid"))
+			})
+		})
+
+		Context("when api valid but not regulated", func() {
```
fraenkel: but ends with a /

src/engine/server/scaling_handler.go
```diff
@@ -0,0 +1,117 @@
...
+	var newInstances int
+	newInstances, err = h.Scale(trigger)
+	if err != nil {
+		h.logger.Error("handle-scale-perform-scaling-action", err, lager.Data{"trigger": trigger})
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Internal-server-error",
+			Message: "Error taking scaling action"})
+		return
+	}
+	handlers.WriteJSONResponse(w, http.StatusCreated, models.AppProperties{Instances: newInstances})
```
fraenkel: Is this really a 201? Shouldn't this be a 200? Since it can be either created or updated.

src/engine/server/scaling_handler_test.go
```diff
@@ -0,0 +1,303 @@
...
+		newInstances int
+		adjustment   string
+		trigger      *models.Trigger
+		buffer       *gbytes.Buffer
+	)
+
+	BeforeEach(func() {
+		cfc = &fakes.FakeCfClient{}
+
+		logger := lager.NewLogger("scaling-handler-test")
```
fraenkel: Use lagertest.NewTestLogger

src/engine/server/server.go
```diff
@@ -0,0 +1,25 @@
...
+	"db"
+
+	"code.cloudfoundry.org/lager"
+	"github.com/gorilla/mux"
+	"github.com/tedsuo/ifrit"
+	"github.com/tedsuo/ifrit/http_server"
+)
+
+const (
+	PathScalingActions = "/v1/scaling_actions"
```
fraenkel: Do we want to call this /v1/scale? or /v1/scalingengine/scale?

src/eventgenerator/aggregator/metricPoller_test.go
```diff
@@ -3,8 +3,8 @@ package aggregator_test
 import (
 	. "eventgenerator/aggregator"
 	. "eventgenerator/appmetric"
-	. "metricscollector/metrics"
-	. "metricscollector/server"
+	// . "metricscollector/server"
```
fraenkel: Why is this here?

src/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
```diff
@@ -106,12 +106,11 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	cfg.Db.MetricsDbUrl = os.Getenv("DBURL")
 	cfg.Db.PolicyDbUrl = os.Getenv("DBURL")
 
-	cfg.Collector.PollInterval = 10 * time.Second
-	cfg.Collector.RefreshInterval = 30 * time.Second
+	cfg.Collector.PollInterval = 10
```
fraenkel: Do you really mean 10 ns?

src/metricscollector/config/config.go
```diff
@@ -4,17 +4,16 @@ import (
...
 
 	"cf"
 )
 
 const (
-	DefaultLoggingLevel                  = "info"
-	DefaultRefreshInterval time.Duration = 60 * time.Second
-	DefaultPollInterval    time.Duration = 30 * time.Second
+	DefaultLoggingLevel    = "info"
+	DefaultRefreshInterval = 60
```
fraenkel: Why are we switching to int from Duration?  Duration gives us more flexibility.

src/cf/app.go
boyang9527: sure, will change to /v2/apps/{appid}

src/cf/client.go
boyang9527: this is used to avoid another  refresh if other thread has already refreshed the token and changed the expire time.

We can remove the outer check while this will have more overhead since  no matter the token is to be expired or not we need lock/unlock

src/cf/config.go
boyang9527: I will change it to fail the validation if there is no scheme

src/cf/config_test.go
boyang9527: sure will change to make the work more precise

src/engine/server/scaling_handler.go
boyang9527: 201 is used to indicate that "a scaling action has been created".  Since we don't observe whether it is actually scaled to the number that we expect.

and CF API PUT /v2/apps/:guid returns "201 created"

src/engine/server/scaling_handler_test.go
boyang9527: will change to lagertest

src/cf/client.go
fraenkel: So I am not convinced that you don't have a race without always locking.

src/engine/server/scaling_handler.go
fraenkel: which is just odd.
I prefer 200 which is what we discussed for the scheduler as well.

src/engine/server/server.go
boyang9527: I am trying to make the API a POST to a resource, not a verb/action.  I am good to change to /v1/scale, 

or maybe POST /v1/apps/{app-id}/scale

fraenkel: Yes, that sounds better.

src/eventgenerator/aggregator/metricPoller_test.go
boyang9527: I forgot to remove the commented line after testing. Will remove it

The ErrorResponse struct was moved to "models" package, eventgen  does not need to import from metricscollector any more

src/metricscollector/config/config.go
boyang9527: It might be better to set time interval in seconds in the configuration file.  Do we want to have a finer grain?

fraenkel: Yes, for testing.
For the end user, they will only see seconds because we will generate the config.

src/cf/client.go
boyang9527: yes there is potential data race for  expire time. I will remove the outer check.

src/engine/server/scaling_handler.go
boyang9527: ok.. sure I can change to 200

src/metricscollector/config/config.go
boyang9527: ok, I will change back

boyang9527:
@fraenkel  
I did another commit based on comments. Please review.

fraenkel:
Merged with some modifications

PR #87, opened 2016-09-05
128903155: Unit test failure: Service instance test failure

kanekoh:


fraenkel:
Thank you for finding the issues. I will rework what you have found to align better with the current model. We are trying to avoid "done". I am not convinced about the require check which I will also verify.

PR #91, opened 2016-09-08
Introduce updated_at column to policy db

pradyutsarma:
1. Introduce the column in db changelog and sequelize model
2. Trigger value on insert/update of row
3. Requisite tests

### DIFF
--- a/api/test/unit/models/policy.test.js
+++ b/api/test/unit/models/policy.test.js
@@ -13,8 +13,9 @@ describe('Create a Policy', function() {
   });
 
   context ('Test create policies', function () {
-	  beforeEach(function() {
+	  beforeEach(function(done) {
 		policy.truncate();
+		done();
 	  });
 
 	  it('Should create a policy for app id 99999 successfully', function() {
@@ -40,8 +41,9 @@ describe('Create a Policy', function() {
   });
   
   context ('Test create and update policies with updated_at checks', function () {
-	  beforeEach(function() {
+	  beforeEach(function(done) {
 		policy.truncate();
+		done();
 	  });
 
 	  it('Create a policy with app_id 99999 and then updates the policy for the same', function(done) {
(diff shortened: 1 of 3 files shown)


api/test/unit/models/policy.test.js
```diff
@@ -11,23 +12,57 @@ describe('Create a Policy', function() {
...
+		        expect(error).to.have.deep.property('message').equal('Validation error');
+		        done();
+		    });
+	    });
+	  });	  
   });
+  
+  context ('Test create and update policies with updated_at checks', function () {
+	  beforeEach(function() {
+		policy.truncate();
```
fraenkel: missing return

api/test/unit/models/policy.test.js
```diff
@@ -11,23 +12,57 @@ describe('Create a Policy', function() {
     fakePolicy = JSON.parse(fs.readFileSync(__dirname+'/../fakePolicy.json', 'utf8'));
   });
 
-  beforeEach(function() {
-    policy.truncate();
-  })
+  context ('Test create policies', function () {
+	  beforeEach(function() {
+		policy.truncate();
```
fraenkel: missing return

fraenkel:
Is there a separate story to flow this as some type of etag or something so other components have some idea of what "version" of data they are using?

api/test/unit/models/policy.test.js
pradyutsarma: Fixed.

pradyutsarma:
@fraenkel we discussed this in the last call. Maybe we can introduce some type of etag in the GET API response header generated out of the attributes including the updated_at data. But then, will the other components query for this per app?

fraenkel:
We need something. But more importantly, is the usage for scheduler and other components which will use it to determine if the underlying policy has changed/replaced.
We do need something to track this.

pradyutsarma:
I will create a story. Let's discuss there.

fraenkel:
Merged

PR #94, opened 2016-09-15
130344867: Remove the metrics data from metric collector db based on the cut off days

kanekoh:
[#130344867]

- Implemented prune functionality for metric collector - applicationmetrics table.
- Added corresponding tests


src/metricscollector/appmetricprune/prune/prune.go
```diff
@@ -0,0 +1,33 @@
+package prune
```
fraenkel: Why does this exist?

src/metricscollector/appmetricprune/prune/prune.go
```diff
@@ -0,0 +1,33 @@
...
+}
+
+func NewPrune(logger lager.Logger, metricDB db.MetricsDB) *Prune {
+	return &Prune{
+		logger:   logger,
+		metricDB: metricDB,
+	}
+}
+
+func (pr *Prune) PruneMetricsOlderThan(cutoffDays int) {
```
fraenkel: This is the only relevant code

src/metricscollector/appmetricprune/prune/prune.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+const TokenTypeBearer = "bearer"
+
+type Prune struct {
+	logger   lager.Logger
+	metricDB db.MetricsDB
+}
+
+type ErrorResponse struct {
```
fraenkel: Why is this here?

src/metricscollector/appmetricprune/prune/prune.go
fraenkel: The naming isn't good. We don't need it to live so far down.
prune.go should just like in the metricscollector and call the struct Pruner.

You will then wire this in as an ifrit.Process that runs 1,2 or 4 times a day. I don't really care on the interval. We can make that configurable too.

src/metricscollector/appmetricprune/prune/prune_test.go
```diff
@@ -0,0 +1,56 @@
...
+		logger          lager.Logger
+		prune           *Prune
+		timeBeforePrune time.Time
+		timeAfterPrune  time.Time
+		cutoffDays      int
+	)
+
+	BeforeEach(func() {
+
+		logger = lager.NewLogger("prune-test")
```
fraenkel: Why does this need to be exposed? its only used when we create the Pruner below.

src/metricscollector/appmetricprune/prune/prune.go
```diff
@@ -0,0 +1,33 @@
...
+	return &Prune{
+		logger:   logger,
+		metricDB: metricDB,
+	}
+}
+
+func (pr *Prune) PruneMetricsOlderThan(cutoffDays int) {
+	timestamp := time.Now().AddDate(0, 0, -cutoffDays).UnixNano()
+
+	pr.metricDB.PruneMetrics(timestamp)
```
fraenkel: Error handling?

src/metricscollector/appmetricprune/prune/prune_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+		Context("prune metric data", func() {
+			BeforeEach(func() {
+				cutoffDays = 20
+
+				prune = NewPrune(logger, metricDB)
+			})
+
+			JustBeforeEach(func() {
+				timeBeforePrune = time.Now()
```
fraenkel: You can just do all of this in the It()

src/metricscollector/appmetricprune/prune/prune_test.go
```diff
@@ -0,0 +1,56 @@
...
+				prune.PruneMetricsOlderThan(cutoffDays)
+				timeAfterPrune = time.Now()
+			})
+
+			It("It has the the correct timestamp cutoff for pruning", func() {
+				Expect(metricDB.PruneMetricsCallCount()).To(Equal(1))
+
+				Expect(metricDB.PruneMetricsArgsForCall(0) >= timeBeforePrune.AddDate(0, 0, -cutoffDays).UnixNano()).To(Equal(true))
```
fraenkel: Just use BeTemporally

src/metricscollector/config/config.go
```diff
@@ -54,12 +55,21 @@ var defaultCollectorConfig = CollectorConfig{
 	PollInterval:    DefaultPollInterval,
 }
 
+type DataPruneConfig struct {
+	CutoffDays int `yaml:"cutoff_days"`
+}
+
+var defaultDataPruneConig = DataPruneConfig{
```
fraenkel: Just so you know, I am not convinced I want the pruning to be done by the metricscollector.

I think we will have a standalone pruner that will do it for all the DBs. So this code should just move to src/autoscaler/pruner as the top level package. The pruners themselves can live at the root and the main would be autoscaler/pruner/cmd/pruner.

ghaih: Just thinking aloud for clarity,

We no longer want to have any pruning code in respective packages i.e in metricscollector or data aggregator etc

The approach is to have 

- autoscaler/src/pruner/cmd/ pruner.go -> running the specific ifrit pruner processesas follows
 -  autoscaler/src/pruner/metriccollector_metric_pruner.go
 - autoscaler/src/pruner/aggregator_metric_pruner.go etc

fraenkel: you are duplicating yourself with your naming. You are already in the pruner package so metriccollector or whatever db table name .go is sufficient, pruner is redundant.
And these things are doing nothing more than driving the underlying DB code or in the case of scheduler, making the remote call.

ghaih: Yeah, those were not intended names, just to put across our understanding for clarity to depict their jobs used those names:)

So we will go ahead with this structure which is more clear. We did struggle with the previous approach.

src/metricscollector/appmetricprune/prune/prune_test.go
ghaih: It is not time but int64 so will change to use BeNumerically?

src/metricscollector/appmetricprune/prune/prune.go
ghaih: You mean have the cutoffDays in the struct Pruner

src/metricscollector/appmetricprune/prune/prune_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	BeforeEach(func() {
+
+		logger = lager.NewLogger("prune-test")
+
+		metricDB = &fakes.FakeMetricsDB{}
+
+	})
+	Describe("when a request for prune old metric data comes", func() {
+
```
ghaih: Didn't get it. Check logs for any errors?

src/metricscollector/appmetricprune/prune/prune.go
fraenkel: yes, it doesn't change. You need to think about how this is really going to be used. I envision you create a Pruner that is just called on some time sequence. The cutoff is always the same.

src/metricscollector/appmetricprune/prune/prune_test.go
fraenkel: Where are the tests when errors occur?

src/autoscaler/pruner/cmd/pruner/pruner_test.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gbytes"
+	. "github.com/onsi/gomega/gexec"
+	"io/ioutil"
+	"os"
+)
+
+var _ = Describe("Pruner", func() {
```
fraenkel: We actually don't verify that the MetricsPruner starts.

src/autoscaler/pruner/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+	Level: DefaultLoggingLevel,
+}
+
+type DbConfig struct {
+	MetricsDbUrl string `yaml:"metrics_db_url"`
+}
+
+type PrunerConfig struct {
+	IntervalInHours int `yaml:"interval_in_hours"`
+	CutoffDays      int `yaml:"cutoff_days"`
```
fraenkel: I have a feeling that the cutoff will be different per DB but we can leave this here and fix it once we add a second DB/Table.

src/autoscaler/pruner/metricsdb.go
```diff
@@ -0,0 +1,69 @@
...
+		metricsDB:  metricsDB,
+		interval:   time.Duration(interval) * time.Hour,
+		cutoffDays: cutoffDays,
+		clock:      clock,
+		doneChan:   make(chan bool),
+	}
+}
+
+func (mdp *MetricsDBPruner) Start() {
+	mdp.ticker = mdp.clock.NewTicker(mdp.interval)
```
fraenkel: why does the ticker live here? It should be scoped to the startMetricPrune which is really pruneMetrics().

src/autoscaler/pruner/metricsdb_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+		It("Stop the pruner", func() {
+			fclock.Increment(TestRefreshInterval)
+			Eventually(metricsDB.PruneMetricsCallCount).Should(Equal(2))
+
+			pruner.Stop()
+			Eventually(buffer).Should(gbytes.Say("metrics-db-pruner-stopped"))
+
+			fclock.Increment(TestRefreshInterval)
+			Eventually(metricsDB.PruneMetricsCallCount).Should(Equal(2))
```
fraenkel: You mean Consistently.

src/autoscaler/pruner/config/config.go
ghaih: Yes thought about it. Will change as part of data aggregator prune story.

src/autoscaler/pruner/cmd/pruner/pruner_test.go
ghaih: Will add the check in prune_suite_test and a another in pruner_test.

src/autoscaler/pruner/metricsdb_test.go
ghaih: Yes.

src/autoscaler/pruner/metricsdb.go
ghaih: Moving it to startMetricPrune results in 2 data race. So will have it in Start.

fraenkel:
Merged with some minor changes

PR #98, opened 2016-09-21
Aggregator can execute independently from policyPoller

qibobo:
[#119505309]

### DIFF
--- a/src/autoscaler/eventgenerator/aggregator/aggregator.go
+++ b/src/autoscaler/eventgenerator/aggregator/aggregator.go
@@ -14,12 +14,10 @@ import (
 
 type Aggregator struct {
 	logger                    lager.Logger
-	metricCollectorUrl        string
 	doneChan                  chan bool
 	appChan                   chan *model.AppMonitor
 	policyPoller              *PolicyPoller
 	metricPollerArray         []*MetricPoller
-	policyDatabase            db.PolicyDB
 	appMetricDatabase         db.AppMetricDB
 	evaluationManager         *generator.AppEvaluationManager
 	cclock                    clock.Clock
@@ -30,15 +28,13 @@ type Aggregator struct {
 
 func NewAggregator(logger lager.Logger, clock clock.Clock, aggregatorExecuteInterval time.Duration, policyPollerInterval time.Duration, policyDatabase db.PolicyDB, appMetricDatabase db.AppMetricDB, metricCollectorUrl string, metricPollerCount int, evaluationManager *generator.AppEvaluationManager, appMonitorChan chan *model.AppMonitor) *Aggregator {
 	aggregator := &Aggregator{
-		logger:             logger.Session("Aggregator"),
-		metricCollectorUrl: metricCollectorUrl,
-		doneChan:           make(chan bool),
-		appChan:            appMonitorChan,
-		metricPollerArray:  []*MetricPoller{},
-		policyDatabase:     policyDatabase,
-		appMetricDatabase:  appMetricDatabase,
-		evaluationManager:  evaluationManager,
-		cclock:             clock,
+		logger:            logger.Session("Aggregator"),
+		doneChan:          make(chan bool),
+		appChan:           appMonitorChan,
+		metricPollerArray: []*MetricPoller{},
+		appMetricDatabase: appMetricDatabase,
+		evaluationManager: evaluationManager,
+		cclock:            clock,
 		aggregatorExecuteInterval: aggregatorExecuteInterval,
 		appMonitorArray:           []*model.AppMonitor{},
 	}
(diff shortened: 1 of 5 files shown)


src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -8,23 +8,28 @@ import (
...
-	metricPollerArray  []*MetricPoller
-	policyDatabase     db.PolicyDB
-	appMetricDatabase  db.AppMetricDB
-	evaluationManager  *generator.AppEvaluationManager
+	logger                    lager.Logger
+	metricCollectorUrl        string
+	doneChan                  chan bool
+	appChan                   chan *model.AppMonitor
+	policyPoller              *PolicyPoller
+	metricPollerCount         int
```
fraenkel: Why do we need this?
You have many pieces of data that are unneeded.

src/autoscaler/eventgenerator/aggregator/aggregator_test.go
```diff
@@ -139,15 +139,17 @@ var _ = Describe("Aggregator", func() {
 		})
 		Context("when there are data in triggerMap", func() {
 			JustBeforeEach(func() {
-				aggregator.ConsumePolicy(policyMap, appChan)
+				aggregator.ConsumePolicy(policyMap, appMonitorChan)
+				aggregator.Start()
 				evaluationManager.Start()
-				Eventually(clock.WatcherCount).Should(Equal(1))
+				Eventually(clock.WatcherCount).Should(Equal(3)) //policyPoller:1,aggregator:1,evaluationManager:1
 				clock.Increment(1 * testEvaluateInteval)
```
fraenkel: Why both? Shouldn't it be the max between them?

src/autoscaler/eventgenerator/aggregator/aggregator_test.go
```diff
@@ -232,16 +234,17 @@ var _ = Describe("Aggregator", func() {
 	})
 	Context("Start", func() {
 		JustBeforeEach(func() {
-			aggregator = NewAggregator(logger, clock, testPolicyPollerInterval, policyDatabase, appMetricDatabase, metricServer.URL(), metricPollerCount, evaluationManager, appMonitorChan)
+			aggregator = NewAggregator(logger, clock, testAggregatorExecuteInterval, testPolicyPollerInterval, policyDatabase, appMetricDatabase, metricServer.URL(), metricPollerCount, evaluationManager, appMonitorChan)
 			aggregator.Start()
+			Eventually(clock.WatcherCount).Should(Equal(2)) //policyPoller:1,aggregator:1
```
fraenkel: How does this work? How do we know it's the right 2 since you expect 3 above?

src/autoscaler/eventgenerator/aggregator/aggregator.go
qibobo: Yes. I will remove the unneeded variables.

src/autoscaler/eventgenerator/aggregator/aggregator_test.go
qibobo: updated.

src/autoscaler/eventgenerator/aggregator/aggregator_test.go
qibobo: Above when testing ConsumePolicy we need ensure the appMonitors and triggers which are parsed from policies to be both checked. So aggregator and appEvaluationManager should both be started and the ticker count here is 3(aggregator,policyPoller and appEvaluationManager). And for here we test aggregator should be started. So I think the appEvaluationManager need not be started though it can be started. So the ticker count should be 2. I updated it to 3 for consistence.

src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -18,7 +18,6 @@ type Aggregator struct {
 	doneChan                  chan bool
 	appChan                   chan *model.AppMonitor
 	policyPoller              *PolicyPoller
-	metricPollerCount         int
 	metricPollerArray         []*MetricPoller
 	policyDatabase            db.PolicyDB
```
fraenkel: Sorry but there are more things to remove....

src/autoscaler/eventgenerator/generator/appEvaluationManager.go
```diff
@@ -21,7 +21,6 @@ type AppEvaluationManager struct {
 	triggerChan      chan []*model.Trigger
```
fraenkel: There are more things to remove.

qibobo: Removed. Hope no more is here. Thank you!

PR #100, opened 2016-09-22
When an alarm fires during the cool down period, no action is taken

boyang9527:
[#119505393]

- add db interface to store cooldown expire time
- skip scaling if the app is in cool down period
- write scaling Igored record to history db, so that it can be quried

### DIFF
--- a/src/autoscaler/scalingengine/server/applock_test.go
+++ b/src/autoscaler/scalingengine/server/applock_test.go
@@ -1,0 +1,106 @@
+package server_test
+
+import (
+	. "autoscaler/scalingengine/server"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"time"
+)
+
+var (
+	ch      chan string
+	appLock *AppLock
+	err     interface{}
+)
+
+var _ = Describe("Applock", func() {
+	BeforeEach(func() {
+		appLock = NewAppLock()
+		ch = make(chan string)
+	})
+
+	AfterEach(func() {
+		close(ch)
+	})
+
+	Describe("Lock", func() {
+		JustBeforeEach(func() {
+			go lockApp(appLock, "an-app-id")
+		})
+		Context("when acquiring the lock of an app", func() {
+			It("gets the lock", func() {
+				Eventually(hasAcquiredLock).Should(BeTrue())
+			})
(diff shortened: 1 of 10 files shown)


src/autoscaler/scalingengine/server/applock_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+var (
+	ch      chan string
+	appLock *AppLock
+	err     interface{}
+)
+
+var _ = Describe("Applock", func() {
+	BeforeEach(func() {
+		appLock = NewAppLock()
```
boyang9527: @fraenkel 
there is a data race between line 20(write appLock) and line 30/95 (read appLock).  Theoretically there is guarantee that line 20 is executed before line 30 go routine is created and line 95 is executed.  Seems like it is because they are too close?  Is there good way to handle it without introducing weird code?

fraenkel: I see problems with AppLock.
1. At runtime, we keep adding a lock for each AppId but I never see where we drop all the locks, i.e., do any cleanup.

2. As far as the test cases are concerned, the problem you have is that you have a goroutine using the AppLock which uses it after the next test starts. Your afterEach only guarantees that the goroutine will do something, but it doesn't guarantee that it happens before the next test.
You can either pass the applock directly to the goroutine, or wait for the goroutine to exit properly before allowing the next test to run.

boyang9527: for 1)   Do we have concerns of holding too much memory, only one lock is created per app? how do we decide when to remove the Mutex from map? 
For 2)  solved by passing AppLock as an argument. I did not realize the data race is across test cases.

boyang9527:
I revised the commit to get rid of the data race.

src/autoscaler/scalingengine/server/applock_test.go
fraenkel: Ok, so now taking a closer look, what you did conceptually is correct but not quite the right implementation. You are layering a blocking request within something that expects to be periodic.

hasAcquiredLock should just be 
```
select {
  case <-ch:
    return true
  default:
    return false
}
```
Time shouldn't matter if you are using Eventually/Consistently around it.

PR #102, opened 2016-09-23
Use time.Duration in yaml config for intervals and fix unclosed respo…

boyang9527:
…nse body in test cases

[#130876133]

- change interval config to time.Duration
- replace  "github.com/cloudfoundry-incubator/candiedyaml" with  "gopkg.in/yaml.v2"
- fix unclosed response body in test cases

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,9 +10,6 @@
 [submodule "src/github.com/onsi/gomega"]
 	path = src/github.com/onsi/gomega
 	url = https://github.com/onsi/gomega
-[submodule "src/github.com/cloudfoundry-incubator/candiedyaml"]
-	path = src/github.com/cloudfoundry-incubator/candiedyaml
-	url = https://github.com/cloudfoundry-incubator/candiedyaml
 [submodule "src/github.com/cloudfoundry/noaa"]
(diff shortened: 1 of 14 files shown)


src/autoscaler/metricscollector/config/config_test.go
```diff
@@ -39,29 +39,34 @@ server:
 			})
 
 			It("returns an error", func() {
-				Expect(err).To(BeAssignableToTypeOf(&candiedyaml.ParserError{}))
+				Expect(err).To(MatchError(MatchRegexp("yaml:*")))
```
fraenkel: This is not a correct regexp.

.gitmodules
```diff
@@ -10,9 +10,6 @@
 [submodule "src/github.com/onsi/gomega"]
```
fraenkel: You didn't submodule in goyaml.

src/autoscaler/metricscollector/config/config_test.go
boyang9527: sorry my careless, correct it with .*

.gitmodules
boyang9527: the package was already there. It was added by aggregator/eventgen before.

PR #104, opened 2016-09-26
event-generator should not send trigger when there is no metric from …

qibobo:
…metric-collector db

[#119505309]
    1 when there is no metric returned from metric-collector,set
appMetric’s value to
    -1; and evaluator will not send trigger alarm for this kind of
metric.
    2 update readme file


src/autoscaler/eventgenerator/README.md
```diff
@@ -14,7 +14,7 @@ Data Aggregator is one of the components of CF `app-autoscaler`.
 1. change directory to `app-autoscaler`
 1. run `source .envrc`
 1. pull the submodules : `git submodule update --init --recursive`
-1. change directory to `src/eventgenerator`
+1. change directory to `src/autoscaler/eventgenerator`
```
fraenkel: This seems wrong.  It makes sense from the autoscaler/README.md but if you are reading this README, you are already in the right directory.

src/autoscaler/eventgenerator/aggregator/metricPoller.go
```diff
@@ -104,17 +102,17 @@ func (m *MetricPoller) doAggregate(appId string, metricType string, metrics []*m
 		avgAppMetric = &model.AppMetric{
 			AppId:      appId,
 			MetricType: metricType,
-			Value:      0,
+			Value:      -1,
 			Unit:       "",
-			Timestamp:  0,
+			Timestamp:  time.Now().UnixNano(),
```
fraenkel: This seems wrong.
We are walking over a set of metrics. The current code will assign a different timestamp to each. Logically, they should all have the same timestamp.

src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -70,6 +70,10 @@ func (e *Evaluator) doEvaluate(triggerArray []*model.Trigger) {
 			return
 		}
 		for _, appMetric := range appMetricList {
+			if appMetric.Value < 0 {
+				e.logger.Info("should not send trigger alarm to scaling engine because there is nil-value metric", lager.Data{"trigger": trigger, "appMetric": appMetric})
```
fraenkel: who is this for?
if our appMetric.Value is less than 0, why would an admin care? What are they supposed to do when they see this?

src/autoscaler/eventgenerator/aggregator/metricPoller.go
```diff
@@ -104,17 +102,17 @@ func (m *MetricPoller) doAggregate(appId string, metricType string, metrics []*m
 		avgAppMetric = &model.AppMetric{
 			AppId:      appId,
 			MetricType: metricType,
-			Value:      0,
+			Value:      -1,
```
fraenkel: Would it make more sense to use nil as no value rather than -1?

qibobo: The type of Value is int64, assign a nil to it will alert compile error.

fraenkel: I know that, but you can also make Value a *int64 if we need nil.

src/autoscaler/eventgenerator/generator/evaluator.go
qibobo: When aggregator get a empty list from metric-collector, it may mean that the application for aggregating may have some problems that the metric-collector can not retrieve its metrics in the last statWindow. A nil record should be saved so that when do trigger evaluation, if there is a nil record, no alarm should be sent. Above I set the nil record's value to -1, so here it check whether the value is LT 0.

fraenkel: Again, as an admin, if I saw this message, what do I do?
The message also says there is a nil-value but in reality it is just < 0.

qibobo: Yes. Will change the Value to *int64 and check whether it is nil.

src/autoscaler/eventgenerator/aggregator/metricPoller.go
qibobo: We aggregate a set of metrics and should get an AVG metric. The timestamps of this metric set range from T1 to Tn, and for the AVG one , what its timestamp should be? T1 or Tn or NOW(). Also, if the set is empty and we will still save a nil-value metric, and there is no T1 and Tn. So here I set the AVG one's timestamp to NOW().

fraenkel: My point is that you can define the timestamp before the loop and use a single common value for all generated metrics.

src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -70,27 +70,27 @@ func (e *Evaluator) doEvaluate(triggerArray []*model.Trigger) {
 			return
 		}
 		for _, appMetric := range appMetricList {
-			if appMetric.Value < 0 {
+			if appMetric.Value == nil {
 				e.logger.Info("should not send trigger alarm to scaling engine because there is nil-value metric", lager.Data{"trigger": trigger, "appMetric": appMetric})
```
fraenkel: What does this mean to the operator?

qibobo: When it get an empty metric list from metric-collector, it sets the avg AppMetric.Value to nil. So here check the nil value. If it is nil , no trigger should be sent.

fraenkel: I know what the code is doing. You have marked this as an info message. I want to know what an operator is supposed to do when they see this message because they aren't going to understand it. The message describes what the code is doing not what the operator should be concerned about.

src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -71,7 +71,7 @@ func (e *Evaluator) doEvaluate(triggerArray []*model.Trigger) {
 		}
 		for _, appMetric := range appMetricList {
 			if appMetric.Value == nil {
-				e.logger.Info("should not send trigger alarm to scaling engine because there is nil-value metric", lager.Data{"trigger": trigger, "appMetric": appMetric})
+				e.logger.Info("should not send trigger alarm to scaling engine because there is nil-value metric( When aggrgator gets an empty metric array from metric-collector in a statWindow,it creates an AppMetric instance with nil Value)", lager.Data{"trigger": trigger, "appMetric": appMetric})
```
fraenkel: Huh?
Did you really mean to do this?
Again, what does this mean to the operator? When they see this message, what are they supposed to do?
Why is this one Info and the others Debug?

PR #105, opened 2016-10-03
Adopt to JSONLayout for scheduler logs

pradyutsarma:


pradyutsarma:
Will create a separate PR for this story.

PR #107, opened 2016-10-06
set/remove active schedule and change app instances accordingly

boyang9527:
[#119511307]

- REST API for scheduler to set/remove active schedule
- manage the active schedules in memory
- change app instances when setting/removing active schedule
    - change app instances to be initial-min-instances when setting an active schedule
    - enforce default instance limit when removing an active schedule

### DIFF
--- a/src/autoscaler/scalingengine/schedule/schedule_test.go
+++ b/src/autoscaler/scalingengine/schedule/schedule_test.go
@@ -1,0 +1,240 @@
+package schedule_test
+
+import (
+	"autoscaler/models"
+	"autoscaler/scalingengine/fakes"
+
+	. "autoscaler/scalingengine/schedule"
+
+	"code.cloudfoundry.org/lager/lagertest"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+
+	"errors"
+)
+
+var _ = Describe("Schedule", func() {
+
+	var (
+		schedules      *AppSchedules
+		cfc            *fakes.FakeCfClient
+		policyDB       *fakes.FakePolicyDB
+		buffer         *gbytes.Buffer
+		activeSchedule ActiveSchedule
+		err            error
+	)
+
+	BeforeEach(func() {
+		cfc = &fakes.FakeCfClient{}
+		policyDB = &fakes.FakePolicyDB{}
+		logger := lagertest.NewTestLogger("schedule-test")
+		buffer = logger.Buffer()
+		schedules = NewAppSchedules(logger, cfc, policyDB)
+		activeSchedule = ActiveSchedule{
+			ScheduleId:      "a-schedule-id",
+			InstanceInitial: 5,
+			InstanceMin:     3,
+			InstanceMax:     10,
+		}
+	})
+
+	Describe("SetActiveSchedule", func() {
+		JustBeforeEach(func() {
+			err = schedules.SetActiveSchedule("an-app-id", activeSchedule)
+		})
+
+		Context("when no active schedule exists for the app", func() {
+			It("sets the active schedule and initial instances", func() {
+				Expect(err).NotTo(HaveOccurred())
+				appId, instances := cfc.SetAppInstancesArgsForCall(0)
+				Expect(appId).To(Equal("an-app-id"))
+				Expect(instances).To(Equal(5))
+
+				schedule, exists := schedules.GetActiveSchedule("an-app-id")
+				Expect(exists).To(BeTrue())
+				Expect(schedule).To(Equal(activeSchedule))
+			})
+
+		})
+
+		Context("when active schedule exists for the app", func() {
+			BeforeEach(func() {
+				err = schedules.SetActiveSchedule("an-app-id", ActiveSchedule{
+					ScheduleId:      "currrent-schedule-id",
+					InstanceInitial: 6,
+					InstanceMin:     2,
+					InstanceMax:     8,
+				})
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("Sets the new active scheule and logs the info", func() {
+				Expect(err).NotTo(HaveOccurred())
+				schedule, exists := schedules.GetActiveSchedule("an-app-id")
+				Expect(exists).To(BeTrue())
+				Expect(schedule).To(Equal(activeSchedule))
+
+				appId, instances := cfc.SetAppInstancesArgsForCall(1)
+				Expect(appId).To(Equal("an-app-id"))
+				Expect(instances).To(Equal(5))
+
+				Eventually(buffer).Should(gbytes.Say("active schedule exists"))
+			})
+		})
+
+		Context("when fails to set initial instances number", func() {
+			BeforeEach(func() {
+				cfc.SetAppInstancesReturns(errors.New("test error"))
+			})
+			It("should sets the active schedule and return an error", func() {
+				Expect(err).To(HaveOccurred())
+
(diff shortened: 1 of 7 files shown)


src/autoscaler/scalingengine/schedule/schedule_test.go
```diff
@@ -0,0 +1,240 @@
...
+				err = schedules.SetActiveSchedule("an-app-id", ActiveSchedule{
+					ScheduleId:      "currrent-schedule-id",
+					InstanceInitial: 6,
+					InstanceMin:     2,
+					InstanceMax:     8,
+				})
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("Sets the new active scheule and logs the info", func() {
```
fraenkel: I agree it's a thing, is this really testing anything other than logging?

src/autoscaler/scalingengine/schedule/schedules.go
```diff
@@ -0,0 +1,104 @@
...
+	as.schedules[appId] = schedule
+	as.lock.Unlock()
+
+	logger := as.logger.WithData(lager.Data{"appId": appId})
+
+	if exists {
+		logger.Info("set-active-schedule", lager.Data{"message": "active schedule exists", "currentSchedule": currentSchedule})
+	}
+
+	err := as.cfClient.SetAppInstances(appId, schedule.InstanceInitial)
```
fraenkel: Is InstanceInitial optional?  We may need to think about how this is used.
If you are below the min, you should get the initialmin.
If you are above max, you should go to max?
if you are above initialMin, you should stay where you are.

src/autoscaler/scalingengine/schedule/schedule_test.go
```diff
@@ -0,0 +1,240 @@
...
+				Expect(exists).To(BeFalse())
+			})
+		})
+
+		Context("when  active schedule does not exist", func() {
+			BeforeEach(func() {
+				policyDB.GetAppPolicyReturns(&models.ScalingPolicy{}, nil)
+			})
+
+			It("logs the information", func() {
```
fraenkel: testing logging doesn't test the function.
It("does not error") is what you care about.

src/autoscaler/scalingengine/schedule/schedule_test.go
```diff
@@ -0,0 +1,240 @@
...
+		Context("when active schedule id does not match", func() {
+			BeforeEach(func() {
+				err = schedules.SetActiveSchedule("an-app-id", ActiveSchedule{
+					ScheduleId: "current-schedule-id",
+				})
+				Expect(err).NotTo(HaveOccurred())
+				policyDB.GetAppPolicyReturns(&models.ScalingPolicy{}, nil)
+			})
+
+			It("logs the information and removes the app schedule", func() {
```
fraenkel: Why do we remove it when the schedule-id doesn't match?

src/autoscaler/scalingengine/schedule/schedule_test.go
```diff
@@ -0,0 +1,240 @@
...
+				Expect(instances).To(Equal(6))
+			})
+		})
+
+		Context("when fails to get app policy", func() {
+			BeforeEach(func() {
+				policyDB.GetAppPolicyReturns(nil, errors.New("an error"))
+			})
+
+			It("should removes the schedule and error", func() {
```
fraenkel: This is very strange. If I report an error, I would assume the remove did not happen.
If we are going to remove regardless, then there is no error to report.

src/autoscaler/scalingengine/server/scaling_handler.go
```diff
@@ -255,3 +258,37 @@ func getScalingReason(trigger *models.Trigger) string {
...
+			Code:    "Bad-Request",
+			Message: "Incorrect active schedule in request body",
+		})
+		return
+	}
+
+	activeSchedule.ScheduleId = scheduleId
+	h.logger.Debug("set-active-schedule", lager.Data{"appid": appId, "activeSchedule": activeSchedule})
+
+	h.appLock.Lock(appId)
```
fraenkel: Why do we lock?, appSchedules also locks.

src/autoscaler/scalingengine/server/scaling_handler.go
```diff
@@ -255,3 +258,37 @@ func getScalingReason(trigger *models.Trigger) string {
...
+		h.logger.Error("failed-set-active-schedule-unmarshal", err, lager.Data{"appid": appId, "scheduleid": scheduleId})
+		handlers.WriteJSONResponse(w, http.StatusBadRequest, models.ErrorResponse{
+			Code:    "Bad-Request",
+			Message: "Incorrect active schedule in request body",
+		})
+		return
+	}
+
+	activeSchedule.ScheduleId = scheduleId
+	h.logger.Debug("set-active-schedule", lager.Data{"appid": appId, "activeSchedule": activeSchedule})
```
fraenkel: my hunch is that you will want this as an info so we can track this easier.

src/autoscaler/scalingengine/server/scaling_handler.go
```diff
@@ -255,3 +258,37 @@ func getScalingReason(trigger *models.Trigger) string {
...
+
+	h.appLock.Lock(appId)
+	h.appSchedules.SetActiveSchedule(appId, activeSchedule)
+	h.appLock.UnLock(appId)
+}
+
+func (h *ScalingHandler) RemoveActiveSchedule(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	appId := vars["appid"]
```
fraenkel: probably need to log when we are removing it.

src/autoscaler/scalingengine/server/scaling_handler_test.go
```diff
@@ -674,4 +675,60 @@ var _ = Describe("ScalingHandler", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			It("returns 200", func() {
+				Expect(resp.Code).To(Equal(http.StatusOK))
+			})
+		})
+
+		Context("when active schedule is invalid", func() {
+			BeforeEach(func() {
```
fraenkel: why isn't this part of the It?

src/autoscaler/scalingengine/server/scaling_handler_test.go
```diff
@@ -674,4 +675,60 @@ var _ = Describe("ScalingHandler", func() {
...
 	})
+
+	Describe("SetActiveSchedule", func() {
+		JustBeforeEach(func() {
+			handler.SetActiveSchedule(resp, req, map[string]string{"appid": "an-app-id", "schduleid": "a-schdule-id"})
+		})
+
+		Context("when active schedule is valid", func() {
+			BeforeEach(func() {
+				body = []byte(`{"instance_min_count":1, "instance_max_count":5, "initial_min_instance_count":3}`)
```
fraenkel: This is part of the test (It block)

src/autoscaler/scalingengine/server/server_test.go
```diff
@@ -144,4 +143,96 @@ var _ = Describe("Server", func() {
 
 	})
 
+	Context("when setting active schedule", func() {
+		BeforeEach(func() {
```
fraenkel: You have a bunch of Before blocks but if you just focus on what really needs to change, you would just tweak what is needed per test rather than a lot of extra code.

The request should be created in the JustBeforeEach.
Expose the method.

Now you just have to tweak the urlPath, method or body.

boyang9527:
There will be major changes based on the discussions today.  I will submit a new one.  Some of your code comments will become invalid while I will take care of the ones that still apply for the new implementation.

PR #108, opened 2016-10-08
get notified of active schedule start/end and enforce new instance li…

boyang9527:
…mits

[#119511307]

- REST API to handle active schedule start/end
- enforce new instance limit (min, initial min, max) if an active schedule starts
- enforce policy default limit (min, max) is an active schedule ends


fraenkel:
I cleaned up somethings. We still have too many packages. Packages aren't nouns but grouping of things. We should fix this as we add new code.

PR #109, opened 2016-10-12
use active schedule instance min/max when scaling

boyang9527:
[#119511369]

- new DB interface for schedule db
- scaling engine config for shedule db
- move ActiveSchedule type definition to model package
- check active schedule max/min instance when scaling


src/autoscaler/db/sqldb/schedule_sqldb.go
```diff
@@ -0,0 +1,84 @@
...
+	logger := sdb.logger.WithData(lager.Data{"query": query, "appid": appId})
+
+	rows, err := sdb.sqldb.Query(query, appId)
+	if err != nil {
+		logger.Error("failed-get-active-schedule-query", err)
+		return nil, err
+	}
+	defer rows.Close()
+
+	if rows.Next() {
```
fraenkel: How are you guaranteeing that with multiples the first is the correct one? Do we have a guarantee that the active_schedule_id provides any such ordering?

src/autoscaler/db/sqldb/schedule_sqldb_test.go
```diff
@@ -0,0 +1,145 @@
...
+		})
+
+		Context("when there is no active schedule for the given app", func() {
+			It("should not error and return nil", func() {
+				Expect(err).NotTo(HaveOccurred())
+				Expect(activeSchedule).To(BeNil())
+			})
+		})
+
+		Context("when there is active schedule with not null InstanceMinInitial for the given app", func() {
```
fraenkel: with an instance initial min

src/autoscaler/db/sqldb/schedule_sqldb_test.go
```diff
@@ -0,0 +1,145 @@
...
+				Expect(activeSchedule).To(Equal(&models.ActiveSchedule{
+					ScheduleId:         "111111",
+					InstanceMin:        2,
+					InstanceMax:        10,
+					InstanceMinInitial: 5,
+				}))
+			})
+		})
+
+		Context("when there is active schedule with null InstanceMinInitial for the given app", func() {
```
fraenkel: with no instance min initial

src/autoscaler/db/db.go
```diff
@@ -34,3 +34,8 @@ type HistoryDB interface {
 	CanScaleApp(appId string) (bool, error)
 	Close() error
 }
+
+type ScheduleDB interface {
```
fraenkel: How are we pruning these?

src/autoscaler/db/sqldb/schedule_sqldb.go
boyang9527: the active_schedule_id will increase automatically when there is a new record inserted to the table.  When query the table, the following is used to make sure the latest record is the first one.  
```
query := "SELECT active_schedule_id, instance_min_count, instance_max_count, initial_min_instance_count" +
+		" FROM app_scaling_active_schedule WHERE app_id = $1 ORDER BY active_schedule_id DESC"
```
Suppose scheduler will guarantee that  active schedule does not overlap and the latest active schedule record in the table is the one currently valid.

src/autoscaler/db/sqldb/schedule_sqldb_test.go
boyang9527: will fix it

src/autoscaler/db/db.go
boyang9527: when the active schedule end is handled properly, the record will be deleted.  This will be addressed together with user story #131776923

src/autoscaler/db/sqldb/schedule_sqldb.go
fraenkel: If we are going with an increasing id, it might as well reflect a creation_timestamp rather than perform some magic to have a monotonically increasing id.

src/autoscaler/db/db.go
fraenkel: If the end is lost which could happen, we still need this DB to go under the Pruner to catch the case of multiple schedules for an App as well as Apps that have been removed.

boyang9527: Since we have scaling engine to synchronize active_schedule table to catch lost end, it will do delete. When schedulers sync with policy database, it will do delete too.  So looks like we don't need a dedicated pruner to remove stale active_schedule. 

There is a possibility for DB delete failure, which may cause stale data not be deleted, where a pruner will be useful. 

This is out of the scope of this PR. I will submit a fix to add timestamp in the active schedule table and use that to retrieve the latest active schedule.

fraenkel: Just want to make sure we aren't losing things which is why I asked. Separate story is fine.

boyang9527:
I checked in the changes to use timestamp in active_scehdule table , please review.

scheduler/db/scheduler.changelog-master.yaml
```diff
@@ -160,6 +160,11 @@ databaseChangeLog:
                   constraints:
                     primaryKey: true
               - column:
+                  name: timestamp
```
fraenkel: Should be a timestamp type (see api)
The default value for this column should be NOW().
We don't need to insert or touch it.

src/autoscaler/db/sqldb/sqldb_suite_test.go
```diff
@@ -159,17 +159,17 @@ func cleanActiveScheduleTable() error {
 	return e
 }
 
-func insertActiveSchedule(sheduleId int64, appId string, instanceMin, instanceMax, instanceMinInitial, status int) error {
+func insertActiveSchedule(timestamp int64, appId string, instanceMin, instanceMax, instanceMinInitial, status int) error {
 	var e error
 	var query string
 	if instanceMinInitial <= 0 {
-		query = "INSERT INTO app_scaling_active_schedule(active_schedule_id, app_id, instance_min_count, instance_max_count, status) " +
+		query = "INSERT INTO app_scaling_active_schedule(timestamp, app_id, instance_min_count, instance_max_count, status) " +
```
fraenkel: If done right, timestamp shouldn't be provided.

boyang9527:
@fraenkel  committed the changes, please review

PR #115, opened 2016-10-27
Retry metrics polling and requisite tests

pradyutsarma:


fraenkel:
There are multiple issues here.
1. We don't need to add yet another package to do a simple for loop.
2. We are not making the retries configurable. It should just be a constant.
3. We want to be selective on when we retry. If you dig deeper you will see that noaa library is broken in this regard.

pradyutsarma:
Got the first two points.
For 3) Saw this issue https://github.com/cf-tm-bot/loggregator/issues/146 and a PR https://github.com/cloudfoundry/cli/pull/966 for the CF CLI. Do we emulate something similar based on the (retry)error codes then?

fraenkel:
Not quite. Those are strictly for token refresh which happens under the covers. With bosh-lite try different failures like take down your dopplers, or your traffic controller, what errors do you get back?

pradyutsarma:
Closing the PR. Will create a separate one once we have the noaa issue resolved. More details in conversation here https://www.pivotaltracker.com/story/show/127404259

PR #116, opened 2016-11-01
120339579: When a scheduling policy is created, it notifies the scaling engine

kanekoh:
The test case testCreateScheduleAndNotifyScalingEngine is ignored. Details on story in Pivotal Tracker.


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
```diff
@@ -0,0 +1,61 @@
...
+		try {
+			activeScheduleDao.delete(activeScheduleEntity.getId());
+		} catch (DatabaseValidationException dve) {
+			// Refire the job immediately
+			String errorMessage = messageBundleResourceHelper
+					.lookupMessage("database.error.delete.activeschedule.failed", dve.getMessage());
+			logger.error(errorMessage, dve);
+			if(jobExecutionContext.getRefireCount() < maxJobRefireCount) {
+				try {
+					Thread.sleep(jobRefireInterval);
```
fraenkel: This seems bad. Why not just reschedule it with the existing jobData.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/ScalingEngineUtil.java
```diff
@@ -0,0 +1,54 @@
...
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.cloudfoundry.autoscaler.scheduler.entity.ActiveScheduleEntity;
+import org.springframework.http.MediaType;
+import org.springframework.stereotype.Component;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+@Component
+public class ScalingEngineUtil {
```
fraenkel: This isn't a utility class, it's your connection.

scheduler/src/main/resources/application.properties
```diff
@@ -3,4 +3,11 @@
 spring.datasource.driverClassName=org.postgresql.Driver
 spring.datasource.url=jdbc:postgresql://127.0.0.1/autoscaler
 spring.datasource.username=postgres
-spring.datasource.password=postgres
\ No newline at end of file
+spring.datasource.password=postgres
+
+#quartz job
+scalingenginejob.refire.interval=10000
```
fraenkel: What units are these?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleJobTest.java
```diff
@@ -0,0 +1,492 @@
...
+		Mockito.when(mockAppender.getName()).thenReturn("MockAppender");
+		Mockito.when(mockAppender.isStarted()).thenReturn(true);
+		Mockito.when(mockAppender.isStopped()).thenReturn(false);
+
+		setLogLevel(Level.INFO);
+	}
+
+	@AfterClass
+	public static void afterClass() {
+		System.setProperty("scalingenginejob.refire.interval", String.valueOf(originalJobRefireInterval));
```
fraenkel: Why are we bothering with system properties? Just set the values you want in the before.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleJobTest.java
```diff
@@ -0,0 +1,492 @@
...
+		ArgumentCaptor<URL> urlArgumentCaptor = ArgumentCaptor.forClass(URL.class);
+		ArgumentCaptor<ActiveScheduleEntity> activeScheduleEntityArgumentCaptor = ArgumentCaptor
+				.forClass(ActiveScheduleEntity.class);
+		Mockito.when(scalingEngineUtil.getConnection(urlArgumentCaptor.capture(),
+				activeScheduleEntityArgumentCaptor.capture())).thenReturn(mockHttpURLConnection);
+
+		scheduler.scheduleJob(jobInformation.getJobDetail(), jobInformation.getTrigger());
+
+		//sleep
+		Thread.sleep(TimeUnit.SECONDS.toMillis(2));
```
fraenkel: This isn't going to work well if you rely on timing.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleJobTest.java
```diff
@@ -0,0 +1,492 @@
...
+		String expectedMessage = messageBundleResourceHelper.lookupMessage("scalingengine.notification.failed", 500,
+				"test error message", activeScheduleEntity.getAppId(), activeScheduleEntity.getId(), JobActionEnum.END);
+
+		assertThat("Log level should be ERROR", logCaptor.getValue().getLevel(), is(Level.ERROR));
+		assertThat(logCaptor.getValue().getMessage().getFormattedMessage(), is(expectedMessage));
+
+	}
+
+	@Test
+	public void testNotifyScalingEngine_when_invalidURL() throws Exception {
```
fraenkel: An invalid URL would fail differently.
There is the case where I give you a garbage url, e.g., ftp://wrong.scheme
Or an non-existent endpoint, e.g., http://going.no.where

None of these return a 404.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController_ScalingEngine_IntegrationTest.java
```diff
@@ -0,0 +1,186 @@
+package org.cloudfoundry.autoscaler.scheduler.rest;
```
fraenkel: Integration tests should live elsewhere. We need to run units and know they succeed before we attempt integration tests.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/ActiveScheduleDaoImpl.java
```diff
@@ -0,0 +1,67 @@
...
+
+	private static final String SELECT_SQL = "SELECT * FROM " + TABLE_NAME + " WHERE active_schedule_id=?";
+
+	private static final String INSERT_SQL = "INSERT INTO " + TABLE_NAME
+			+ "(active_schedule_id, app_id, instance_min_count, instance_max_count, initial_min_instance_count, status) "
+			+ "VALUES (?, ?, ?, ?, ?, ?)";
+
+	private static final String DELETE_SQL = "DELETE FROM " + TABLE_NAME + " WHERE active_schedule_id=?";
+
+	@Autowired
```
ghaih: Will change it as mentioned.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
```diff
@@ -0,0 +1,61 @@
...
+
+		// Delete the active schedule
+		deleteActiveSchedule(activeScheduleEntity, context);
+
+		notifyScalingEngine(activeScheduleEntity, jobEnd);
+
+	}
+
+	@Transactional
+	private void deleteActiveSchedule(ActiveScheduleEntity activeScheduleEntity, JobExecutionContext jobExecutionContext) throws JobExecutionException {
```
ghaih: Ok. Understood wrongly, will make the make the change.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
ghaih: The job is getting refired throw new JobExecutionException(dve, true); We added a sleep to give a gap before next refire. Should we get rid of sleep and just refire?

fraenkel: Hammering doesn't help. We don't want to waste threads. We should be able to reschedule this appropriately.

ghaih: Ye we can reschedule it.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleJobTest.java
kanekoh: The same property is used in the job class 'AppScalingScheduleJob' so we applied this approach.

fraenkel:
Please rebase next time.

PR #117, opened 2016-11-02
service broker should assign config file path when start

qibobo:
[#133527149]

### DIFF
--- a/servicebroker/lib/index.js
+++ b/servicebroker/lib/index.js
@@ -1,50 +1,8 @@
-'use strict';
-
-var express = require('express');
-var basicAuth = require('basic-auth');
-var bodyParser = require('body-parser');
-
-var fs = require('fs');
+'use strict'
 var path = require('path');
-var logger = require(path.join(__dirname, './logger/logger.js'));
-var settings = require(path.join(__dirname, './config/setting.js'))((JSON.parse(
-  fs.readFileSync(path.join(__dirname, '../config/settings.json'), 'utf8'))));
-var validateResult = settings.validate();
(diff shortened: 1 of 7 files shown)


servicebroker/lib/index.js
```diff
@@ -1,50 +1,8 @@
...
 var path = require('path');
-var logger = require(path.join(__dirname, './logger/logger.js'));
-var settings = require(path.join(__dirname, './config/setting.js'))((JSON.parse(
-  fs.readFileSync(path.join(__dirname, '../config/settings.json'), 'utf8'))));
-var validateResult = settings.validate();
-if (validateResult.valid === false) {
-  logger.error("Invalid configuration: " + validateResult.message);
-  throw new Error('settings.json is invalid');
+var args = process.argv;
+if (args.length == 4 && args[2] == "-c" && args[3] != "") {
```
fraenkel: Can we flip this so we throw the new Error if the args are wrong.
Then we just have the var server = require ... as the last line which makes it obvious.

servicebroker/test/routes/binding_test.js
```diff
@@ -60,11 +61,14 @@ describe('binding RESTful API', function() {
   };
   var policy = { "policy": "testPolicy" };
   before(function(done) {
-    server = require(path.join(__dirname, '../../lib/index.js'));
+    server = require(path.join(__dirname, '../../lib/server.js'))(configFilePath);
```
fraenkel: We should just require the server outside, and just call the function directly each time.
We don't need the delete below either.

qibobo:
updated.

PR #118, opened 2016-11-03
store raw app instances metrics to database

boyang9527:
[#131839215]

- get raw instance metrics from loggregator
- store raw instance metrics to metrics database

Note:
1. using the raw instance metrics for aggregation will be done in story #131839361
2. will remove previous metrics table and related code after story #131839361 is done

### DIFF
--- a/src/autoscaler/metricscollector/db/metricscollector.db.changelog.yml
+++ b/src/autoscaler/metricscollector/db/metricscollector.db.changelog.yml
@@ -30,4 +30,46 @@ databaseChangeLog:
                   name: value
                   type: varchar
                   constraints:
-                    nullable: true
+                    nullable: false
+  - changeSet:
+      id: 2
+      author: byang
+      changes:
+        - createTable:
+            tableName: appinstancemetrics
+            columns:
+              - column:
+                  name: appid
+                  type: varchar(255)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: instanceindex
+                  type: int
+                  constraints:
+                    nullable: false
+              - column:
+                  name: collectedat
+                  type: bigint
+                  constraints:
+                    nullable: false                 
+              - column:
+                  name: name
+                  type: varchar(100)
+                  constraints:
+                    nullable: false
+              - column:
+                  name: unit
+                  type: varchar(20)
+                  constraints:
+                    nullable: false                    
+              - column:
+                  name: value
+                  type: varchar
+                  constraints:
+                    nullable: false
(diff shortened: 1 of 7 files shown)


src/autoscaler/metricscollector/db/metricscollector.db.changelog.yml
```diff
@@ -30,4 +30,46 @@ databaseChangeLog:
...
                   constraints:
-                    nullable: true
\ No newline at end of file
+                    nullable: false
+  - changeSet:
+      id: 2
+      author: byang
+      changes:
+        - createTable:
+            tableName: appinstancemetrics
```
fraenkel: I believe we are missing indexes based on how we query and delete.
You should do an explain and see how much a timestamp helps. Name would be another.

src/autoscaler/models/metrics.go
```diff
@@ -31,6 +31,35 @@ type InstanceMetric struct {
...
+				Unit:          UnitBytes,
+				Value:         fmt.Sprintf("%d", cm.GetMemoryBytes()),
+				Timestamp:     e.GetTimestamp(),
+			})
+		}
+	}
+	return metrics
+}
+
 func GetMemoryMetricFromContainerMetrics(appId string, containerMetrics []*events.Envelope) *Metric {
```
fraenkel: Why would we still have this?

boyang9527: I keep this so as not to break the current integration between metrics collector and event generator. 

I will remove them after I make the corresponding changes on API and eventgenerator

src/autoscaler/metricscollector/db/metricscollector.db.changelog.yml
boyang9527: I will add index.

timestamp will give the right time metrics are reported, otherwise we can only use the time when we poll metrics from loggregator, which does not reflect the correct time of the metric.  And loggregator caches the metrics, so even the app instance is stopped, we still get a metric value when we poll before it expires in cache. 

'name' is reserved for other metric type.  Do we want to store different type of metrics in different table?  
Making it generic will be better for future extension with more metric types.

boyang9527:
I committed a change to add index, and revised test cases accordingly. Please review

src/autoscaler/metricscollector/db/metricscollector.db.changelog.yml
fraenkel: That is fine for now.

PR #119, opened 2016-11-08
use local active schedule table

boyang9527:
[#119511307]
- create local active schedule table for scaling engine
- merge historydb and activescheduledb to scalingenginedb
- save and remove active schedule when being notified of active schedule start/end
- return 404 when active schedule does not exist in local database


fraenkel:
I fixed your testcase malfunction

PR #120, opened 2016-11-08
api server should get settings from config file

qibobo:
[#133449817]


api/config/settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "port": 8080,
+  "dbUri": "postgres://postgres@127.0.0.1:5432/autoscaler",
```
fraenkel: It would look and feel better if it was 
db {
  uri: xxx,
 maxConnections:
  minConnections:
  idleTimeout:
}

api/lib/routes/policies.js
```diff
@@ -1,81 +1,85 @@
...
-var router = express.Router();
-var logger = require('../log/logger');
-var HttpStatus = require('http-status-codes');
-var validationMiddleWare = require('../validation/validationMiddleware');
-var routeHelper = require('./routeHelper');
-var schedulerUtil = require('../utils/schedulerUtils');
-var async = require('async');
+'use strict';
+module.exports = function(settings){
+  var models = require('../models')(settings);
```
fraenkel: Just pass the db settings once its separate.

qibobo:
updates for separating db settings

PR #122, opened 2016-11-11
133495707: Scheduler should delete the active schedule when schedule is deleted/updated on delete/update of policy

ghaih:


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -710,4 +716,21 @@ private void deleteRecurringSchedule(RecurringScheduleEntity recurringScheduleEn
 		}
 	}
 
+    private void deleteActiveSchedules(String appId) {
+        try {
+            logger.info("Get all active schedules for application: " + appId);
+            List<ActiveScheduleEntity> activeSchedules = activeScheduleDao.findAllActiveSchedulesByAppId(appId);
```
fraenkel: Why aren't we just deleting all active schedules with appId = appId?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManagerTest.java
```diff
@@ -46,7 +50,7 @@
 @ActiveProfiles("ScheduleDaoMock")
 @RunWith(SpringRunner.class)
 @SpringBootTest
-public class ScheduleManagerTest  extends TestConfiguration {
```
fraenkel: Where are the tests for delete active schedules?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
ghaih: Yes could have simply added method to delete by appId , was trying to use existing method. Will change it.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManagerTest.java
ghaih: testDeleteSchedules, has the test for delete active schedules
Mockito.verify(activeScheduleDao, Mockito.times(1)).findAllActiveSchedulesByAppId(appId);
Mockito.verify(activeScheduleDao, Mockito.times(8)).delete(1L);

Which will now change because of above change.

PR #125, opened 2016-11-15
Service-broker does not rollback the binding when api server returns …

qibobo:
…error.

[#134273599]


fraenkel:
Where is the test case? We should have had a test case cover this or the previous behavior.

qibobo:
added the relevant test cases.

PR #126, opened 2016-11-16
use promise-pattern to call commit and rollback

qibobo:


fraenkel:
I did clean this up a bit since it was a bit unclear where things ended. I would have been smart to use an http library that supported promises. The mix of callbacks and promises makes this look far worse than it is.

PR #127, opened 2016-11-17
Integration test between broker and api server

qibobo:
[#133153945]


fraenkel:
This is incomplete.
You have 2 tests that barely cover bind.
Why aren't there tests for provision and deprovision?
Why aren't there tests when the API server is down?
Use the ginkgomon helpers to Interrupt or Kill.
I don't see how this can ever run in parallel given one DB.

src/autoscaler/integration/integration_broker_api_test.go
```diff
@@ -112,102 +113,109 @@ var _ = Describe("Integration_Broker_Api", func() {
...
-				req.Header.Set("Authorization", "Basic "+brokerAuth)
-				resp, err := httpClient.Do(req)
+			It("should return 400, save no scaling rule to db and not call scheduler", func() {
+
+				resp, err := bindService(bindingId, appId, serviceInstanceId, invalidPolicyJson)
 				Expect(err).NotTo(HaveOccurred())
-				Expect(resp.StatusCode).To(Equal(201))
-				Eventually(getNumberOfBinding).Should(Equal(1))
-				Eventually(getNumberOfPolicyJson).Should(Equal(1))
+				Expect(resp.StatusCode).To(Equal(400))
```
fraenkel: defer resp.Body.Close()

src/autoscaler/integration/integration_broker_api_test.go
```diff
@@ -112,102 +113,109 @@ var _ = Describe("Integration_Broker_Api", func() {
...
+				}).Should(Equal(0))
 				Consistently(scheduler.ReceivedRequests).Should(HaveLen(0))
 			})
 		})
-		Context("Invalid policy", func() {
+		Context("Api-server is down", func() {
 			BeforeEach(func() {
-				scheduler.RouteToHandler("PUT", regPath, ghttp.RespondWith(http.StatusOK, "successful"))
+				stopApiServer()
+				scheduler.RouteToHandler("PUT", regPath, ghttp.RespondWith(http.StatusInternalServerError, "error"))
```
fraenkel: You need to make sure that the API server is actually down before leaving the BeforeEach.

I also noticed that the service broker => API has no timeouts on those requests.

src/autoscaler/integration/integration_suite_test.go
```diff
@@ -65,6 +64,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	if e != nil {
 		Fail("can not connect database: " + e.Error())
 	}
+	clearDatabase()
```
fraenkel: This should be done above. This is done on every parallel instance rather than on instance 0. We should only clear once.

src/autoscaler/integration/integration_suite_test.go
```diff
@@ -65,6 +64,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 })
 
 var _ = AfterSuite(func() {
```
fraenkel: You should be using the SynchronizedAfterSuite to match the above.

src/autoscaler/integration/integration_suite_test.go
```diff
@@ -75,19 +75,19 @@ var _ = AfterSuite(func() {
 })
 
 var _ = BeforeEach(func() {
+	httpClient = &http.Client{}
```
fraenkel: If you are going to create a client, please set the appropriate timeouts. You will need to lift the DefaultTransport definition and just modify/add timeouts as needed.

src/autoscaler/integration/integration_suite_test.go
```diff
@@ -114,7 +114,27 @@ func PreparePorts() integration.Ports {
...
+}
+func stopApiServer() {
+	ginkgomon.Kill(processMap["apiServer"], 5*time.Second)
+}
+func stopAll() {
+	for _, process := range processMap {
+		if process == nil {
+			continue
+		}
+		ginkgomon.Kill(process, 5*time.Second)
```
fraenkel: Normally we use Interrupt so it does a proper shutdown versus a hard stop.

src/autoscaler/integration/integration_broker_api_test.go
qibobo: In ginkomon.Interruput the below code exist:
Eventually(process.Wait(), intervals...).Should(Receive(), "interrupted ginkgomon process failed to exit in time")
Does it ensure that the process is actually down?

fraenkel: That is what Wait() means, wait for the process to exit.

qibobo: So how to ensure that the api server is actually down?

fraenkel: Send a request and expect a connection error.

PR #129, opened 2016-11-17
Synchronize active schedules  from schedler database

boyang9527:
[#131776923]

- add schedlerDB, and remove "status" column
- add configuration for schedulerdb url and synchronization time interval
- add synchronizer to sync up active schedules with scheduler
- move lock from scaling_handler to scalingengine
- add synchronizer to main


src/autoscaler/scalingengine/config/config_test.go
```diff
@@ -102,6 +109,7 @@ cf:
 db:
   policy_db_url: test-policy-db-url
   scalingengine_db_url: test-scalingengine-db-url
+  SchedulerDbUrl: test-scheduler-db-url
```
fraenkel: Upper case?!?
This is working because Go does case-insensitive matching for you.

src/autoscaler/scalingengine/schedule/sync.go
```diff
@@ -0,0 +1,89 @@
...
+			wg.Add(1)
+			go func(aid string, as *models.ActiveSchedule) {
+				defer wg.Done()
+				ss.engine.SetActiveSchedule(aid, as)
+			}(appId, schedule)
+		}
+	}
+
+	for appId, scheduleId := range asEngine {
+		if asScheduler[appId] == nil {
```
fraenkel: This could be optimized to do bulk deletes, but we will go with one at a time for now.

src/autoscaler/scalingengine/schedule/sync.go
```diff
@@ -0,0 +1,89 @@
...
+
+	asEngine, err := ss.engineDB.GetActiveSchedules()
+	if err != nil {
+		ss.logger.Error("failed-synchronize-active-schedules-get-schedules-from-engineDB", err)
+		return
+	}
+
+	wg := &sync.WaitGroup{}
+
+	for appId, schedule := range asScheduler {
```
fraenkel: We may want to add logging, so we know how frequently we are "fixing" our active schedules.

src/autoscaler/scalingengine/config/config_test.go
boyang9527: sorry it is a typo, should be "scheduler_db_url". This test case is to test default values, so it passed even the name is not correct.

src/autoscaler/scalingengine/schedule/sync.go
boyang9527: added.

src/autoscaler/scalingengine/schedule/sync.go
boyang9527: sure

boyang9527:
@fraenkel  pushed another commit to fix the typo and add info logs when synchronizing active schedules

PR #134, opened 2016-11-22
add TLS support for metrics collector and scaling engine endpoint

boyang9527:
[#121340317]

- add SSL config for metrics collector and scaling engine
- add SSL support for http servers
- test https in main


src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
```diff
@@ -227,3 +265,75 @@ func marshalMessage(message *events.Envelope) []byte {
 
 	return data
 }
+
+var testCaCert = []byte(`
```
fraenkel: Why aren't these just test assests as files?

src/autoscaler/metricscollector/config/config.go
```diff
@@ -23,11 +24,13 @@ var defaultCfConfig = cf.CfConfig{
 }
 
 type ServerConfig struct {
-	Port int `yaml:"port"`
+	Port      int  `yaml:"port"`
+	EnableSSL bool `yaml:"enable_ssl"`
```
fraenkel: Why isn't the SSL parts just here as an option? If present, we do SSL, if not, we don't.

boyang9527: removed EnableSSL setting

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
boyang9527: moved the certificates/key  to files

boyang9527:
@fraenkel  changes submitted, pls review.

fraenkel:
Merged

PR #137, opened 2016-11-26
Integration test between api server and scheduler

qibobo:
[#133153945]
1 integration test for api and scheduler
2 read fake policy from file

### DIFF
--- a/src/integration/integration_api_scheduler_test.go
+++ b/src/integration/integration_api_scheduler_test.go
@@ -29,10 +29,10 @@ var _ = Describe("Integration_Api_Scheduler", func() {
 		resp, err := detachPolicy(appId)
 		resp.Body.Close()
 		Expect(err).NotTo(HaveOccurred())
-		Expect(resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound).Should(Equal(true))
 	})
 	AfterEach(func() {
 		fakeScalingEngine.Close()
+		stopAll()
 	})
 	Describe("Create policy", func() {
 		Context("Policies with schedules", func() {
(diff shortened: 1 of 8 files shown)


src/integration/integration_api_scheduler_test.go
```diff
@@ -0,0 +1,240 @@
...
+		})
+		Context("with an existing app", func() {
+			BeforeEach(func() {
+				//attach a policy first with 4 recurring and 2 specific_date schedules
+				policyStr = readPolicyFromFile("fakePolicyWithSchedule.json")
+				resp, err := attachPolicy(appId, policyStr)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(resp.StatusCode).To(Equal(http.StatusCreated))
+				resp.Body.Close()
+				By("checking the API Server")
```
fraenkel: Same as above.

src/integration/integration_suite_test.go
```diff
@@ -36,16 +37,19 @@ var (
...
-	dbHelper           *sql.DB
-	scheduler          *ghttp.Server
-	httpClient         *http.Client
-	httpRequestTimeout time.Duration            = 1000 * time.Millisecond
-	processMap         map[string]ifrit.Process = map[string]ifrit.Process{}
+	dbHelper                       *sql.DB
+	fakeScheduler                  *ghttp.Server
+	fakeScalingEngine              *ghttp.Server
+	httpClient                     *http.Client
+	brokerApiHttpRequestTimeout    time.Duration            = 1000 * time.Millisecond
```
fraenkel: why not just say 1 * time.Second and 5 * time.Second?

src/integration/integration_suite_test.go
```diff
@@ -93,9 +92,7 @@ var _ = SynchronizedAfterSuite(func() {
 })
 
 var _ = BeforeEach(func() {
-	httpClient = cfhttp.NewCustomTimeoutClient(httpRequestTimeout)
-	startApiServer()
-	startServiceBroker()
```
fraenkel: Since you moved the starts into the respective tests, I would move the stopAll into each one as well. It is very confusing to have it stop out here but start inside.

PR #138, opened 2016-11-28
Retry container envelopes in metrics collector

pradyutsarma:
1. metricscollector retries 3 times in case of failure while fetching container envelopes
2. Requisite tests

Will have to refactor the code and do a more selective retry based on the noaa responses . Depends on the issue here https://github.com/cloudfoundry/noaa/issues/27

### DIFF
--- a/src/autoscaler/metricscollector/collector/apppoller.go
+++ b/src/autoscaler/metricscollector/collector/apppoller.go
@@ -9,6 +9,8 @@ import (
 	"code.cloudfoundry.org/clock"
 	"code.cloudfoundry.org/lager"
 
+	"github.com/cloudfoundry/sonde-go/events"
+
 	"time"
 )
 
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
 func (ap *appPoller) pollMetric() {
 	ap.logger.Debug("poll-metric", lager.Data{"appid": ap.appId})
 
-	containerEnvelopes, err := ap.noaaConsumer.ContainerEnvelopes(ap.appId, "bearer"+" "+ap.cfc.GetTokens().AccessToken)
-	if err != nil {
-		ap.logger.Error("poll-metric-from-noaa", err)
-		return
+	var containerEnvelopes []*events.Envelope
+	var err error
+
+	for attempt := 1; attempt <= 3; attempt++ {
+		if attempt != 0 {
+			ap.logger.Debug("poll-metric-from-noaa-retry", lager.Data{"attempt": attempt})
+		}
+
+		containerEnvelopes, err = ap.noaaConsumer.ContainerEnvelopes(ap.appId, "bearer"+" "+ap.cfc.GetTokens().AccessToken)
+
+		if err != nil {
+			ap.logger.Error("poll-metric-from-noaa", err)
+		} else {
+			break
+		}
 	}
 
 	metrics := models.GetInstanceMemoryMetricFromContainerEnvelopes(ap.pclock.Now().UnixNano(), ap.appId, containerEnvelopes)
--- a/src/autoscaler/metricscollector/collector/apppoller.go
+++ b/src/autoscaler/metricscollector/collector/apppoller.go
@@ -77,18 +77,18 @@ func (ap *appPoller) pollMetric() {
 	var containerEnvelopes []*events.Envelope
 	var err error
 
-	for attempt := 1; attempt <= 3; attempt++ {
-		if attempt != 0 {
-			ap.logger.Debug("poll-metric-from-noaa-retry", lager.Data{"attempt": attempt})
-		}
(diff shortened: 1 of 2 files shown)


src/autoscaler/metricscollector/collector/apppoller.go
```diff
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
...
 	ap.logger.Debug("poll-metric", lager.Data{"appid": ap.appId})
 
-	containerEnvelopes, err := ap.noaaConsumer.ContainerEnvelopes(ap.appId, "bearer"+" "+ap.cfc.GetTokens().AccessToken)
-	if err != nil {
-		ap.logger.Error("poll-metric-from-noaa", err)
-		return
+	var containerEnvelopes []*events.Envelope
+	var err error
+
+	for attempt := 1; attempt <= 3; attempt++ {
```
fraenkel: It's better like  for attempt := 0; attempt < 3; attempt++ {

src/autoscaler/metricscollector/collector/apppoller.go
```diff
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
...
+		if attempt != 0 {
```
fraenkel: Remove the guard, I would want to see it for all attempts. Just remove the debug statement above.

src/autoscaler/metricscollector/collector/apppoller.go
```diff
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
...
+			ap.logger.Debug("poll-metric-from-noaa-retry", lager.Data{"attempt": attempt})
```
fraenkel: Not enough information. You need to include the appId.

src/autoscaler/metricscollector/collector/apppoller.go
```diff
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
...
+		}
+
+		containerEnvelopes, err = ap.noaaConsumer.ContainerEnvelopes(ap.appId, "bearer"+" "+ap.cfc.GetTokens().AccessToken)
+
+		if err != nil {
```
fraenkel: It's better to do 
```
if err == nil {
  break
}
```

src/autoscaler/metricscollector/collector/apppoller.go
```diff
@@ -72,10 +74,21 @@ func (ap *appPoller) startPollMetrics() {
...
+			ap.logger.Error("poll-metric-from-noaa", err)
```
fraenkel: Need more data.
It would be better to just do a 
logger := ap.logger.WithData({"appId":appId}) above and then all the log statements have the most important information.

src/autoscaler/metricscollector/collector/apppoller_test.go
```diff
@@ -211,7 +211,7 @@ var _ = Describe("Apppoller", func() {
 					Expect(appid).To(Equal("test-app-id"))
 					Expect(token).To(Equal("bearer test-access-token"))
 
-					if noaa.ContainerEnvelopesCallCount()%2 == 0 {
+					if noaa.ContainerEnvelopesCallCount() >= 2 && noaa.ContainerEnvelopesCallCount() <= 4 {
```
fraenkel: Why a range?

src/autoscaler/metricscollector/collector/apppoller_test.go
```diff
@@ -256,6 +256,46 @@ var _ = Describe("Apppoller", func() {
 			})
 		})
 
+		Context("when retrieving container envelopes fail, the metrics collector retries", func() {
```
fraenkel: All you needed to do was check that the call count of ContainerEnvelopes is 3.

src/autoscaler/metricscollector/collector/apppoller_test.go
pradyutsarma: The range is required because the idea was to simulate "a poll success and the corresponding three poll failures and thereafter success again". Basically, we pass the first time (when call count is 1) and then fail 3 times (between 2 to 4) and allow a pass again after that.

fraenkel: How is this different than the fail and retry?

I would also recommend that we do not rely on logging for testing outcomes. It would be far better to use the fact that we call models.GetInstanceMemoryMetricFromContainerEnvelopes as a means to verify.

pradyutsarma: @fraenkel Are you then suggesting that we remove this test that we have a fail and retry test below already?

fraenkel: Yes, unless there is something this is testing that is different.

PR #141, opened 2016-11-29
Scheduler logs converted to JSON format

paltanmoy:
We've modified the LOG_PATTERN to have the scheduler logs in JSON format with predefined attributes. 
It doesn’t have the context object like the apiserver or the broker. We could use the Mapped Diagnostic Context (MDC) http://logback.qos.ch/manual/mdc.html to add that, but that would add some amount of complication into the scheduler logging like for example, creating JSONObjects within the java code and attach them to the logs.
Just wanted to know what you think?

### DIFF
--- a/scheduler/src/main/resources/log4j2.xml
+++ b/scheduler/src/main/resources/log4j2.xml
@@ -2,7 +2,7 @@
 <!--<!DOCTYPE xml>-->
 <Configuration status="WARN" monitorInterval="30">
     <Properties>
-        <Property name="LOG_PATTERN">%d{yyyy-MM-dd HH:mm:ss} [%thread] %-5level %logger{36} - %msg%n</Property>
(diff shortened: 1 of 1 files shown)


paltanmoy:
An example of an error log after this change is below - 
`{"timestamp":"2016-11-28T13:41:11,975","source":"o.c.a.s.q.AppScalingScheduleStartJob:81","log_level":"ERROR","message":"Error connecting to scaling engine, failed with error: I/O error on PUT request for \"http://127.0.0.1/v1/apps/122131313/active_schedules/789\": Software caused connection abort: recv failed; nested exception is java.net.SocketException: Software caused connection abort: recv failed for app id: 122131313 and schedule id: 789 to START active schedule."}`

fraenkel:
I plan to close this without merging. We are not hacking some JSON looking thing and pretending it will work.

 This doesn't cover any of the more common string issues.

As I have stated in the past, we will punt on this part for now.

PR #144, opened 2016-11-30
Refactor apppoller to use a timer instead of ticker

pradyutsarma:
1. Use timer instead of ticker
2. Add a 'missed' return to the poll method (missed in last retry PR)


fraenkel:
I made some additional changes in a separate commit.

PR #145, opened 2016-11-30
Fix the aggregator unit-test failure when aggregator starts MetricPoller

qibobo:
[#132791853]
1 Make sure that setAppMonitors is executed before
addToAggregateChannel the first time they are executed.

### DIFF
--- a/src/autoscaler/eventgenerator/aggregator/aggregator.go
+++ b/src/autoscaler/eventgenerator/aggregator/aggregator.go
@@ -24,6 +24,8 @@ type Aggregator struct {
 	aggregatorExecuteInterval time.Duration
 	appMonitorArray           []*model.AppMonitor
 	lock                      sync.Mutex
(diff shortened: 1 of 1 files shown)


fraenkel:
This is a hack. I will not be merging this.
Please explain what the actual issue is.

fraenkel:
This took me a while to figure out the real issue.
Part of the problem is how Aggregator is structured which I will get to in a minute.

The issue is that the Policy needs to be retrieved and pushed out before we start testing the pollers. The way to detect this is by seeing SetTriggers get invoked. This needs to be an interface so we can inject a fake rather than the real thing. Once we do that, you just wait until we see it invoked and then tick off.

fraenkel:
I just created this https://www.pivotaltracker.com/story/show/135286961 which affects all the other failing tests I have seen.

fraenkel:
This is no longer valid or needed.

Issue #148, opened 2016-12-07
AutoScaling app failed to start saying 'org.ektorp.DbAccessException: Connection refused'


jessehu:
[Note: the original subject and description was removed]

I'm deploying app-autoscaler to my latest Open Source Cloud Foundry. AutoScalingAPI started but AutoScaling and AutoScalingServiceBroker failed to start.

The CF app logs says "org.cloudfoundry.autoscaler.data.couchdb.connection.manager.CouchDbConnectionManager:27 [localhost-startStop-1] [<init>] java.net.ConnectException: Connection refused (Connection refused)". However, I can connect to the CouchDB on 192.168.101.208 using a ruby couchdb client script.  
Verified that the deployed AutoScaling App reads 192.168.101.208 as the CouchDB host (by changed 192.168.101.208 to a192.168.101.208 which throws error '... a192.168.101.208 ...')
All CF VMs are in subnet 192.168.101.0.

Gurus, please give some advise. Thanks in advance.

------- commands and config file --------
mvn clean package -Denv=myenv -DskipTests
bin/deploy.sh myenv

$ cat profiles/myenv.properties 
```
couchdbUsername=autoscaler
couchdbPassword=password
couchdbHost=192.168.101.208
couchdbPort=5984
couchdbServerDBName=couchdb-scaling
couchdbMetricDBPrefix=couchdb-scalingmetric
couchdbBrokerDBName=couchdb-scalingbroker
```

$ cf logs AutoScaling 
```
Connected, tailing logs for app AutoScaling in org xxx / space xxx as xxx ...
2016-12-07T05:58:44.31-0500 [API/0]      OUT Updated app with guid 37aeb982-ca36-4f18-93ac-862abd2480da ({"state"=>"STOPPED"})
2016-12-07T05:58:44.80-0500 [DEA/0]      OUT Got staging request for app with id 37aeb982-ca36-4f18-93ac-862abd2480da
2016-12-07T05:58:46.54-0500 [API/0]      OUT Updated app with guid 37aeb982-ca36-4f18-93ac-862abd2480da ({"state"=>"STARTED"})
2016-12-07T05:58:46.75-0500 [STG/0]      OUT -----> Downloaded app package (5.6M)
2016-12-07T05:58:46.82-0500 [STG/0]      OUT -----> Downloaded app buildpack cache (4.0K)
[...]

PR #149, opened 2016-12-07
Correct the couchdb name property name in README.md

jessehu:
Fix issue #148 
- Correct the couchdb name property name in README.md
- Add myenv.properties


jessehu:
I made a silly mistake...  Please ignore this.

PR #151, opened 2016-12-08
134273037: Handle multiple active schedules.

kanekoh:
- Added start_job_identifier to connect the start and the end job.
- In recurring schedule end job is not cron job any more. 
- In both specific date and recurring schedule end job is created as a simple job when start job starts.


fraenkel:
Glad we got this all working.

PR #152, opened 2016-12-13
update metric collector url in event generator

qibobo:


fraenkel:
This seems very error prone. Can you see what it would take to expose the Path(s) and use the mux.Router to build the URLs rather than all the hard coded strings we have?

qibobo:
add the route package

src/autoscaler/routes/routes.go
```diff
@@ -0,0 +1,62 @@
...
+	instance.scalingEngineRoutes.Path(scalingHistoriesPath).Name(HistoreisRoute)
+	instance.scalingEngineRoutes.Path(activeSchedulePath).Name(UpdateActiveSchedulesRoute)
+	instance.scalingEngineRoutes.Path(activeSchedulePath).Name(DeleteActiveSchedulesRoute)
+
+	return instance
+
+}
+
+func getInstance() *AutoScalerRoute {
+	once.Do(func() {
```
fraenkel: Why bother, this can be done via func init()

src/autoscaler/eventgenerator/aggregator/metricPoller.go
```diff
@@ -65,7 +66,8 @@ func (m *MetricPoller) retrieveMetric(app *model.AppMonitor) {
 	}
 
 	var url string
-	url = m.metricCollectorUrl + "/v1/apps/" + app.AppId + "/metrics_history/memory?start=" + strconv.FormatInt(startTime.UnixNano(), 10) + "&end=" + strconv.FormatInt(endTime.UnixNano(), 10)
+	path, _ := routes.MetricsCollectorRoutes().Get(routes.MemoryMetricHistoryRoute).URLPath("appid", app.AppId)
+	url = m.metricCollectorUrl + path.Path + "?start=" + strconv.FormatInt(startTime.UnixNano(), 10) + "&end=" + strconv.FormatInt(endTime.UnixNano(), 10)
```
fraenkel: This should be
v:= path.Query()
v.Add("start", strconv.....)

qibobo:
simplify routes module

qibobo:
fix conflicts

PR #153, opened 2016-12-15
135923225: Fixed timezone problem in Scheduler

kanekoh:


fraenkel:
This just doesn't look right.
There is no reason why we need to parse a formatted date string to do time zone conversions. Find another way that is more direct.

kanekoh:
Yes, we had looked at Java8 options before but stuck to Java7 options this was the only way.
Now we will change as per Java 8 and use java.time package.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DateHelper.java
```diff
@@ -603,17 +602,9 @@
...
-
-		SimpleDateFormat sdf = new SimpleDateFormat(DATE_TIME_FORMAT);
-		sdf.setTimeZone(timeZone);
-
-		try {
-			return sdf.parse(localSdf.format(dateTime));
-		} catch (ParseException e) {
-			throw new SchedulerInternalException("Invalid date specified. date=" + dateTime, e);
-		}
+		LocalDateTime local = LocalDateTime.ofInstant(dateTime.toInstant(), ZoneId.systemDefault());
```
fraenkel: Why do we jump through all these hoops?
Why not just use ZonedDateTime.ofInstant(instant, zone)?

kanekoh: Regarding Time zone issue, we had tried ZoneDateTime and other possibilities using java.time package but have not been able to get the expected date time.
The approach we used LocalDateTime  is also failing during DST the whole issue is with date class which implicitly converts the JSON date string to GMT date within the entity.

We are thinking it would be the best to save date as a string in database which has no zone attached. 

What are your thought?

fraenkel: My point above was to simplify the code. We don't have to jump through so many hoops when ZonedDateTime would do it for us in fewer steps.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
```diff
@@ -67,22 +71,32 @@ public static boolean isValidTimeZone(String timeZoneId) {
...
+		LocalDate policyLocalDate = getPolicyZonedDateTime(date, policyZoneId).toLocalDate();
+		LocalDate localDateNow = LocalDate.now(policyZoneId);
+
+		return !policyLocalDate.isBefore(localDateNow);
+
 	}
 
-	public static boolean isDateAfterOrEqualsNow(Date date, TimeZone policyTimeZone) {
-		Date compareTo = DateHelper.getDateWithZoneOffset(date, policyTimeZone);
+	private static ZonedDateTime getPolicyZonedDateTime(Date dateTime, ZoneId policyZone) {
```
fraenkel: Why is this here? Shouldn't we be using the one in DateHelper?

fraenkel:
Where is the test to guarantee we don't fall back into the same trap and that the new code actually does what we think?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/util/DataValidationHelper.java
kanekoh: Yes, that was a mistake.

PR #158, opened 2016-12-30
Add TLS support for service broker.

qibobo:
[#136717867]


servicebroker/lib/config/setting.js
```diff
@@ -83,6 +84,21 @@ module.exports = function(settingsObj) {
     if (settings.httpRequestTimeout < 0) {
       return { valid: false, message: "The value of httpRequestTimeout must be greater than 0" };
     }
+    if (!settings.tls) {
+      return { valid: false, message: "tls is required" };
+    }else if(typeof(settings.tls) != "object"){
```
fraenkel: Style wise I wouldn't have had an else block. You already returned from the if.

servicebroker/lib/server.js
```diff
@@ -19,6 +20,17 @@ module.exports = function(configFilePath) {
         throw new Error('settings.json is invalid');
     }
     var port = settings.port;
+    var options = {};
+    
+    if(!fs.existsSync(settings.tls.keyFile) || !fs.existsSync(settings.tls.certFile) || !fs.existsSync(settings.tls.caCertFile)){
```
fraenkel: be explicit. Check each one and give the appropriate error message.

qibobo:
update the code and divide the commits.

PR #161, opened 2017-01-05
Add TLS support for scheduler.

qibobo:


fraenkel:
Where are the changes for the scheduler itself?

qibobo:
Add TLS support for scheduler

scheduler/pom.xml
```diff
@@ -91,6 +91,12 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-actuator</artifactId>
 		</dependency>
+		<!-- https://mvnrepository.com/artifact/org.apache.httpcomponents/httpclient -->
+		<dependency>
+		    <groupId>org.apache.httpcomponents</groupId>
```
fraenkel: why?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/ssl/CustomClientHttpRequestFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.apache.http.client.HttpClient;
+import org.apache.http.conn.ssl.NoopHostnameVerifier;
+import org.apache.http.conn.ssl.SSLConnectionSocketFactory;
+import org.apache.http.conn.ssl.TrustSelfSignedStrategy;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.ssl.SSLContextBuilder;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+
+public class CustomClientHttpRequestFactory extends HttpComponentsClientHttpRequestFactory {
```
fraenkel: why?
We should just support HTTPS which means we just configure application.properties and that is it.

scheduler/src/main/resources/application.properties
```diff
@@ -12,3 +12,11 @@ scalingengine.notification.reschedule.maxcount=3
...
 # scaling engine url
 autoscaler.scalingengine.url=http://test-scalingengine-test
+
+#ssl
+server.ssl.key-store=../test-certs/scheduler.jks
+caCert=../test-certs/autoscaler-ca.crt
+server.ssl.key-alias=scheduler
+server.ssl.key-store-password=123456
+#server.ssl.key-password=123456
+server.ssl.key-store-type=JKS
```
fraenkel: why not just use pkcs12? makes it easier for everyone. which is what you generated from the tooling.

qibobo:
remove useless code.

PR #163, opened 2017-01-09
Added ability to shutdown nodejs servers gracefully 

paltanmoy:
1) Gracefully shutting down APIServer
2) Gracefully shutting down ServiceBroker

### DIFF
--- a/api/index.js
+++ b/api/index.js
@@ -1,7 +1,31 @@
 'use strict'
 var path = require('path');
+var logger = require('./lib/log/logger');
 var args = process.argv;
 if (!(args.length == 4 && args[2] == "-c" && args[3] != "")) {
     throw new Error("missing config file\nUsage:use '-c' option to specify the config file path");
 }
 var apiServer = require(path.join(__dirname, 'app.js'))(args[3]);
+
+var gracefulShutdown = function(signal) {
+  logger.info("Received " + signal + " signal, shutting down gracefully...");
+  apiServer.shutdown(function() {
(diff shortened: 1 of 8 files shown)


api/index.js
```diff
@@ -1,7 +1,31 @@
...
+
+var gracefulShutdown = function(signal) {
+  logger.info("Received " + signal + " signal, shutting down gracefully...");
+  apiServer.shutdown(function() {
+    logger.info('Everything is cleanly shutdown');
+  })
+}
+
+//listen for SIGTERM signal e.g. kill
+process.on ('SIGTERM', function(){
```
fraenkel: We shouldn't be doing a graceful shutdown on SIGTERM. That should be the escape valve for kill immediate.

servicebroker/lib/index.js
```diff
@@ -1,7 +1,30 @@
...
+
+var gracefulShutdown = function(signal) {
+  logger.info("Received " + signal + " signal, shutting down gracefully...");
+  server.shutdown(function() {
+    logger.info('Everything is cleanly shutdown');
+  })
+}
+
+//listen for SIGTERM signal e.g. kill
+process.on ('SIGTERM', function(){
```
fraenkel: See above

api/index.js
paltanmoy: SIGTERM removed

fraenkel:
Normally there should be a test added especially for the keep-alive, but we can go without it for now.

PR #165, opened 2017-01-24
137513231 : Integration test failure: api-scheduler

rohitsharma04:
1. Force shutdown in case graceful shutdown timeouts.
2. Updated Readme.


fraenkel:
You are missing tests.

rohitsharma04:
Added tests.

api/app.js
```diff
@@ -63,7 +63,33 @@ module.exports = function(configFilePath) {
...
+      logger.info('Everything is cleanly shutdown');
+      process.exit();
+    })
+    
+    setTimeout(function(){
+      server.forceShutdown(function(){
+        logger.info('Could not close connections in time, forcefully shutting down');
+        process.exit();
+      })
+    }, 10*1000);
```
fraenkel: Make a separate PR to clean this up with a constant.
Realize that when this is integrated, drain will issue SIGUSR2 until some time period and invoke a SIGKILL.

fraenkel:
I should have been more clear. An integration test would be far better than a unit test.
Start the server, have a client connect and disconnect, send a SIGUSR2 and watch the process die.
Do it again with a client that connects but doesn't disconnect and send a SIGUSR2, the process shouldn't die. Kill the process with a SIGKILL and it should die.
We really don't need the timeout. If our tests are hanging, we either do one of two things, send a SIGKILL always so we don't worry about this issue or we fix the issue. If we expect to have long running clients, then we go the SIGKILL route, if we don't then we are missing a proper close on the client side.

fraenkel:
@rohitsharma04 @pradyutsarma Is this going to be fixed? This is still breaking our pipeline.

pradyutsarma:
@fraenkel We fixed this by opening and closing tcp connections on the apiserver. Were trying to see if there is any other "better" way of achieving this. Will update the PR tonite.

fraenkel:
@pradyutsarma I don't mind if we keep connections open. It will help once we switch to SSL. What is important is that we test the shutdown, and that our BOSH drain and BOSH stop send the right signals. This also means that when testing we should use the appropriate signals.  If we don't care about graceful shutdown during tests, just kill the server rather than trigger a quiesce.

boyang9527:
After I run API sever - scheduler integration test,  I did not see any opening TCP connections to API sever, but there were remaining TCP connections from API sever to postgres.  In four test cases, the API servers eventually exited after some time. But the other two test cases,  the node processes never exited.

rohitsharma04:
Added integration tests.

src/integration/integration_suite_test.go
```diff
@@ -200,6 +201,18 @@ func stopApiServer() {
 	ginkgomon.Interrupt(processMap[APIServer], 5*time.Second)
 }
 
+func sendSigusr2Signal(component string) {
+	process := processMap[component]
+	if process != nil {
+		process.Signal(syscall.SIGUSR2)
+		Eventually(process.Wait(), 5*time.Second).Should(Receive(), "sent SIGUSR2 signal to process, failed to exit in time")
```
fraenkel: Just send the signal, don't wait since it is test dependent.
The check on the process should occur in the tests. It is something I was going to mention since it is confusing for it to be hidden here.

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+	Describe("Graceful Shutdown", func() {
+
+		Context("ApiServer", func() {
+
+			BeforeEach(func() {
+				startApiServer()
+			})
+
+			It("stops receiving new connections after being interrupted", func() {
```
fraenkel: Isn't this the case where the server should stop since there is no outstanding connections?

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -0,0 +1,177 @@
...
+				_, err = conn.Write([]byte(MessageForServer))
+				Expect(err).NotTo(HaveOccurred())
+
+				newConn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", components.Ports[APIServer]))
+				Expect(err).NotTo(HaveOccurred())
+
+				conn.Close()
+				newConn.Close()
+
+				sendSigusr2Signal(APIServer)
```
fraenkel: I would expect afterwards that the APIServer has exited.

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -0,0 +1,177 @@
...
+				_, err = net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", components.Ports[APIServer]))
+				Expect(err).To(HaveOccurred())
+
+			})
+
+			It("waits for in-flight request to finish before shutting down", func() {
+				conn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", components.Ports[APIServer]))
+				Expect(err).NotTo(HaveOccurred())
+
+				sendSigusr2Signal(APIServer)
```
fraenkel: The APIServer should not have exited.

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+				_, err = net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", components.Ports[APIServer]))
+				Expect(err).To(HaveOccurred())
+
+				_, err = conn.Write([]byte(MessageForServer))
+				Expect(err).NotTo(HaveOccurred())
+
+				conn.Close()
```
fraenkel: The APIServer should have exited.

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -0,0 +1,177 @@
...
+
+		})
+
+		Context("Service Broker", func() {
+
+			BeforeEach(func() {
+				startServiceBroker()
+			})
+
+			It("stops receiving new connections after being interrupted", func() {
```
fraenkel: Same as above.

src/integration/integration_api_broker_graceful_shutdown_test.go
rohitsharma04: Are you suggesting that we also check that the apiserver has gone down; as of now we're only checking that it's not able to receive any new connections ?

fraenkel: Yes. Isn't that what we expect?
You have hidden some of that check in the sendsignal function which is wrong.
In theory, the sigusr2 should stop accepting connections and when the last connection is closed have it shut down.
So it should be open a persistent connection, sendsigusr2, no new connections allowed, process is running, close connection, process has exited.

src/integration/integration_api_broker_graceful_shutdown_test.go
```diff
@@ -52,17 +62,24 @@ var _ = Describe("Integration_Api_Broker_Graceful_Shutdown", func() {
...
 
-			It("waits for in-flight request to finish before shutting down", func() {
+			It("waits for in-flight request to finish before shutting down server on receiving SIGUSR2 signal", func() {
 				conn, err := net.Dial("tcp", fmt.Sprintf("127.0.0.1:%d", components.Ports[APIServer]))
 				Expect(err).NotTo(HaveOccurred())
 
 				sendSigusr2Signal(APIServer)
+				Eventually(buffer, 5*time.Second).Should(gbytes.Say(`Received SIGUSR2 signal`))
+
+				Expect(runner.ExitCode()).Should(Equal(-1))
```
fraenkel: Something isn't right here.
I added the following 
				Eventually(runner.ExitCode).Should(BeZero())
and it did exit which is not expected.

fraenkel:
Added the one comment that applies to both which seems to show that we are exiting with an open connection.

fraenkel:
I have run these tests a few times and have hit a few different issues.

```
[d][serviceBroker] spawned /usr/local/bin/node (pid: 24921)
[o][serviceBroker] {"timestamp":"2017-02-04T13:28:31.747Z","source":"autoscaler:servicebroker","text":"Service broker app is running","log_level":"info","data":{"port":11002}}
[o][serviceBroker] {"timestamp":"2017-02-04T13:28:31.782Z","source":"autoscaler:servicebroker","text":"Received SIGUSR2 signal, shutting down gracefully...","log_level":"info"}
[o][serviceBroker] {"timestamp":"2017-02-04T13:28:31.790Z","source":"autoscaler:servicebroker","text":"Everything is cleanly shutdown","log_level":"info"}

• Failure [1.099 seconds]
Integration_Api_Broker_Graceful_Shutdown
/Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:206
  Graceful Shutdown
  /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:152
    Service Broker
    /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:150
      waits for in-flight request to finish before shutting down server on receiving SIGUSR2 signal [It]
      /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:148

      Expected error:
          <*net.OpError | 0xc42007a0f0>: {
              Op: "write",
              Net: "tcp",
              Source: {IP: [127, 0, 0, 1], Port: 50729, Zone: ""},
              Addr: {IP: [127, 0, 0, 1], Port: 11002, Zone: ""},
              Err: {Syscall: "write", Err: 0x20},
          }
          write tcp 127.0.0.1:50729->127.0.0.1:11002: write: broken pipe
      not to have occurred

      /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:142
```

```
[d][serviceBroker] spawned /usr/local/bin/node (pid: 25271)
[o][serviceBroker] {"timestamp":"2017-02-04T13:31:12.108Z","source":"autoscaler:servicebroker","text":"Service broker app is running","log_level":"info","data":{"port":11006}}


• Failure [0.537 seconds]
Integration_Api_Broker_Graceful_Shutdown
/Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:206
  Non Graceful Shutdown
  /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:204
    Service Broker
    /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:202
      kills server immediately without serving the in-flight requests [It]
      /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:201

      Expected an error to have occurred.  Got:
          <nil>: nil

      /Users/fraenkel/workspace/app-autoscaler/src/integration/integration_api_broker_graceful_shutdown_test.go:196
------------------------------
```

fraenkel:
There again is something definitely wrong with the API server and shutdown.
I modified the test to PUT a new policy and have the scheduler hang. This means there should be an outstanding request.
The moment the SIGUSR2 happens, the server still shuts down. The request sees an EOF rather than a 200. You see below that the creating/updating schedule does not finish.

```
[d][apiServer] spawned /usr/local/bin/node (pid: 32389)
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.565Z","source":"autoscaler:apiserver","text":"Autoscaler API server started","log_level":"info","data":{"port":10001}}
[o][apiServer] Executing (default): SELECT 1+1 AS result
[o][apiServer] Executing (default): SELECT 1+1 AS result
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.925Z","source":"autoscaler:apiserver","text":"DB Connection has been established successfully","log_level":"info"}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.926Z","source":"autoscaler:apiserver","text":"DB Connection has been established successfully","log_level":"info"}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.966Z","source":"autoscaler:apiserver","text":"Input policy JSON schema structure is valid","log_level":"info","data":{"app id":"1486219285580474885"}}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.970Z","source":"autoscaler:apiserver","text":"Input policy JSON validated successfully. Creating policy..","log_level":"info","data":{"app id":"1486219285580474885"}}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.970Z","source":"autoscaler:apiserver","text":"Policy creation request received","log_level":"info","data":{"app id":"1486219285580474885"}}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.971Z","source":"autoscaler:apiserver","text":"Policy has schedules, creating/updating schedules ..","log_level":"info","data":{"app id":"1486219285580474885"}}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.989Z","source":"autoscaler:apiserver","text":"Received SIGUSR2 signal, shutting down gracefully...","log_level":"info"}
[o][apiServer] {"timestamp":"2017-02-04T14:41:25.995Z","source":"autoscaler:apiserver","text":"Everything is cleanly shutdown","log_level":"info"}
```

fraenkel:
I have pushed my changes to https://github.com/cloudfoundry-incubator/app-autoscaler/tree/graceful_shutdown
so you can see how I rewrote the API tests.

fraenkel:
Thanks for the work to get this fixed properly.
I did some minor cleanup.
Waiting on a green build before I close this out.

PR #168, opened 2017-01-31
134933465 : Integration test between scheduler and scaling engine

ghaih:


ghaih:
All the unit tests are working and integration tests working in parallel when skipped API integration tests.

scripts/generate_test_certs.sh
```diff
@@ -37,6 +37,7 @@ certstrap --depot-path ${depot_path} request-cert --passphrase '' --common-name
 certstrap --depot-path ${depot_path} sign api --CA autoscaler-ca
 
 # scheduler certificate
-certstrap --depot-path ${depot_path} request-cert --passphrase 123456 --common-name scheduler --ip 127.0.0.1
+certstrap --depot-path ${depot_path} request-cert --passphrase '' --common-name scheduler --ip 127.0.0.1
 certstrap --depot-path ${depot_path} sign scheduler --CA autoscaler-ca
-openssl pkcs12 -export -in ${depot_path}/scheduler.crt -inkey ${depot_path}/scheduler.key -out ${depot_path}/scheduler.p12 -name scheduler -passin pass:123456 -password pass:123456
+openssl pkcs12 -export -in ${depot_path}/scheduler.crt -inkey ${depot_path}/scheduler.key -out ${depot_path}/scheduler.p12 -name scheduler -password pass:123456
+keytool -import -alias scheduler -file ${depot_path}/autoscaler-ca.crt -keystore ${depot_path}/scheduler.truststore -storeType pkcs12 -storepass 123456 -noprompt
```
fraenkel: pkcs12 is a standard format. It seems odd that we have to use keytool to generate a pkcs12 file from a pkcs12 file.
Having a password is meaningless since it is exposed.

src/integration/components.go
```diff
@@ -29,7 +30,7 @@ const (
 	ScalingEngine    = "scalingEngine"
 )
 
-var testCertDir string = "../../test-certs"
+var testCertDir string = os.Getenv("GOPATH") + "/test-certs"
```
fraenkel: This is a bit dangerous. A GOPATH can have multiple paths.

src/integration/integration_api_scheduler_test.go
```diff
@@ -30,7 +32,9 @@ var _ = Describe("Integration_Api_Scheduler", func() {
 		httpClient.Transport.(*http.Transport).TLSClientConfig = apiTLSConfig
 		httpClient.Timeout = apiSchedulerHttpRequestTimeout
 		apiServerConfPath = components.PrepareApiServerConfig(components.Ports[APIServer], dbUrl, fmt.Sprintf("https://127.0.0.1:%d", components.Ports[Scheduler]), tmpDir)
+		schedulerConfPath = components.PrepareSchedulerConfig(dbUrl, fmt.Sprintf("https://127.0.0.1:%d", components.Ports[ScalingEngine]), tmpDir)
 		startApiServer()
+		schedulerProcess = startScheduler()
```
fraenkel: Do we need to start/stop the scheduler for each test or is it sufficient to start/stop it once for all tests?

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+			It("creates active schedule in scaling engine", func() {
+				Expect(getNumberOfActiveSchedules(testAppId)).To(Equal(0))
+				resp, err := createSchedule(testAppId, policyStr)
+				Expect(err).NotTo(HaveOccurred())
+				defer resp.Body.Close()
+
+				body, err := ioutil.ReadAll(resp.Body)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(string(body)).To(Equal(""))
```
fraenkel: Expect(body).To(HaveLen(0))

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+				Expect(resp.StatusCode).To(Equal(http.StatusOK))
+				Eventually(getAppIdOfActiveSchedules, 5*time.Minute).Should(ContainElement(testAppId))
```
fraenkel: Why 5 minutes?

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+				Expect(getNumberOfActiveSchedules(testAppId)).To(Equal(1))
+
+			})
+		})
+
+		Context("ScalingEngine Server is down", func() {
+			BeforeEach(func() {
+				stopAll()
```
fraenkel: Why would I expect stopAll to just stop the scalingengine?

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+
+			It("should return 500", func() {
+				resp, err := createSchedule(testAppId, policyStr)
+				Expect(err).NotTo(HaveOccurred())
+				defer resp.Body.Close()
+
+				body, err := ioutil.ReadAll(resp.Body)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(string(body)).To(Equal(""))
+				Expect(resp.StatusCode).To(Equal(http.StatusOK))
```
fraenkel: The test says 500 and the code says 200.

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+				Eventually(stdOutbufferMap[Scheduler], 5*time.Minute).Should(gbytes.Say("Error connecting to scaling engine, failed with error: .* for app id: " + testAppId))
+				Expect(getNumberOfActiveSchedules(testAppId)).To(Equal(0))
```
fraenkel: We should think about this test.
Looking at the logs is prone for eventual failure.
A different way to look at this is to Consistently not have any active schedules for a period, start the scaling engine and see we have 1 active schedule.

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+			})
+		})
+
+	})
+
+	Describe("Delete Schedule", func() {
+		BeforeEach(func() {
+			resp, err := createSchedule(testAppId, policyStr)
```
fraenkel: Maybe make this a function since it is used a few times.

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+			Expect(getNumberOfActiveSchedules(testAppId)).To(Equal(1))
+			resp, err := deleteSchedule(testAppId)
+			Expect(err).NotTo(HaveOccurred())
+			defer resp.Body.Close()
+
+			body, err := ioutil.ReadAll(resp.Body)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(body)).To(Equal(""))
+			Expect(resp.StatusCode).To(Equal(http.StatusNoContent))
+			Eventually(getAppIdOfActiveSchedules, 5*time.Minute).ShouldNot(ContainElement(testAppId))
```
fraenkel: Why not just combine the 2 lines.
Eventually(func() int {
return getNumberOfActiveSchedules...
}).Should(BeZero())

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -0,0 +1,161 @@
...
+			Expect(getNumberOfActiveSchedules(testAppId)).To(Equal(0))
+		})
+
+		Context("ScalingEngine Server is down", func() {
```
fraenkel: Why is the here again?

src/integration/integration_suite_test.go
```diff
@@ -73,7 +73,8 @@ var (
 	httpClient *http.Client
 	logger     lager.Logger
 
-	testCertDir string = "../../test-certs"
+	rootDir     string = os.Getenv("GOPATH")
```
fraenkel: again dangerous.

src/integration/integration_suite_test.go
```diff
@@ -278,6 +281,50 @@ func getSchedules(appId string) (*http.Response, error) {
...
+		err := rows.Scan(&appId)
+		if err != nil {
+			Fail("Fail")
+		}
+		appIds = append(appIds, appId)
+	}
+	return appIds
+}
+
+func getNumberOfActiveSchedules(appId string) int {
```
fraenkel: This seems wrong for an integration test. We should be using APIs to test from the outside.

scheduler/src/main/resources/applicationContext.xml
```diff
@@ -76,8 +80,30 @@
...
                 <prop key="org.quartz.jobStore.driverDelegateClass">org.quartz.impl.jdbcjobstore.PostgreSQLDelegate
                 </prop>
+                <prop key="org.quartz.jobStore.isClustered">true</prop>
+                <prop key="org.quartz.scheduler.instanceId">AUTO</prop>
             </props>
         </property>
     </bean>
     <bean id="restTemplate" class="org.springframework.web.client.RestTemplate"/>
+
+    <bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
```
ghaih: We tried using org.apache.http.client.HttpClient, and worked fine. Should we go ahead with it and then donot have to use the system properties.

fraenkel: yes

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestConfiguration.java
```diff
@@ -4,6 +4,6 @@
 
 @TestPropertySource(properties = { "scalingenginejob.reschedule.interval.millisecond=100",
 		"scalingenginejob.reschedule.maxcount=5", "autoscaler.scalingengine.url=http://localhost:8090",
-		"scalingengine.notification.reschedule.maxcount=2" })
+		"scalingengine.notification.reschedule.maxcount=2", "server.ssl.enabled=false" })
```
ghaih: No we donot, specified server.ssl.enabled=false property because by specifying server.ssl properties in application.properties, tomcat goes looking for the files  and junit test cases fail if keystore file etc are not found. Also embedded tomcat looks for them in temp directory if relative path is given.

scheduler/src/main/resources/application.properties
```diff
@@ -15,8 +15,12 @@ autoscaler.scalingengine.url=http://test-scalingengine-test
...
 #ssl
 server.ssl.key-store=../test-certs/scheduler.p12
-caCert=../test-certs/autoscaler-ca.crt
 server.ssl.key-alias=scheduler
 server.ssl.key-store-password=123456
-#server.ssl.key-password=123456
 server.ssl.key-store-type=PKCS12
+client.ssl.keyStore=../test-certs/scheduler.p12
+client.ssl.keyStorePassword=123456
+client.ssl.trustStore=../test-certs/scheduler.truststore
```
ghaih: There will be one truststore for scheduler. Now for keystore there can be two, one for server and one for  client. Having one keystore is also fine. But still better to have separate properties for server and client as the server .ssl... properties are are required for server authentication and are recognised by springboot. Also when doing client authentication when talking to scaling engine specifying server.ssl.. properties are a misnomer. What are your thoughts?

scheduler/src/main/resources/application.properties
```diff
@@ -15,8 +15,12 @@ autoscaler.scalingengine.url=http://test-scalingengine-test
 
 #ssl
 server.ssl.key-store=../test-certs/scheduler.p12
-caCert=../test-certs/autoscaler-ca.crt
 server.ssl.key-alias=scheduler
 server.ssl.key-store-password=123456
-#server.ssl.key-password=123456
 server.ssl.key-store-type=PKCS12
+client.ssl.keyStore=../test-certs/scheduler.p12
+client.ssl.keyStorePassword=123456
```
ghaih: Java needs keystore password.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestConfiguration.java
ghaih: Now since we implemented SSL using org.apache.http.client.HttpClient, the unit tests will fail, so is it ok to have a flag to specify enable/disable SSL. Also in general it would be good to have a flag because of adding SSL support using HTTPClient...it is too closed and will give issues if certificates are not there. What are your thoughts?

scripts/generate_test_certs.sh
ghaih: Keytool here is generate truststore.

src/integration/integration_scheduler_scalingengine_test.go
ghaih: Yes incorrect. Were trying something, missed to change it back.

scheduler/src/main/resources/application.properties
fraenkel: Let's do separate properties.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestConfiguration.java
fraenkel: Why?  Why aren't we setting things to make unit tests pass as well? We do it for the others.

ghaih: Was not sure to proceed wrt my initial reply.So we should have unit tests SSL enabled?

fraenkel: Yes.

src/integration/integration_suite_test.go
ghaih: We will use scaling engine API for  checking active schedule but still need it for test cases where in scaling engine is down.

ghaih:
The travis bulid was failing ref, https://travis-ci.org/cloudfoundry-incubator/app-autoscaler/builds/202483378. So modified travis yaml. Just for information openjdk 8 is supported on trusty as of now. https://github.com/travis-ci/travis-ci/issues/6483

scheduler/src/main/resources/application.properties
```diff
@@ -11,16 +11,18 @@ scalingenginejob.reschedule.maxcount=6
 scalingengine.notification.reschedule.maxcount=3
 
 # scaling engine url
-autoscaler.scalingengine.url=http://test-scalingengine-test
+autoscaler.scalingengine.url=https://scalingengine
 
 #ssl
-server.ssl.key-store=../test-certs/scheduler.p12
+server.ssl.key-store=classpath:scheduler.p12
```
fraenkel: Why is this a classpath and the others are relative paths?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/EmbeddedTomcatUtil.java
```diff
@@ -69,6 +77,18 @@ public void setup(String appId, int statusCode, String message) {
 		appContext.addServletMapping(url, appId);
 	}
 
+	private void setupSSLConfig(Connector httpsConnector) {
+		httpsConnector.setPort(8091);
+		httpsConnector.setSecure(true);
+		httpsConnector.setScheme("https");
+		httpsConnector.setAttribute("keystoreFile", "../src/test/resources/certs/fake-scalingengine.p12");
```
fraenkel: Where do we document how to generate this?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/EmbeddedTomcatUtil.java
```diff
@@ -69,6 +77,18 @@ public void setup(String appId, int statusCode, String message) {
...
+		httpsConnector.setAttribute("keyAlias", "fake-scalingengine");
+		httpsConnector.setAttribute("keystorePass", "123456");
+		httpsConnector.setAttribute("clientAuth", "false");
```
fraenkel: We are doing mutual SSL.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestConfiguration.java
```diff
@@ -3,7 +3,17 @@
 import org.springframework.test.context.TestPropertySource;
 
 @TestPropertySource(properties = { "scalingenginejob.reschedule.interval.millisecond=100",
-		"scalingenginejob.reschedule.maxcount=5", "autoscaler.scalingengine.url=http://localhost:8090",
-		"scalingengine.notification.reschedule.maxcount=2", "server.ssl.enabled=false" })
+		"scalingenginejob.reschedule.maxcount=5", "autoscaler.scalingengine.url=https://localhost:8091",
+		"scalingengine.notification.reschedule.maxcount=2",
+		"server.ssl.key-store=classpath:certs/scheduler.p12",
```
fraenkel: similar issue as above.

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -159,3 +108,19 @@ func setPolicyDateTime(policyByte []byte) string {
 
 	return fmt.Sprintf(string(policyByte), timeZone, startTime, timeNowInTimeZone.Add(2*time.Hour).Format(dateTimeFormat))
 }
+
+func checkResponse(resp *http.Response, err error, expectedStatus int) {
+	Expect(err).NotTo(HaveOccurred())
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(body).To(HaveLen(0))
```
fraenkel: Seems a bit weird to have a method that always assumes a 0 length body.

ghaih: Will rename the method, since expecting some length when not 0 will not make sense. Can keep it same and return body but think renaming better.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/EmbeddedTomcatUtil.java
ghaih: Will add the truststore.

scheduler/src/main/resources/application.properties
ghaih: I'm not sure what we were missing previously. Tomcat was picking the file only when classpath was specified. Still in application.properties could have left the relative path,. But TestConfiguration required the classpath for tomcat. After your comment tried giving the relative path just to be sure and it worked. Not sure what why was the behaviour different previously. Will change to relative path.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/EmbeddedTomcatUtil.java
ghaih: Will update the scheduler readme.

fraenkel:
The latest commit did not address all my comments.

scripts/generate_test_certs.sh
ghaih: Java needs keystore password.

src/integration/integration_scheduler_scalingengine_test.go
ghaih: Yes this one was not properly understood, so done in an incorrect fashion. Modified now and got rid of the db call.

PR #170, opened 2017-02-09
acceptance test for service broker

boyang9527:
[139440941]

- changes only for running service broker acceptance test
- other packages not changed yet, will do in other commits


fraenkel:
Given our 2 repos, I believe the acceptance tests should live in the app-autoscaler-release repo. If I change the acceptance test, the autoscaler repo shouldn't be affected. The acceptance tests should change as the app-autoscaler repo changes which is what the release repo is for.

boyang9527:
Ok... then I will move the acceptance test to release repo.

PR #171, opened 2017-02-09
fix missing triggers defect

boyang9527:
[#139452839]

- fix the defect
- add test cases

### DIFF
--- a/src/autoscaler/eventgenerator/generator/evaluator_test.go
+++ b/src/autoscaler/eventgenerator/generator/evaluator_test.go
@@ -64,26 +64,26 @@ var _ = Describe("Evaluator", func() {
 			Adjustment:            "1",
 		}}
 
-		triggerArrayMultipleTriggers []*models.Trigger = []*models.Trigger{
-			&models.Trigger{
-				AppId:                 testAppId,
-				MetricType:            testMetricType,
-				BreachDurationSeconds: 300,
-				CoolDownSeconds:       300,
-				Threshold:             500,
-				Operator:              ">=",
-				Adjustment:            "1",
-			},
-			&models.Trigger{
-				AppId:                 testAppId,
-				MetricType:            testMetricType,
-				BreachDurationSeconds: 300,
-				CoolDownSeconds:       300,
-				Threshold:             500,
-				Operator:              "<=",
-				Adjustment:            "1",
-			},
+		firstTrigger = models.Trigger{
+			AppId:                 testAppId,
+			MetricType:            testMetricType,
+			BreachDurationSeconds: 300,
+			CoolDownSeconds:       300,
+			Threshold:             500,
+			Operator:              ">=",
+			Adjustment:            "1",
 		}
+
+		secondTrigger = models.Trigger{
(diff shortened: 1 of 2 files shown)


src/autoscaler/eventgenerator/generator/evaluator_test.go
```diff
@@ -467,10 +527,66 @@ var _ = Describe("Evaluator", func() {
...
+				})
+
+				Context("when both tiggers breach", func() {
+					BeforeEach(func() {
+						database.RetrieveAppMetricsStub = func(appId string, metricType string, start int64, end int64) ([]*models.AppMetric, error) {
+							return appMetricMultipleTriggerAllBreach, nil
+						}
+					})
+					It("should send one trigger alarm to scaling engine", func() {
+						Eventually(scalingEngine.ReceivedRequests).Should(HaveLen(1))
```
fraenkel: If both trigger, why would we only have 1 request?

src/autoscaler/eventgenerator/generator/evaluator_test.go
```diff
@@ -467,10 +527,66 @@ var _ = Describe("Evaluator", func() {
...
+
+				Context("when no tigger breaches", func() {
+					BeforeEach(func() {
+						database.RetrieveAppMetricsStub = func(appId string, metricType string, start int64, end int64) ([]*models.AppMetric, error) {
+							return appMetricMultipleTriggerAllNotBreach, nil
+						}
+					})
+					It("should not send trigger alarm to scaling engine", func() {
+						Eventually(scalingEngine.ReceivedRequests).Should(HaveLen(0))
+						Eventually(logger.LogMessages).Should(ContainElement(ContainSubstring("should not send trigger alarm to scaling engine")))
```
fraenkel: This isn't so important since we have the above.

src/autoscaler/eventgenerator/generator/evaluator_test.go
boyang9527: It depends on how we handle trigger conflict.  There are two options
1) first breach wins
2) trigger scaling for all the breaches
for 2, usually the second one will be ignored since there is cool down period control.  The scaling action can be taken only if the previous triggers did not trigger actual scaling for whatever reason(failed or ignored)

boyang9527: 2 will need to explain more to users

src/autoscaler/eventgenerator/generator/evaluator_test.go
boyang9527: yes, you are right. 

I will remove this one once we agreed the strategy on how to handle the trigger conflict.

src/autoscaler/eventgenerator/generator/evaluator_test.go
fraenkel: The problem I have is that the test is no different than 1 breach. So I don't see how this really tests what it states if it looks identical.

boyang9527:
@fraenkel  I added the verification of trigger content in multiple trigger test

PR #172, opened 2017-02-12
139520515: Add support for API to get Active schedules for app  in scaling engine

ghaih:
Also minor typo Fix, Removed unused variables

### DIFF
--- a/src/autoscaler/routes/routes.go
+++ b/src/autoscaler/routes/routes.go
@@ -11,14 +11,16 @@ const (
 	MemoryMetricRoute        = "memory-metric"
 	MemoryMetricHistoryRoute = "memory-metric-histories"
 
-	scalePath            = "/v1/apps/{appid}/scale"
-	scalingHistoriesPath = "/v1/apps/{appid}/scaling_histories"
-	activeSchedulePath   = "/v1/apps/{appid}/active_schedules/{scheduleid}"
+	scalePath             = "/v1/apps/{appid}/scale"
+	scalingHistoriesPath  = "/v1/apps/{appid}/scaling_histories"
+	activeSchedulePath    = "/v1/apps/{appid}/active_schedules/{scheduleid}"
+	appActiveSchedulePath = "/v1/apps/{appid}/active_schedules"
 
 	ScaleRoute                 = "scale"
-	HistoreisRoute             = "histories"
+	HistoriesRoute             = "histories"
 	UpdateActiveSchedulesRoute = "updateActiveSchedules"
 	DeleteActiveSchedulesRoute = "deleteActiveSchedules"
+	GetActiveScheduleRoute     = "getActiveSchedule"
 )
 
 type AutoScalerRoute struct {
@@ -38,9 +40,10 @@ func newRouters() *AutoScalerRoute {
 	instance.metricsCollectorRoutes.Path(memoryMetricHistoriesPath).Name(MemoryMetricHistoryRoute)
 
 	instance.scalingEngineRoutes.Path(scalePath).Name(ScaleRoute)
-	instance.scalingEngineRoutes.Path(scalingHistoriesPath).Name(HistoreisRoute)
+	instance.scalingEngineRoutes.Path(scalingHistoriesPath).Name(HistoriesRoute)
 	instance.scalingEngineRoutes.Path(activeSchedulePath).Name(UpdateActiveSchedulesRoute)
 	instance.scalingEngineRoutes.Path(activeSchedulePath).Name(DeleteActiveSchedulesRoute)
+	instance.scalingEngineRoutes.Path(appActiveSchedulePath).Name(GetActiveScheduleRoute)
 
 	return instance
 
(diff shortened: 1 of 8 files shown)


src/autoscaler/routes/routes.go
```diff
@@ -38,9 +40,10 @@ func newRouters() *AutoScalerRoute {
 	instance.metricsCollectorRoutes.Path(memoryMetricHistoriesPath).Name(MemoryMetricHistoryRoute)
 
 	instance.scalingEngineRoutes.Path(scalePath).Name(ScaleRoute)
-	instance.scalingEngineRoutes.Path(scalingHistoriesPath).Name(HistoreisRoute)
+	instance.scalingEngineRoutes.Path(scalingHistoriesPath).Name(HistoriesRoute)
```
fraenkel: This should have been a separate commit.

src/autoscaler/routes/routes_test.go
```diff
@@ -97,23 +97,23 @@ var _ = Describe("Routes", func() {
 		Context("HistoreisRoute", func() {
```
fraenkel: You missed the spelling change

src/autoscaler/scalingengine/server/scaling_handler_test.go
```diff
@@ -427,4 +429,55 @@ var _ = Describe("ScalingHandler", func() {
...
+			})
+
+			It("returns 200 with active schedule in message body", func() {
+				Expect(resp.Code).To(Equal(http.StatusOK))
+
+				actualActiveSchedule := &models.ActiveSchedule{}
+				err = json.Unmarshal(resp.Body.Bytes(), actualActiveSchedule)
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(*actualActiveSchedule).To(Equal(*activeSchedule))
```
fraenkel: Why the dereference of pointers on both sides?

src/autoscaler/scalingengine/server/server_test.go
```diff
@@ -231,5 +231,47 @@ var _ = Describe("Server", func() {
...
+					rsp.Body.Close()
+				})
+			})
+
+			Context("when using the wrong method", func() {
+				BeforeEach(func() {
+					method = http.MethodDelete
+				})
+
+				It("should return 404", func() {
```
fraenkel: This doesn't seem right.  It should be 405?

src/autoscaler/scalingengine/server/server_test.go
```diff
@@ -231,5 +231,47 @@ var _ = Describe("Server", func() {
...
+					Expect(err).ToNot(HaveOccurred())
+					Expect(rsp.StatusCode).To(Equal(http.StatusNotFound))
+					rsp.Body.Close()
+				})
+			})
+
+			Context("when requesting the wrong path", func() {
```
fraenkel: This test doesn't make much sense since it is not testing getting the active schedule since you have picked a random path.

src/autoscaler/scalingengine/server/server_test.go
ghaih: Yes copy paste from above, will fix rest as well.

src/autoscaler/scalingengine/server/server_test.go
ghaih: Yes it is a repeat, will remove it. We should add test case for invalid appid?

fraenkel: Yes, there should be a test when the appid is not found aka invalid versus validation of the id which should also be there if we did it but I don't think we do.

src/autoscaler/scalingengine/server/server_test.go
ghaih: Currently the mux library implementation cannot distinguish between 404 and 405
Looked at the code, the routes and methods don't seem to be connected

https://github.com/gorilla/mux/issues/6

fraenkel: Then it becomes a pointless test since 404 doesn't make any sense to me.

src/autoscaler/scalingengine/server/scaling_handler_test.go
ghaih: You mean , we can save value in one of them,actualActiveSchedule := models.ActiveSchedule{} instead of actualActiveSchedule := &models.ActiveSchedule{}

src/autoscaler/scalingengine/server/server_test.go
ghaih: Yes. Should get rid of it wherever used?

src/autoscaler/scalingengine/server/scaling_handler_test.go
fraenkel: Equal handles pointers properly and will compare values.

ghaih: Ok got it. Great feature.

fraenkel:
You need to rebase.

PR #173, opened 2017-02-13
unify the memory metric name to be "memoryused" with unit "megabytes"

boyang9527:
[#139247331]

### DIFF
--- a/src/autoscaler/eventgenerator/generator/evaluator.go
+++ b/src/autoscaler/eventgenerator/generator/evaluator.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"time"
 )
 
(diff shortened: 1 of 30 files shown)


src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -77,31 +78,38 @@ func (e *Evaluator) doEvaluate(triggerArray []*models.Trigger) {
 
 		isBreached := true
 		for _, appMetric := range appMetricList {
-			if appMetric.Value == nil {
-				e.logger.Debug("should not send trigger alarm to scaling engine because there is nil-value metric", lager.Data{"trigger": trigger, "appMetric": appMetric})
+			if appMetric.Value == "" {
```
fraenkel: next time the better way is len(appMetric.Value) == 0 which covers nil and empty string.

src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -77,31 +78,38 @@ func (e *Evaluator) doEvaluate(triggerArray []*models.Trigger) {
...
 				break
 			}
+			value, err := strconv.ParseInt(appMetric.Value, 10, 64)
+			if err != nil {
+				e.logger.Debug("should not send trigger alarm to scaling engine because parse metric value fails", lager.Data{"trigger": trigger, "appMetric": appMetric})
+				isBreached = false
+				break
+			}
+
 			if operator == ">" {
```
fraenkel: Would have replaced this with a switch.

PR #175, opened 2017-02-19
140195477: removed memory threshold limits in apiserver validation.

rohitsharma04:


api/lib/validation/schemaValidator.js
```diff
@@ -93,7 +93,7 @@ var getScalingRuleSchema = function() {
       'metric_type':{ 'type':'string' ,'enum':metricTypeEnum },
       'stat_window_secs':{ 'type':'number','minimum': 60,'maximum': 3600 },
       'breach_duration_secs':{ 'type':'number','minimum': 60,'maximum': 3600 },
-      'threshold':{ 'type':'number','minimum': 1,'maximum': 100 },
```
fraenkel: Given the threshold must be >= 1 why isn't the minimum still valid?

pradyutsarma: The value could be between 1-100 only when we were looking at usage and not MBs. It could just be any >0 number now.

fraenkel: I know which is why I am asking why we removed the minimum. If it is > 0, we must have a value which is >= 1. Rather than code the check above, why not have the validator deal with it?

rohitsharma04:
Updated

fraenkel:
Thanks.

PR #176, opened 2017-02-21
#140295751: Removed unused table applicationmetrics

rohitsharma04:


fraenkel:
Good catch. Since we haven't shipped a release, one can get away with just deleting the table from the changelog rather than dropping which would be needed if we weren't rebuilding the dbs.

rohitsharma04:
Should I delete the table from the existing changeset and update the PR ?

fraenkel:
Yes.
Before we release v1, we will review the changelog and go to a single entry per log.
There is no point in having the db do more than it needs until we absolutely need to.

rohitsharma04:
Updated.

PR #179, opened 2017-02-22
syncer is a module

qibobo:


src/autoscaler/db/sqldb/scheduler_sqldb.go
```diff
@@ -82,3 +83,23 @@ func (sdb *SchedulerSQLDB) GetActiveSchedules() (map[string]*models.ActiveSchedu
 	return schedules, nil
 
 }
+
+func (sdb *SchedulerSQLDB) SynchronizeActiveSchedules(appIdMap map[string]bool) error {
```
fraenkel: for future reference, a set in Go is map[string]struct{}. The size of a struct{} in Go is 0.

src/autoscaler/db/sqldb/scheduler_sqldb_test.go
```diff
@@ -123,4 +124,83 @@ var _ = Describe("SchedulerSqldb", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		JustBeforeEach(func() {
+			schedules, err = sdb.GetActiveSchedules()
+		})
+		Context("when appIdMap is empty", func() {
+			BeforeEach(func() {
+				appIdMap = map[string]bool{}
+				err = sdb.SynchronizeActiveSchedules(appIdMap)
```
fraenkel: Why isn't this done in a higher JBE?
All you are doing in the BE is setting the appIdMap.

src/autoscaler/db/sqldb/scheduler_sqldb_test.go
```diff
@@ -123,4 +124,83 @@ var _ = Describe("SchedulerSqldb", func() {
...
+		})
+		Context("when appIdMap is not empty", func() {
+			BeforeEach(func() {
+				appIdMap = map[string]bool{
+					"app-id-2": true,
+				}
+				err = sdb.SynchronizeActiveSchedules(appIdMap)
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("active schedules of app which appId is not in appIdMap are removed", func() {
```
fraenkel: Why not just check the length? It should be one and the only item should be for app-id-1.

fraenkel:
I am a bit confused. This exposes an endpoint to synchronize. Who is going to call it?

src/autoscaler/db/sqldb/scheduler_sqldb_test.go
qibobo: JBE means?

src/autoscaler/db/sqldb/scheduler_sqldb.go
```diff
@@ -82,3 +83,23 @@ func (sdb *SchedulerSQLDB) GetActiveSchedules() (map[string]*models.ActiveSchedu
...
+	if len(appIdMap) == 0 {
+		sdb.logger.Debug("No application exists")
+		return nil
+	}
+	var query string = "DELETE FROM app_scaling_active_schedule WHERE app_id NOT IN("
```
kanekoh: When using NOT IN, the index won't be used for executing the SQL.

qibobo: Known it.  This is the easier way and we can optimize it later.

kanekoh: ok

fraenkel: I would rather we do this the right way and get the simple wins. It is easier to just collect the app guids from the active schedules and just determine the ones we need to remove.

kanekoh: Just adding to the NOT IN discussion, if the input to NOT IN gets too large, the query might fail.

qibobo:
@fraenkel  reset the code and update.

src/autoscaler/db/sqldb/scheduler_sqldb.go
fraenkel: Since we are keeping this, it should be renamed to what it actually does.
1. This only does Pruning and should follow the conventions of the Pruners.
1. Let us do this the right way and grab the active app ids and compare. We can easily build a loop which will remove X number at a time where X can start with 50. This way we can use the index.

src/autoscaler/syncer/activescheduledb.go
```diff
@@ -0,0 +1,44 @@
...
+func NewActiveScheduleSyncer(policyDb db.PolicyDB, schedulerDb db.SchedulerDB, logger lager.Logger) *ActiveScheduleSyncer {
+
+	return &ActiveScheduleSyncer{
+		policyDb:    policyDb,
+		schedulerDb: schedulerDb,
+		logger:      logger,
+	}
+}
+
+func (ass *ActiveScheduleSyncer) Synchronize() error {
```
fraenkel: See above comment. It's a Pruner

PR #180, opened 2017-02-23
140444579: Add missing server down test case api-scheduler integration test.

kanekoh:
Will apply the code refactor to all other integration test in a separate story.

### DIFF
--- a/src/integration/integration_api_scheduler_test.go
+++ b/src/integration/integration_api_scheduler_test.go
@@ -70,6 +70,34 @@ var _ = Describe("Integration_Api_Scheduler", func() {
 				Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
 				resp.Body.Close()
 			})
+
+			Context("Scheduler is down", func() {
+				BeforeEach(func() {
+					stopScheduler(schedulerProcess)
+
+					policyStr = readPolicyFromFile("fakePolicyWithSchedule.json")
+					resp, err := attachPolicy(appId, policyStr)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(resp.StatusCode).To(Equal(http.StatusInternalServerError))
+					resp.Body.Close()
+
+					schedulerProcess = startScheduler()
+				})
+
+				It("should neither create policy nor schedule", func() {
+					By("checking the API Server")
+					resp, err := getPolicy(appId)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
+					resp.Body.Close()
+
+					By("checking the Scheduler")
+					resp, err = getSchedules(appId)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
+					resp.Body.Close()
+				})
+			})
 		})
 
 		Context("Policies without schedules", func() {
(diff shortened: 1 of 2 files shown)


src/integration/integration_api_scheduler_test.go
```diff
@@ -70,6 +70,34 @@ var _ = Describe("Integration_Api_Scheduler", func() {
...
+				BeforeEach(func() {
+					stopScheduler(schedulerProcess)
+
+					policyStr = readPolicyFromFile("fakePolicyWithSchedule.json")
+					resp, err := attachPolicy(appId, policyStr)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(resp.StatusCode).To(Equal(http.StatusInternalServerError))
+					resp.Body.Close()
+
+					schedulerProcess = startScheduler()
```
fraenkel: Why is this here?
It should be in an AfterEach.

src/integration/integration_api_scheduler_test.go
```diff
@@ -94,6 +122,34 @@ var _ = Describe("Integration_Api_Scheduler", func() {
...
+				BeforeEach(func() {
+					stopScheduler(schedulerProcess)
+
+					policyStr = readPolicyFromFile("fakePolicyWithoutSchedule.json")
+					resp, err := attachPolicy(appId, policyStr)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(resp.StatusCode).To(Equal(http.StatusCreated))
+					resp.Body.Close()
+
+					schedulerProcess = startScheduler()
```
fraenkel: AfterEach

I would restructure the tests slightly.
Pull the "Scheduler is down" up, and really we only need 1 test since with or without schedule should affect the outcome but I would do it with a schedule since that would be something the scheduler participates in.

src/integration/integration_api_scheduler_test.go
```diff
@@ -168,6 +224,30 @@ var _ = Describe("Integration_Api_Scheduler", func() {
 				Expect(resp.StatusCode).To(Equal(http.StatusNotFound))
 				resp.Body.Close()
 			})
+
+			Context("Scheduler is down", func() {
```
fraenkel: Move this.

src/integration/integration_api_scheduler_test.go
kanekoh: The reason is to check if there are no schedules in scheduler or not with REST API.
Is it enough to check API server respond status is Internal Server Error when API server received a policy with schedules while schedule is down?

fraenkel: Sorry, my comment was the startScheduler.
The context is when the scheduler is down. Therefore we should never start the scheduler.
It should be started in the AfterEach, i.e., after the test is complete not before.

kanekoh: I was not clear in my explanation before. The startScheduler is part of the test case to verify the state of the data. Maybe better to have it in the "It" section. But still I understand we should have another start scheduler in the afterEach in case something goes wrong in the middle.

src/integration/integration_api_scheduler_test.go
kanekoh: Yes understand your point about having only with schedule.
The point we were trying to make by having a test without schedule was to verify the scheduler doesn't participate when there is no schedule.

src/integration/integration_api_scheduler_test.go
fraenkel: You don't need to verify the scheduler since the API server will tell us the answer.

src/integration/integration_api_scheduler_test.go
fraenkel: If we know it is down, then we also know the data cannot change.

kanekoh:
ok, will fix them.

PR #183, opened 2017-03-01
#140730873 : Metrics collector needs to register the active service

rohitsharma04:


src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
```diff
@@ -45,7 +45,8 @@ var _ = Describe("MetricsCollector", func() {
 			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say(runner.acquiredLockCheck))
 		})
 
-		It("should start", func() {
+		It("should register itself as the active instance and start", func() {
```
fraenkel: These don't actually test much since you are looking at logs rather than actual facts.
See https://github.com/cloudfoundry/bbs/blob/master/cmd/bbs/service_registration_test.go

PR #184, opened 2017-03-01
140540769: Fix Properties and messages in scheduler

kanekoh:


scheduler/src/main/resources/application.properties
```diff
@@ -1,28 +1,26 @@
...
-spring.datasource.driverClassName=org.postgresql.Driver
-spring.datasource.url=jdbc:postgresql://127.0.0.1/autoscaler
-spring.datasource.username=postgres
-spring.datasource.password=postgres
-
-#quartz job
-scalingenginejob.reschedule.interval.millisecond=10000
-scalingenginejob.reschedule.maxcount=6
-scalingengine.notification.reschedule.maxcount=3
+# datasource for application and quartz
```
fraenkel: Is this even needed anymore? There is no content.

kanekoh: We kept application.properties for reference only. Just to know what properties are being used.
If you think it's not required, we can remove it?

fraenkel: Git will always remember it but if we don't need it, it is more confusing to have it.

kanekoh: OK, We will remove it.

PR #185, opened 2017-03-03
#140679451 : Implement HA for the event generator

rohitsharma04:
- Added consul based lock in eventgenerator.
- Eventgenerator does not expose a server. So there's no service registration. If this is true, then we will create another PR to do some cleanup of the eventgenerator.

### DIFF
--- a/src/autoscaler/eventgenerator/cmd/eventgenerator/eventgenerator_suite_test.go
+++ b/src/autoscaler/eventgenerator/cmd/eventgenerator/eventgenerator_suite_test.go
@@ -18,6 +18,8 @@ import (
 	"time"
 
 	"code.cloudfoundry.org/cfhttp"
+	"code.cloudfoundry.org/consuladapter/consulrunner"
+	"code.cloudfoundry.org/locket"
 	"github.com/onsi/gomega/gbytes"
 	"github.com/onsi/gomega/gexec"
 	"github.com/onsi/gomega/ghttp"
@@ -34,6 +36,7 @@ var (
 	conf            *config.Config
 	metricCollector *ghttp.Server
 	scalingEngine   *ghttp.Server
+	consulRunner    *consulrunner.ClusterRunner
 	metrics         []*models.AppInstanceMetric = []*models.AppInstanceMetric{
 		&models.AppInstanceMetric{
 			AppId:         testAppId,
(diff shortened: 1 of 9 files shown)


fraenkel:
While it is not a service, I would still add the service registration so one can easily locate the current event generator. Looking up in DNS would be easier than querying consul to find the active event generator.

src/autoscaler/eventgenerator/cmd/eventgenerator/eventgenerator_suite_test.go
```diff
@@ -88,16 +91,31 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+		consulRunner.Stop()
+	}
 }, func() {
 	gexec.CleanupBuildArtifacts()
 })
 
+func initConsul() {
+	consulRunner = consulrunner.NewClusterRunner(
+		consulrunner.ClusterRunnerConfig{
+			StartingPort: 9001,
```
fraenkel: How does this work when you run in parallel?

src/autoscaler/eventgenerator/cmd/eventgenerator/eventgenerator_test.go
```diff
@@ -1,42 +1,112 @@
...
+			runner.Start()
+
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say(runner.acquiredLockCheck))
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say("eventgenerator.started"))
+
+			consulRunner.Reset()
+		})
+
+		It("exits with failure", func() {
+			Eventually(runner.Session.Buffer, 4*time.Second).Should(gbytes.Say("exited-with-failure"))
```
fraenkel: If you flip these, the check for exited becomes an Expect.

src/autoscaler/eventgenerator/cmd/eventgenerator/eventgenerator_suite_test.go
rohitsharma04: @fraenkel  should I also update the metricscollector to run in parallel?

fraenkel: yes, we should make all of our units run in parallel when possible

PR #186, opened 2017-03-03
[#139969067] Service catalog definition injected through configuration file

paltanmoy:
We will create a separate PR to autoscaler-release repository to inject the catalog definition through manifest, once this PR is merged and bumped up.


fraenkel:
I will not be merging this until we discuss how plans are parameterized. Having more than 1 plan does not provide any value to a user. Until it does, there is no point in having it configurable.

paltanmoy:
Closing this PR. We will create a new PR with configurable Service Name, Service ID, Plan Name and Plan ID.

PR #191, opened 2017-03-08
[#140805269] Different certificate used in APIServer for client and server

paltanmoy:


api/lib/utils/schedulerUtils.js
```diff
@@ -1,11 +1,17 @@
 'use strict';
-module.exports = function(settings, tlsOptions) {
```
fraenkel: Why do we pass in the entire settings when all it needs is the scheduler portion?

api/lib/config/setting.js
```diff
@@ -20,12 +20,12 @@ module.exports = function(settingsObj) {
       return schedulerUri.replace(/\/$/g, "").toLowerCase();
     }
   };
-
+  settingsObj.scheduler.uri = scheduler(settingsObj.scheduler.uri);
```
fraenkel: Can we name this function better? Calling it scheduler isn't really meaningful.

paltanmoy: Renamed and updated.

PR #192, opened 2017-03-08
140444445: Fixed unreachable request to Scheduler application.

kanekoh:
Just for Information.
We observed the message 'Started SchedulerApplication in' is appearing twice after applying Spring Cloud Consul.
First time appears even before tomcat has started which could lead to invalid test so we have changed the message in component.go and We will investigate this in a separate story.

### DIFF
--- a/scheduler/pom.xml
+++ b/scheduler/pom.xml
@@ -45,6 +45,16 @@
 			<scope>provided</scope>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-commons</artifactId>
+			<version>1.1.7.RELEASE</version>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-starter-consul-all</artifactId>
+			<version>1.1.2.RELEASE</version>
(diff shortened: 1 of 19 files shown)


fraenkel:
If you introduced the issue with this change then the fix needs to be apart of this PR.

kanekoh:
Issue was because of multiple dependencies.

fraenkel:
I don't see where you disable registering the health check.

kanekoh:
When I disabled registering the health check, Consul had no checks for scheduler even scheduler was down.

When Spring Cloud Consul is set both parameters 'spring.cloud.consul.discovery.registerHealthCheck=false' and 'spring.cloud.consul.discovery.heartbeat.enabled=true' and Spring Cloud Consul couldn't be completed to register scheduler to consul because Consul agent has no checks for scheduler.

Scheduler had no CLOSE_WAIT connections when Consul agent had checks for scheduler with TTL and noticed if scheduler was down or not.
So, I set the parameters 'spring.cloud.consul.discovery.registerHealthCheck=true' (default) and 'spring.cloud.consul.discovery.heartbeat.enabled=true' to Spring Cloud Consul.

fraenkel:
weird, the docs say otherwise but that is fine for now.

PR #195, opened 2017-03-10
[#140805361] Different certificate used in ServiceBroker for client and server

paltanmoy:


servicebroker/lib/config/setting.js
```diff
@@ -14,16 +14,17 @@ module.exports = function(settingsObj) {
     }
   };
 
-  var apiServer = function(apiServerUri) {
+  var cleanupAPIServerURI = function(apiServerUri) {
```
kanekoh: I just want to understand about this variable name. Does api server url variable need the word 'cleanup'?

paltanmoy: It's the name of the anonymous function used to clean up the apiserver url.

fraenkel: Let us be clear. This is not specific to the APIServer and should have been name as to what it does like removeLastSlash since it would do it for any URL.

PR #197, opened 2017-03-16
135415719: Fixed Integration test failure - API/Scheduler

kanekoh:


fraenkel:
This is still no guarantee that the scheduler is ready.
We should verify the scheduler is ready to receive requests before we start any tests.

kanekoh:
We are trying to register scheduler with consul after schedule is ready. Will update you.

PR #198, opened 2017-03-16
141746315 : Scheduler end job not deleting active schedule from scheduler database.

ghaih:
Fixed Issue and minor code refactoring.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
@@ -43,12 +43,12 @@ public class AppScalingScheduleEndJob extends AppScalingScheduleJob {
 		long scheduleId = jobDataMap.getLong(ScheduleJobHelper.SCHEDULE_ID);
 		long startJobIdentifier = jobDataMap.getLong(ScheduleJobHelper.START_JOB_IDENTIFIER);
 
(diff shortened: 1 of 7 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/quartz/AppScalingScheduleEndJob.java
```diff
@@ -43,12 +43,12 @@ private void deleteActiveSchedule(JobExecutionContext jobExecutionContext) throw
...
 
-		boolean activeScheduleTableTaskDone = jobDataMap.getBoolean(ScheduleJobHelper.ACTIVE_SCHEDULE_TABLE_TASK_DONE);
+		boolean activeScheduleTableDeleteTaskDone = jobDataMap.getBoolean(ScheduleJobHelper.ACTIVE_SCHEDULE_TABLE_DELETE_TASK_DONE);
 
-		if (!activeScheduleTableTaskDone) {
+		if (!activeScheduleTableDeleteTaskDone) {
 			try {
 				activeScheduleDao.delete(scheduleId, startJobIdentifier);
-				jobDataMap.put(ScheduleJobHelper.ACTIVE_SCHEDULE_TABLE_TASK_DONE, true);
+				jobDataMap.put(ScheduleJobHelper.ACTIVE_SCHEDULE_TABLE_DELETE_TASK_DONE, true);
```
fraenkel: Where is the test to verify this is being set?
Why do we bother guarding this? Why not just always attempt the delete?

ghaih: Yes, mistake, blindly followed the code. No need of flag for end job.

PR #199, opened 2017-03-16
[#140730323] Added consul based lock for Pruner

paltanmoy:


src/autoscaler/pruner/cmd/pruner/main.go
```diff
@@ -127,3 +149,26 @@ func getLogLevel(level string) (lager.LogLevel, error) {
...
 }
+
+func initializeRegistrationRunner(
+	logger lager.Logger,
+	consulClient consuladapter.Client,
+	port int,
+	clock clock.Clock) ifrit.Runner {
+	registration := &api.AgentServiceRegistration{
+		Name: "pruner",
+		Port: port,
```
fraenkel: Why do we have a port?

fraenkel:
You didn't remove all Port references you created like DefaultServerPort.

PR #201, opened 2017-03-21
[#139969067] Made Service Name,Service ID,Plan Name and Plan ID configurable

paltanmoy:


servicebroker/lib/index.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 var path = require('path');
 var args = process.argv;
-if (!(args.length == 4 && args[2] == "-c" && args[3] != "")) {
-    throw new Error("missing config file\nUsage:use '-c' option to specify the config file path");
+if (!(args.length == 6 && args[2] == "-c" && args[3] != "" && args[4] == "-sc" && args[5] != "")) {
```
fraenkel: This is getting very messy. We are now requiring positional args which seems wrong.
I will allow it for this PR, but we should just parse args and determine if we have both -c and -sc as required options.

Why didn't we just have the catalog location as part of the single config?

src/integration/components.go
```diff
@@ -64,6 +65,23 @@ type ServiceBrokerConfig struct {
 	HttpRequestTimeout int             `json:"httpRequestTimeout"`
 	TLS                models.TLSCerts `json:"tls"`
 }
+type Service struct {
```
fraenkel: Why do we have this?
We just need the catalog on disk.

servicebroker/lib/index.js
paltanmoy: Should we have the catalog path as part of the settings.json ?

fraenkel: That is what I prefer. It will simplify things.

PR #202, opened 2017-03-22
add updated_at column to specific_date and recurring schedule tables

qibobo:
[#130054353]

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/ScheduleEntity.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/ScheduleEntity.java
@@ -1,5 +1,7 @@
 package org.cloudfoundry.autoscaler.scheduler.entity;
 
+import java.time.LocalDateTime;
+
 import javax.persistence.Column;
 import javax.persistence.GeneratedValue;
 import javax.persistence.GenerationType;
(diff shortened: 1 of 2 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/ScheduleEntity.java
```diff
@@ -64,6 +74,14 @@
 	@Column(name = "initial_min_instance_count")
 	@JsonProperty(value = "initial_min_instance_count")
 	private Integer initialMinInstanceCount;
+	
+	@Column(name = "updated_at")
+	@JsonProperty(value = "updated_at")
+	@JsonFormat(pattern = DateHelper.DATE_TIME_FORMAT)
+	@JsonDeserialize(using = DateTimeDeserializer.class)
+	@JsonSerialize(using = DateTimeSerializer.class)
+	private LocalDateTime updatedAt;
```
kanekoh: Just notify to you all variables in entity have `@ApiModelProperty` for Swagger. What your thought?

qibobo: It is just for internal use and will not be exposed.  So in my opinion "@ApiModelProperty" is not need.

fraenkel:
This isn't right. The problem starts with the API server which has an updated_at field too.
This needs to be sync'd with the scheduler so we know that a given set of schedules is for the same policy.

ghaih:
Why are we creating trigger. If api-server passes the updated_at information across to scheduler, then scheduler can persist it and can use it to compare against the policy's updated_at. Am I missing something. On the other hand i could not find where is api server passing this information, or may be the column was added but passing the info is still pending.

fraenkel:
Correct, we have some new stories to actually generate a GUID and send it across. See the top of backlog. There are a few issues using updated_at and the current create/update order is wrong.

ghaih:
Thanks.Yes did see the stories, just wanted to understand the reason for having trigger.

PR #210, opened 2017-03-30
141897575: Units: metric poller does not save any metrics

kanekoh:


src/autoscaler/eventgenerator/aggregator/metricPoller_test.go
```diff
@@ -78,6 +78,40 @@ var _ = Describe("MetricPoller", func() {
...
+
+		BeforeEach(func() {
+			appMonitor = &models.AppMonitor{
+				AppId:      testAppId,
+				MetricType: metricType,
+				StatWindow: 10,
+			}
+
+			metricServer = ghttp.NewUnstartedServer()
+			metricServer.RouteToHandler("GET", urlPath, ghttp.RespondWithJSONEncoded(http.StatusOK,
```
fraenkel: If we never start it why do we add a route?

kanekoh: Yes, it's a mistake. Will fix it.

PR #211, opened 2017-03-30
#142388235 Fixed Unit : eventgenerator evaluator test failure

paltanmoy:


src/autoscaler/eventgenerator/generator/evaluator_test.go
```diff
@@ -579,11 +579,12 @@ var _ = Describe("Evaluator", func() {
 					database.RetrieveAppMetricsStub = func(appId string, metricType string, start int64, end int64) ([]*models.AppMetric, error) {
 						return appMetricGTBreach, nil
 					}
-					Expect(triggerChan).To(BeSent(triggerArrayGT))
 				})
 				Context("when the send request encounters error", func() {
 					JustBeforeEach(func() {
 						scalingEngine.Close()
+						Eventually(scalingEngine.HTTPTestServer).Should(BeNil())
```
fraenkel: This isn't thread safe. Since the goal is to guarantee the scaling engine is not responding, we should be sending a request until it fails.

paltanmoy: As we are not sending the request directly to Scaling Engine (to check the failure) rather to Evaluator, does this mean that  we should Eventually send the trigger to evaluator channel without waiting for the server to be nil?

kanekoh: @paltanmoy, Sorry to interrupt, I thought this test case means 'when scaling engine is down' so how about using ghttp.NewUnstartedServer instead of ghttp.NewServer and separate context for scaling engine started and not started?

src/autoscaler/eventgenerator/generator/evaluator_test.go
```diff
@@ -613,6 +607,7 @@ var _ = Describe("Evaluator", func() {
 							errorStr = errorStr + tmp
 						}
 						scalingEngine.RouteToHandler("POST", urlPath, ghttp.RespondWithJSONEncoded(http.StatusBadRequest, errorStr))
+						Expect(triggerChan).To(BeSent(triggerArrayGT))
```
julz: looks like we're fixing a pended test here.. does anyone know/remember why it's pended? if we never run it it doesn't seem like it gives us much useful regression coverage, maybe we should either fix it or delete it? I tried running it locally and it seems to consistently fail..

kanekoh: @qibobo I tried to find a reason why it's pended however I haven't found it.
I was wondering if the below story is the first commit to add this file. Do you know/remember the reason?
https://www.pivotaltracker.com/n/projects/1566795/stories/119505309

qibobo: @julz @kanekoh  "PContext("when the scaling engine's response is too long", func()" is the pending case as it need to build a too long string to make the request failed and sometimes it panics. So it is marked as pended. Do you mean this case?

julz: I see.. yeah, seems odd to have a test that never runs, because (a) it's not actually going to fail so it doesn't stop stuff being broken (b) it will probably end up accidentally becoming out of date, like it seems to have here because at least for me it consistently fails right now. Maybe we should either find a way to make it faster so we can un-pend it (e.g. I think it should be possible to create the large string in a more efficient way, for example by pre-computing the string as a constant) or remove it, although I can create a separate bug/chore to do that, it doesn't need to be part of this PR, I was just wondering about it

PR #212, opened 2017-04-02
142916443: Enhanced Scheduler performance

kanekoh:


shakuzen:
I'm curious... Does this have a significant, measured difference in performance? It looks like a maintenance nightmare to manually manage all of the dependencies and their versions. The versions managed by Spring Boot are, to an extent, tested and known to work well together. With the POM this way, modifying any dependency is likely to break the application due to incompatible versions.

Does this really make a difference for runtime performance, or is this purely for startup performance?

kanekoh:
This is for startup performance. 
I understand what you said however I wanted to reduce 'starter dependencies' because they were too many dependencies get downloaded and most are unused ones and results in increased startup time.

kanekoh:
I changed back to starter dependencies and closing this PR.

PR #213, opened 2017-04-03
142652743 : Unit test failing for pruner sometimes

ghaih:


kanekoh:
LGTM

src/autoscaler/pruner/cmd/pruner/pruner_suite_test.go
```diff
@@ -66,16 +66,17 @@ func initConsul() {
 
 func initConfig() {
 	cfg.Logging.Level = "debug"
+	dbUrl := os.Getenv("DBURL")
```
boyang9527: fail the test if $DBURL is not set

src/autoscaler/pruner/cmd/pruner/pruner_test.go
```diff
@@ -73,9 +73,19 @@ var _ = Describe("Pruner", func() {
 				}))
 		})
 
-		It("should start", func() {
```
boyang9527: if you separate into multiple test cases, then you don't need duplicate test of 
Eventually(runner.Session.Buffer, 2*time.Second).Should(Say("pruner.started")) Consistently(runner.Session).ShouldNot(Exit())

You can have another test case to test for "prunder started"

src/autoscaler/pruner/cmd/pruner/pruner_suite_test.go
ghaih: It was moved as a part of refactoring as the same code was repeated. The sqldb_suite_test, covers the $DBURL not set case. Am I missing something. As os.Getenv("DBURL") is being used all over the place.

src/autoscaler/pruner/cmd/pruner/pruner_test.go
ghaih: Yes can separate test for .Should(Say("pruner.started")) but still need eachConsistently(runner.Session).ShouldNot(Exit()) in each test isn't it

src/autoscaler/pruner/cmd/pruner/pruner_suite_test.go
boyang9527: Perhaps I missed something... Pruner test cases can be run separately or before db package test cases. 

we can just add check before the following line and fail the test of $DBURL is not set
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/develop/src/autoscaler/pruner/cmd/pruner/pruner_suite_test.go#L70

src/autoscaler/pruner/cmd/pruner/pruner_test.go
boyang9527: yes, you are right

src/autoscaler/pruner/cmd/pruner/pruner_suite_test.go
ghaih: Fixed for pruner, just a note, similar usage in other components.

src/autoscaler/pruner/cmd/pruner/pruner_test.go
ghaih: Fixed

boyang9527: thanks!

boyang9527:
LGTM

PR #214, opened 2017-04-03
Scheduler should accept a guid from api server request to mark recurr…

qibobo:
…ing and specific_date schedules.

[#142233495]


api/test/unit/routes/policies.test.js
```diff
@@ -31,7 +31,7 @@ describe('Routing Policy Creation', function() {
 
   it('should create a policy for app id 12345', function(done) {
     nock(schedulerURI)
-    .put('/v2/schedules/12345')
+    .put(/\/v2\/schedules\/12345\?guid=.*/)
```
kanekoh: I just want to understand why you didn't use query function like following code?

```
  nock(schedulerURI)
    .put('/v2/schedules/12345')
    .query({guid: '.*'})
```

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/ScheduleEntity.java
```diff
@@ -150,12 +163,19 @@ public boolean equals(Object o) {
 		if (defaultInstanceMaxCount != null ? !defaultInstanceMaxCount.equals(that.defaultInstanceMaxCount)
 				: that.defaultInstanceMaxCount != null)
 			return false;
-		if (!instanceMinCount.equals(that.instanceMinCount))
+		if (instanceMinCount != null ? !instanceMinCount.equals(that.instanceMinCount)
+				: that.instanceMinCount != null)
+			return false;
+		if (instanceMaxCount != null ? !instanceMaxCount.equals(that.instanceMaxCount)
+				: that.instanceMaxCount != null)
+			return false;
```
ghaih: instanceMinCount and instanceMaxCount are required fields so no need to do null checks, It is done for initialMinInstanceCount because initialMinInstanceCount is not required property.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController_CreateScheduleAndNofifyScalingEngineTest.java
```diff
@@ -182,7 +184,7 @@ public void testDeleteSchedule() throws Exception {
 
 		// Delete End job.
 		ResultActions resultActions = mockMvc
-				.perform(delete(getSchedulerPath(appId)).accept(MediaType.APPLICATION_JSON));
+				.perform(delete(TestDataSetupHelper.getSchedulerPath(appId, guid)).accept(MediaType.APPLICATION_JSON));
```
kanekoh: Same question as policies.test.js in apiserver, MockMvc has param method specifying a query parameter to the request.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/ScheduleEntity.java
```diff
@@ -150,12 +163,19 @@ public boolean equals(Object o) {
...
+		if (instanceMaxCount != null ? !instanceMaxCount.equals(that.instanceMaxCount)
+				: that.instanceMaxCount != null)
+			return false;
+		if (initialMinInstanceCount != null ? !initialMinInstanceCount.equals(that.initialMinInstanceCount)
+				: that.initialMinInstanceCount != null)
 			return false;
-		if (!instanceMaxCount.equals(that.instanceMaxCount))
+		if (guid != null ? !guid.equals(that.guid)
+				: that.guid != null)
```
ghaih: guid is required property, doesnot require null check.

api/test/unit/utils/schedulerUtils.test.js
```diff
@@ -31,11 +32,12 @@ describe('Scheduler Utility functions', function() {
...
 	  it('should create new schedules for app id 12345',function(done){
 		    nock(schedulerURI)
-		    .put('/v2/schedules/12345')
+		    .put(/\/v2\/schedules\/12345\?guid=.*/)
 		    .reply(204);
 		    var mockRequest = {
 		        body : fakePolicy,
-		        params : { 'app_id' : '12345' }
+		        params : { 'app_id' : '12345' },
+		        query : {'policy_guid' : uuidV4() }
```
kanekoh: I was wondering if you specify fixed value to 'policy_guid', you could specify fixed value to 'guid' at line 35 same as 'app_id'.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
```diff
@@ -81,6 +81,8 @@
 	private MockMvc mockMvc;
 
 	private String appId = TestDataSetupHelper.generateAppIds(1)[0];
+	private String guid = TestDataSetupHelper.generateGuid();
+	private String guid1, guid2, guid3;
```
ghaih: The guid 's can be defined in method's scope.I guess the existing usage of appId here can also be refactored. Also can live with single guid variable.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
```diff
@@ -282,6 +291,21 @@ public void testCreateSchedule_without_appId() throws Exception {
...
+
+		String appId = "appId";
+		ObjectMapper mapper = new ObjectMapper();
+		ApplicationSchedules applicationPolicy = TestDataSetupHelper.generateApplicationPolicy(1, 0);
+		String content = mapper.writeValueAsString(applicationPolicy);
+
+		ResultActions resultActions = mockMvc
+				.perform(put("/v2/schedules/" + appId).contentType(MediaType.APPLICATION_JSON).content(content));
+
+		resultActions.andExpect(status().isInternalServerError());
```
kanekoh: The request without guid should return 404 or 400 because this is not internal server error for scheduler.

api/test/unit/routes/policies.test.js
qibobo: Either way could work, do we have to use the query method?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
qibobo: yes. The 500 interval error is reported by the scheduler as there is no "guid".  How to control the return http status code?

api/test/unit/routes/policies.test.js
ghaih: Since it is query string so would be neater, in my opinion

qibobo:
update the test cases.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerTest.java
kanekoh: Just notify, you can add param argument into @RequestMapping like below code. I haven't verified however it might help to solve this situation.

```
@RequestMapping(method = RequestMethod.PUT, params={"guid"})
```

qibobo: The code have changed to use ExceptionHandler.

qibobo:
@julz Any  more comments?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
```diff
@@ -122,4 +127,8 @@
 		return new ResponseEntity<>(null, null, HttpStatus.NO_CONTENT);
 	}
 
+	@ExceptionHandler(MissingServletRequestParameterException.class)
```
kanekoh: Sorry for late reply, I missed your comment.
Scheduler has a ExceptionHandler class, you can put this method into ControllerExceptionHandler class.

scheduler/db/scheduler.changelog-master.yaml
```diff
@@ -134,6 +139,11 @@ databaseChangeLog:
                   type: integer
                   constraints:
                     nullable: true
+              - column:
+                  name: guid
+                  type: varchar(50)
+                  constraints:
+                    nullable: false
```
pradyutsarma: Can we add these changes to a new changeset? Since a few of us have deployed a version internally and db reverts are not that easy anymore.

qibobo: @pradyutsarma  Did as you suggested in my previous pr "add guid to api server". When Michael merged the pr, he did what I did now. So follow his pattern here.

pradyutsarma: Yep. I'm aware of that. But that was the case when we didn't have any deployment and the idea was to not create any changesets till we have done the first release. But recently we have already deployed a version internally and hence would like to freeze on the current changelog as is and have new changesets for db updates.
@julz your comments?

boyang9527: I guess this can be postponed... while at some point, we will need to have  a 'compact' before releasing out

julz: I feel like we should get in the habit of using change sets even if we haven't officially released yet -- that way we can shake out any issues while we still have an escape hatch if we need it

kanekoh:
LGTM other than having a change-set. Are we planning to have change set for guid change?

qibobo:
@julz 	update scheduler's database change log

ghaih:
LGTM

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ControllerExceptionHandler.java
```diff
@@ -50,4 +52,8 @@
 		List<String> errors = validationErrorResult.getAllErrorMessages();
 		return new ResponseEntity<>(errors, null, HttpStatus.INTERNAL_SERVER_ERROR);
 	}
+	@ExceptionHandler(MissingServletRequestParameterException.class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
```
julz: looks like formatting might be slightly off here?

qibobo: @julz  formatted this part.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestController.java
```diff
@@ -67,13 +68,14 @@
 			@ApiResponse(code = 400, message = "Validation error encountered.") })
 	public ResponseEntity<List<String>> createSchedules(
```
julz: looks like this is mis-formatted too..

julz:
this is possibly a dumb question, but it looks like the guid is now a required parameter.. do all the existing calls supply guids? is it valid to always expect a guid to be provided for every call? for example will this break existing clients or installations like the SAP beta installation?

qibobo:
@julz  the guid is for helping synchronize data between api-server's table policy_json and scheduler's app_scaling_specific_date_schedule&app_scaling_recurring_schedule. So only the method "createSchedules" needs it.

julz:
gotcha, ok, LGTM

PR #215, opened 2017-04-04
142569467: Scaling engine needs to register with consul

kanekoh:


src/autoscaler/scalingengine/cmd/scalingengine/scalingengine_test.go
```diff
@@ -88,7 +95,69 @@ var _ = Describe("Main", func() {
...
+			consulClient = consulRunner.NewClient()
+			runner.startCheck = ""
+		})
+
+		JustBeforeEach(func() {
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say("scalingengine.registration-runner.succeeded-registering-service"))
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say("scalingengine.started"))
+		})
+
+		It("should get scaling engine sservice", func() {
```
ghaih: Typo in there for service

src/autoscaler/scalingengine/cmd/scalingengine/scalingengine_test.go
```diff
@@ -88,7 +95,69 @@ var _ = Describe("Main", func() {
...
+		})
+
+		JustBeforeEach(func() {
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say("scalingengine.registration-runner.succeeded-registering-service"))
+			Eventually(runner.Session.Buffer, 2*time.Second).Should(gbytes.Say("scalingengine.started"))
+
+			runner.Interrupt()
+		})
+
+		It("should not get scaling engine sservice", func() {
```
ghaih: Typo for service.

ghaih:
LGTM, other than the couple of typos as mentioned.

rohitsharma04:
LGTM

julz:
travis tests failed, kicked off again

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -96,7 +96,7 @@ var _ = Describe("Integration_Scheduler_ScalingEngine", func() {
 
 			Eventually(func() bool {
 				return activeScheduleExists(testAppId)
-			}, 2*time.Minute).Should(BeFalse())
+			}, 2*time.Minute, 5*time.Second).Should(BeFalse())
```
julz: out of interest, why 5 seconds rather than 1 second here? seems like some tests changed the interval to 5s, some to 1s.. why? If there's a reason a 1s interval fails here it'd be interesting to understand why? Ideally if we can use 1 second everywhere the tests should be much faster than using 5s.

julz: I see, is it worrying for real usage that the difference between polling every 5 seconds vs every second can make such a big difference? I guess what I'm asking is if polling every 1 second causes enough cpu load to affect other tests that seems like it could indicate an actual performance problem?

kanekoh: In an effort to speed up scheduler I had changed to 5. I will make it 1

julz: ok cool, let's keep it at 1s and if we see performance problems at that rate maybe we should see if we can optimise the code somehow

PR #216, opened 2017-04-06
[scheduler] Move test properties to application.properties

shakuzen:
Previously the test configuration properties were in a `@TestPropertySource` declaration on
a class (`TestConfiguration`) extended by all of the `@SpringBootTest` classes. This is
effectively the same thing as keeping all the properties in `src/test/resources/application.properties`,
which eliminates the need to extend a dummy class from each class annotated with `@SpringBootTest`.

Properties have been sorted in alphabetical order by key, which should help easily locate properties.


ghaih:
Just want to mention, did not maintain a separate properties file for tests as an outcome of review.

ghaih:
LGTM

kanekoh:
LGTM

PR #217, opened 2017-04-07
#143303953: Removed unneeded port from eventgenerator

rohitsharma04:
As event generator is daemon like pruner, it doesn't require a port.


ghaih:
LGTM

src/integration/integration_suite_test.go
```diff
@@ -172,7 +172,6 @@ func PreparePorts() Ports {
 		ServiceBroker:    11000 + GinkgoParallelNode(),
 		Scheduler:        12000 + GinkgoParallelNode(),
 		MetricsCollector: 13000 + GinkgoParallelNode(),
-		EventGenerator:   14000 + GinkgoParallelNode(),
```
kanekoh: LGTM, Just for maintaining sequence can reduce the port numbers for ScalingEngine(e.g. change to 14000), ConsulCluster.

rohitsharma04: Done.

PR #220, opened 2017-04-19
143386683: Allowed test cases of metrics collector package to be run in parallel

kanekoh:


src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
```diff
@@ -42,31 +43,24 @@ var (
...
-			StartingPort: 9001 + GinkgoParallelNode()*consulrunner.PortOffsetLength,
-			NumNodes:     1,
-			Scheme:       "http",
-		},
-	)
-	consulRunner.Start()
-	consulRunner.WaitUntilReady()
-	mcPath = string(pathsByte)
-
 	ccNOAAUAA = ghttp.NewServer()
```
boyang9527: why setting up the ccNOAAUAA in primary node?  We can do it in each node, and have different app IDs for different node so that  test cases do not share any ccNoaaUAA server and data in metricscollector database

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
```diff
@@ -109,8 +103,55 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
+
+	data, err := json.Marshal(&setupParams)
+	Expect(err).NotTo(HaveOccurred())
+
+	return data
+}, func(data []byte) {
+	err := json.Unmarshal(data, &setupParams)
+	Expect(err).NotTo(HaveOccurred())
+	mcPath = setupParams.McBuildPath
+	eLock = setupParams.Lock
```
boyang9527: this will not share the lock status among nodes, but create a new Lock.

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_suite_test.go
kanekoh: Metrics collector get application metrics from cf with app_ids which are in policy_json database.
If ccNOAAUAA set up in each node, it could receive all appIds which are in policy_json database, then return un-handled request error.

boyang9527: I missed the "loading all app IDs" part of metrics collector.  While we can setup database in primary node with only one app in policy_json and each node has its own ccNOAAUAA sever.   The main reason is my comments below on the eLock, it can not be shared among nodes by marshall and unmarshall.

kanekoh: OK, Will fix it.

boyang9527:
LGTM

ghaih:
LGTM

PR #225, opened 2017-05-02
make consul to be configurable for metrics collector

boyang9527:
[#144184433]

### DIFF
--- a/src/autoscaler/metricscollector/cmd/metricscollector/main.go
+++ b/src/autoscaler/metricscollector/cmd/metricscollector/main.go
@@ -113,27 +113,28 @@ func main() {
 		os.Exit(1)
 	}
 
-	consulClient, err := consuladapter.NewClientFromUrl(conf.Lock.ConsulClusterConfig)
-	if err != nil {
-		logger.Fatal("new consul client failed", err)
-	}
-
-	serviceClient := metricscollector.NewServiceClient(consulClient, mcClock)
-
-	lockMaintainer := serviceClient.NewMetricsCollectorLockRunner(
-		logger,
-		generateGUID(logger),
-		conf.Lock.LockRetryInterval,
-		conf.Lock.LockTTL,
-	)
-
-	registrationRunner := initializeRegistrationRunner(logger, consulClient, conf.Server.Port, mcClock)
-
 	members := grouper.Members{
-		{"lock-maintainer", lockMaintainer},
 		{"collector", collectServer},
 		{"http_server", httpServer},
-		{"registration", registrationRunner},
+	}
+
+	if conf.Lock.ConsulClusterConfig != "" {
+		consulClient, err := consuladapter.NewClientFromUrl(conf.Lock.ConsulClusterConfig)
+		if err != nil {
+			logger.Fatal("new consul client failed", err)
+		}
+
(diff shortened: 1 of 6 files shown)


src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -113,27 +113,28 @@ func main() {
...
 
-	registrationRunner := initializeRegistrationRunner(logger, consulClient, conf.Server.Port, mcClock)
-
-	members := grouper.Members{
-		{"lock-maintainer", lockMaintainer},
-		{"collector", collectServer},
-		{"http_server", httpServer},
-		{"registration", registrationRunner},
+		registrationRunner := initializeRegistrationRunner(logger, consulClient, conf.Server.Port, mcClock)
+		members = append(members, grouper.Member{"lock-maintainer", lockMaintainer})
```
rohitsharma04: Lock maintainer should be the first one in the members otherwise collector and http_server would start even when the lock is not achieved.

rohitsharma04: We should also add test to verify lock runner is the first one to run in the case consul is present. We missed this earlier.

boyang9527: right... let me fix it

boyang9527:
@rohitsharma04  committed the fixes.  By the way,  since http_sever is ifritized, it is not able to detect the started log. While it is still good to be in "stand-by" as long as the collector is not started and the instance is not registered.

src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -107,18 +107,20 @@ func main() {
 		return nil
 	})
 
-	httpServer, err := server.NewServer(logger, conf, cfClient, noaa, instanceMetricsDB)
+	httpServer, err := server.NewServer(logger.Session("http_sever"), conf, cfClient, noaa, instanceMetricsDB)
```
rohitsharma04: typo

rohitsharma04:
@boyang9527 agree,
LGTM other than the typo.

src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -107,33 +107,40 @@ func main() {
...
+		serviceClient := metricscollector.NewServiceClient(consulClient, mcClock)
+		lockMaintainer := serviceClient.NewMetricsCollectorLockRunner(
+			logger,
+			generateGUID(logger),
+			conf.Lock.LockRetryInterval,
+			conf.Lock.LockTTL,
+		)
+
+		registrationRunner := initializeRegistrationRunner(logger, consulClient, conf.Server.Port, mcClock)
+		members = grouper.Members{
```
paltanmoy: It would be better to add the common runner i.e http_server and collector once and prepend lock_maintainer and apend registrartion runner for consul

boyang9527: sure if you think it is more readable...

boyang9527:
@paltanmoy  @rohitsharma04 
just had another commit to address your comments to fix the typo and simplify the ifrit members code.

paltanmoy:
LGTM

PR #226, opened 2017-05-03
144184371: As an operator, I would like to not need consul to run the scheduler

kanekoh:


scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/misc/DisableConsulTest.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	private void assertNotRegisterWithConsul() {
+		ConsulClient consulClient = new ConsulClient();
+
+		Response<Map<String, Service>> services = consulClient.getAgentServices();
+		Service service = services.getValue().get("scheduler");
+		assertNull(service);
+	}
+}
```
boyang9527: may need to add a test case to make sure the schedule has started even consul is not enabled.

kanekoh:
@boyang9527, Thanks, I added a test case scheduler end-to-end test case in that class.

boyang9527:
LGTM

PR #228, opened 2017-05-05
[#144184583] As an operator, I would like to not need consul to run the pruner

paltanmoy:


boyang9527:
LGTM

rohitsharma04:
LGTM

src/autoscaler/pruner/config/config_test.go
```diff
@@ -96,10 +96,6 @@ scaling_engine_db:
   db_url: "postgres://postgres:postgres@localhost/autoscaler?sslmode=disable"
   refresh_interval: 36h
   cutoff_days: 30
-lock:
-  lock_ttl: 15s
-  lock_retry_interval: 10s
-  consul_cluster_config: "http://127.0.0.1:8500"
 `)
```
kanekoh: It is better to leave config as it is and not remove consul config here.

src/autoscaler/pruner/config/config_test.go
```diff
@@ -120,10 +116,6 @@ lock:
 				Expect(conf.ScalingEngineDb.RefreshInterval).To(Equal(36 * time.Hour))
 				Expect(conf.ScalingEngineDb.CutoffDays).To(Equal(30))
 
-				Expect(conf.Lock.LockTTL).To(Equal(15 * time.Second))
-				Expect(conf.Lock.LockRetryInterval).To(Equal(10 * time.Second))
-				Expect(conf.Lock.ConsulClusterConfig).To(Equal("http://127.0.0.1:8500"))
-
```
kanekoh: Leave it as it is

src/autoscaler/pruner/config/config_test.go
```diff
@@ -380,8 +372,8 @@ lock:
 				conf.Lock.ConsulClusterConfig = ""
 			})
 
-			It("should error", func() {
-				Expect(err).To(MatchError(MatchRegexp("Configuration error: Consul Cluster Config is empty")))
+			It("should validate successfully", func() {
+				Expect(err).NotTo(HaveOccurred())
```
kanekoh: Since the partial config test will take care of the case "no consul config" so we can remove this test completely.

paltanmoy: @kanekoh This test validates the config without consul-cluster-config section.partial config test is basically testing whether default values for different properties are correctly applied with the partial (mandatory) properties or not. IMHO we should keep this.

src/autoscaler/pruner/config/config_test.go
paltanmoy: Done

src/autoscaler/pruner/config/config_test.go
kanekoh: My point was that we don't need a test for optional parameters. And partial test is passing when consul cluster not specified, so it already validates what we are doing through this test case.

rohitsharma04: partial test is only loading the default config and not validating it.

paltanmoy: @kanekoh  We need test case to check whether config is valid even without consul_cluster_config. As partial test does not validate the config (only loads with default values), we need this test.

kanekoh: I see.

PR #229, opened 2017-05-09
Scheduler should provide a rest api to sync up with api server policy db

qibobo:
[#128924335]

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
@@ -21,7 +21,6 @@ public class PolicyJsonDaoImpl extends JdbcDaoSupport implements PolicyJsonDao {
 	@Override
 	@Transactional(readOnly = true)
 	public List<PolicyJsonEntity> getAllPolicies() {
-		// TODO Auto-generated method stub
 		List<PolicyJsonEntity> list = getJdbcTemplate().query("SELECT * FROM policy_json", new PolicyJsonEntity());
 		return list;
 	}
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
@@ -13,6 +13,7 @@ import org.springframework.transaction.annotation.Transactional;
 @Repository("policyJsonDao")
 public class PolicyJsonDaoImpl extends JdbcDaoSupport implements PolicyJsonDao {
 
+	final String SQL_GET_All_POLICYJSON = "SELECT * FROM policy_json";
 	@Resource(name = "policyDbDataSource")
 	private void setupPolicyDataSource(DataSource policyDbDataSource) {
 		setDataSource(policyDbDataSource);
@@ -21,8 +22,8 @@ public class PolicyJsonDaoImpl extends JdbcDaoSupport implements PolicyJsonDao {
 	@Override
 	@Transactional(readOnly = true)
 	public List<PolicyJsonEntity> getAllPolicies() {
-		List<PolicyJsonEntity> list = getJdbcTemplate().query("SELECT * FROM policy_json", new PolicyJsonEntity());
-		return list;
+			return getJdbcTemplate().query(SQL_GET_All_POLICYJSON, new PolicyJsonEntity());
+		
 	}
 
 }
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
@@ -1,0 +1,29 @@
+package org.cloudfoundry.autoscaler.scheduler.dao;
+
+import java.util.List;
+
+import javax.annotation.Resource;
+import javax.sql.DataSource;
+
+import org.cloudfoundry.autoscaler.scheduler.entity.PolicyJsonEntity;
+import org.springframework.jdbc.core.support.JdbcDaoSupport;
+import org.springframework.stereotype.Repository;
+import org.springframework.transaction.annotation.Transactional;
+
+@Repository("policyJsonDao")
+public class PolicyJsonDaoImpl extends JdbcDaoSupport implements PolicyJsonDao {
+
+	@Resource(name = "policyDbDataSource")
+	private void setupPolicyDataSource(DataSource policyDbDataSource) {
+		setDataSource(policyDbDataSource);
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public List<PolicyJsonEntity> getAllPolicies() {
+		// TODO Auto-generated method stub
+		List<PolicyJsonEntity> list = getJdbcTemplate().query("SELECT * FROM policy_json", new PolicyJsonEntity());
+		return list;
+	}
+
+}
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/PolicyJsonEntity.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/PolicyJsonEntity.java
@@ -1,23 +1,15 @@
 package org.cloudfoundry.autoscaler.scheduler.entity;
 
-import java.io.IOException;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.List;
 
 import javax.persistence.Column;
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.Table;
-import javax.persistence.Transient;
 import javax.validation.constraints.NotNull;
 
-import org.cloudfoundry.autoscaler.scheduler.rest.model.ApplicationSchedules;
-import org.cloudfoundry.autoscaler.scheduler.rest.model.Schedules;
 import org.springframework.jdbc.core.RowMapper;
-
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
 
 @Entity
 @Table(name = "policy_json")
@@ -36,8 +28,6 @@ public class PolicyJsonEntity implements RowMapper<PolicyJsonEntity> {
 	@Column(name = "guid")
 	private String guid;
 
-	@Transient
-	private ApplicationSchedules schedules;
 
 	public String getAppId() {
 		return appId;
@@ -55,14 +45,6 @@ public class PolicyJsonEntity implements RowMapper<PolicyJsonEntity> {
 		this.policyJson = policyJson;
 	}
 
-	public ApplicationSchedules getSchedules() {
-		return schedules;
-	}
-
-	public void setSchedules(ApplicationSchedules schedules) {
-		this.schedules = schedules;
-	}
-
 	public String getGuid() {
 		return guid;
 	}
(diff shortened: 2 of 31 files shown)


qibobo:
@kanekoh  For policy, we need to read the all the policy content and get the relevant schedules. For schedules, currently all of the schedules are operated(create,delete & update) by entity, in my idea, we can follow this pattern until we unify the way of db access.

kanekoh:
@qibobo, it's better to have app id and guid in map object instead of whole schedules. Schedules which is from Schedule DB are unused objects. Create/Update schedule needs whole schedules however it's from Policy DB.

qibobo:
@kanekoh refreshed the code as your comments.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/dao/RecurringScheduleDaoImplTest.java
```diff
@@ -55,13 +59,16 @@ public void before() throws SQLException {
 		// Add fake test records.
 		String appId = "appId1";
 		String guid = TestDataSetupHelper.generateGuid();
-		List<RecurringScheduleEntity> entities = TestDataSetupHelper.generateRecurringScheduleEntities(appId, guid, 1, 0);
+		List<RecurringScheduleEntity> entities = TestDataSetupHelper.generateRecurringScheduleEntities(appId, guid,false, 1,
```
kanekoh: You can use RecurringScheduleEntitiesBuilder  class to create entities instead of TestDataSetupHelper class.
entities = RecurringScheduleEntitiesBuilder builder = new RecurringScheduleEntitiesBuilder(1, 0).setAppId(appId).setGuid(guid).setTimeZone(timeZone).setDefaultInstanceMinCount(1).setDefaultInstanceMaxCount(5).build();

Because we plan to get rid of TestDataSetupHelper later.

qibobo:
@kanekoh  refactor the code

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleSyncRestController.java
```diff
@@ -0,0 +1,38 @@
...
+	@RequestMapping(method=RequestMethod.PUT)
+	@ResponseStatus(HttpStatus.OK)
+	public ResponseEntity<SynchronizeResult> synchronizeSchedules(){
+		try{
+			SynchronizeResult result = scheduleManager.synchronizeSchedules();
+			return new ResponseEntity<>(result, null, HttpStatus.OK);
+		}catch(Exception e){
+			e.printStackTrace();
+			logger.error(e.getMessage());
+			logger.error(e.getStackTrace());
```
qibobo: @paltanmoy  your idea?

paltanmoy: @qibobo you can simply use logger.error("some custom msg",e)
Also it's advisable not to use e.printStackTrace() in production ready code.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -748,7 +757,99 @@ private void deleteActiveSchedules(String appId) {
...
+		}
+		
+		List<PolicyJsonEntity> toCreatePolicyList = new ArrayList<PolicyJsonEntity>();
+		Set<String> toDeletedAppIds = new HashSet<String>();
+		for (String appIdInPolicy : policyMap.keySet()) {
+			if (policyMap.get(appIdInPolicy).getSchedules() != null
+					&& policyMap.get(appIdInPolicy).getSchedules().getSchedules() != null) {
+				if (policyMap.get(appIdInPolicy).getSchedules().getSchedules().getRecurringSchedule() != null
+						&& policyMap.get(appIdInPolicy).getSchedules().getSchedules().getRecurringSchedule()
+								.size() > 0) {
```
kanekoh: Can add hasSpecifuDateSchedule and hasRecurringSchedule in Schedules model class ? and then refactor hasSchedules to use these methods.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleSyncRestControllerTest.java
```diff
@@ -0,0 +1,150 @@
...
+		mockMvc = MockMvcBuilders.webAppContextSetup(wac).build();
+
+		ActiveScheduleEntity activeScheduleEntity = new ActiveScheduleEntity();
+		activeScheduleEntity.setStartJobIdentifier(System.currentTimeMillis());
+		Mockito.when(activeScheduleDao.find(any())).thenReturn(activeScheduleEntity);
+
+	}
+
+	@Test
+	public void testSynchronizeSchedules_with_both_policy_and_schedules_existed() throws Exception {
```
kanekoh: I wonder we need two types of test cases for Update pattern, one for having same guid, one for having different guid.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/PolicyJsonEntityBuilder.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.cloudfoundry.autoscaler.scheduler.entity.PolicyJsonEntity;
+import org.cloudfoundry.autoscaler.scheduler.rest.model.ApplicationSchedules;
+import org.cloudfoundry.autoscaler.scheduler.rest.model.Schedules;
+
+public class PolicyJsonEntityBuilder {
+	
+	private PolicyJsonEntity policyJson;
+	
+	public PolicyJsonEntityBuilder(){
```
kanekoh: Is this constructor needed?

src/integration/integration_scheduler_scalingengine_test.go
```diff
@@ -101,6 +104,107 @@ var _ = Describe("Integration_Scheduler_ScalingEngine", func() {
...
+			BeforeEach(func() {
+				policyStr = string(setPolicyDateTime(readPolicyFromFile("fakePolicyWithSpecificDateSchedule.json")))
+				resp, err := createSchedule(testAppId, testGuid, policyStr)
+				checkResponseIsEmpty(resp, err, http.StatusOK)
+
+				Eventually(func() bool {
+					return activeScheduleExists(testAppId)
+				}, 2*time.Minute, 5*time.Second).Should(BeTrue())
+
+				deletePolicy(testAppId)
```
qibobo: Another method call insertPolicy with the parameter of models.ScalingPolicy. Here I only need to insert with a string._

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
qibobo: I prefer we add another story to do the refactoring.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/PolicyJsonEntity.java
```diff
@@ -0,0 +1,119 @@
...
+	@Override
+	public PolicyJsonEntity mapRow(ResultSet rs, int rowNum) throws SQLException {
+		// TODO Auto-generated method stub
+		PolicyJsonEntity policyJsonEntity = new PolicyJsonEntity();
+		policyJsonEntity.setAppId(rs.getString("app_id"));
+		policyJsonEntity.setPolicyJson(rs.getString("policy_json"));
+		policyJsonEntity.setGuid(rs.getString("guid"));
+		ObjectMapper mapper = new ObjectMapper();
+		//do not mapper scaling_rules in policy_json
+		mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
qibobo: As in scheduler we only care the schedule part of the policy_json, I still prefer we do the mapping here.

src/integration/integration_scheduler_scalingengine_test.go
ghaih: Ok yes saw there is another method insertPolicy , but now two methods  executing same query, please change the invoker of insertPolicy to pass the guid "1234" as aparameter and get rid of insertPolicyDirectly.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/PolicyJsonEntity.java
ghaih: Point made was, it should not be part of entity.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
ghaih: Since this piece of code was part of this PR so made this comment. Decision on that can be made if there is a consensus. Your thoughts on this @boyang9527, @paltanmoy

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/entity/PolicyJsonEntity.java
qibobo: ok.

qibobo:
@kanekoh  updated

src/integration/integration_suite_test.go
```diff
@@ -391,15 +398,27 @@ func clearDatabase() {
 	Expect(err).NotTo(HaveOccurred())
 }
 
-func insertPolicy(appId string, scalingPolicy models.ScalingPolicy) {
+// func insertPolicy-ss(appId string, scalingPolicy models.ScalingPolicy) {
```
paltanmoy: @qibobo  Just wanted to understand the reason behind keeping this piece of commented code

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -789,37 +795,32 @@ public SynchronizeResult synchronizeSchedules(){
...
 						toDeletedAppIds.add(appIdInPolicy);
 						updateCount++;
 						continue;
 					}
 				}
 				
-				if (policyMap.get(appIdInPolicy).getSchedules().getSchedules().getSpecificDate() != null
-						&& policyMap.get(appIdInPolicy).getSchedules().getSchedules().getSpecificDate()
-								.size() > 0) {
+				if (policySchedulesMap.get(appIdInPolicy).getSchedules().hasSpecificDateSchedule()) {
```
ghaih: We don't need to process two maps, appIdAndGuidMap already has data from specific and recurring schedules. It is enough to do (policySchedulesMap.get(appIdInPolicy).getSchedules().hasSchedules

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -842,14 +843,52 @@ public SynchronizeResult synchronizeSchedules(){
...
+						specificDate.setDefaultInstanceMinCount(applicationSchedules.getInstanceMinCount());
+						specificDate.setDefaultInstanceMaxCount(applicationSchedules.getInstanceMaxCount());
+						specificDate.setGuid(policyJsonEntity.getGuid());
+					}
+				}
+				 
+			}
+			
+		} catch (IOException e) {
+			e.printStackTrace();
```
ghaih: Still printStackTrace is present. Please check other spots.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManagerTest.java
```diff
@@ -838,20 +847,29 @@ public void testSynchronizeSchedules_with_both_policy_and_schedules_existed_and_
...
-
+		
 		if (schedules.getSpecificDate() != null) {
+			Mockito.verify(specificDateScheduleDao, Mockito.times(noOfSpecificDateSchedules)).create(specificDateCaptor.capture());
+			List<SpecificDateScheduleEntity> specificDateList = specificDateCaptor.getAllValues();
+			int i = 0;
 			for (SpecificDateScheduleEntity foundSpecificDateScheduleEntity : schedules.getSpecificDate()) {
-				Mockito.verify(specificDateScheduleDao, Mockito.times(1)).create(foundSpecificDateScheduleEntity);
+				assertEquals(foundSpecificDateScheduleEntity,specificDateList.get(i));
+				i++;
```
ghaih: Just for information, the assert could have used list contains, as we cannot assume list to be sorted.

qibobo:
@kanekoh @ghaih  any more comments?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleSyncRestController.java
```diff
@@ -27,7 +27,6 @@
 			SynchronizeResult result = scheduleManager.synchronizeSchedules();
 			return new ResponseEntity<>(result, null, HttpStatus.OK);
 		}catch(Exception e){
-			e.printStackTrace();
 			logger.error("Internal error", e.getMessage());
```
kanekoh: Please see the code as a reference, it should have exception not the message.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/PolicyUtil.java
```diff
@@ -18,7 +18,7 @@ public static String getPolicyJsonContent() {
 				jsonPolicyStr += tmp;
 			}
 		} catch (IOException e) {
-			e.printStackTrace();
+			return null;
```
kanekoh: Please throw an exception.

qibobo:
@ghaih  your comments?

ghaih:
@qibobo , no more comments, still would want an LGTM from another set of eyes. @paltanmoy , since you had a look before or anyone else.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleSyncRestController.java
```diff
@@ -0,0 +1,28 @@
...
+@RequestMapping(value = "/v2/syncSchedules")
+public class ScheduleSyncRestController {
+
+	@Autowired
+	private ScheduleManager scheduleManager;
+
+	@RequestMapping(method = RequestMethod.PUT)
+	@ResponseStatus(HttpStatus.OK)
+	public ResponseEntity<SynchronizeResult> synchronizeSchedules() {
+		SynchronizeResult result = scheduleManager.synchronizeSchedules();
```
paltanmoy: @qibobo Based on our exception in synchronizeSchedules we should delegate it to ExceptionHandler to have proper error response to the API consumer

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -748,7 +761,116 @@ private void deleteActiveSchedules(String appId) {
...
+						specificDate.setDefaultInstanceMinCount(applicationSchedules.getInstanceMinCount());
+						specificDate.setDefaultInstanceMaxCount(applicationSchedules.getInstanceMaxCount());
+						specificDate.setGuid(policyJsonEntity.getGuid());
+					}
+				}
+				 
+			}
+			
+		} catch (IOException e) {
+			applicationSchedules = null;
```
paltanmoy: In case of an exception we should at least log or throw the same to keep track of what went wrong.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/dao/PolicyJsonDaoImpl.java
```diff
@@ -0,0 +1,28 @@
...
+
+	@Resource(name = "policyDbDataSource")
+	private void setupPolicyDataSource(DataSource policyDbDataSource) {
+		setDataSource(policyDbDataSource);
+	}
+
+	@Override
+	@Transactional(readOnly = true)
+	public List<PolicyJsonEntity> getAllPolicies() {
+		List<PolicyJsonEntity> list = getJdbcTemplate().query("SELECT * FROM policy_json", new PolicyJsonEntity());
```
paltanmoy: a) If this fails while accessing the database?
b) Can we make this query as constant ?

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleSyncRestController.java
qibobo: @paltanmoy  The exceptions will be handled by @ExceptionHandler(Exception.class)

qibobo:
@paltanmoy refreshed.

qibobo:
@paltanmoy  any more?

paltanmoy:
LGTM

PR #230, opened 2017-05-09
144184523 : As an operator, I would like to not need consul to run the event generator

ghaih:


src/autoscaler/eventgenerator/config/config_test.go
```diff
@@ -525,17 +521,6 @@ lock:
 			})
 		})
 
-		Context("when consul cluster config is not set", func() {
```
rohitsharma04: The test case can be left here, It should not error.

boyang9527:
LGTM

PR #234, opened 2017-05-18
145494239 : Create basic manifest to push all the components as application without starting them

ghaih:
Push applications with --no-start.


julz:
LGTM but maybe we can document how to use this in the README.md too, e.g. a section along the lines of "Pushing to Cloud Foundry"?

manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+  instances: 1
+  host: pruner
+  path: src/autoscaler/pruner
+- name: scalingengine
+  memory: 512M
+  instances: 1
+  host: scalingengine
+  path: src/autoscaler/scalingengine
+- name: scheduler
+  memory: 512M
```
rohitsharma04: Maybe 1G memory as it's a java app.

manifest.yaml
```diff
@@ -0,0 +1,36 @@
...
+  instances: 1
+  host: apiserver
+  path: api/
+- name: servicebroker
+  memory: 512M
+  instances: 1
+  host: servicebroker
+  path: servicebroker
+- name: eventgenerator
```
rohitsharma04: Go app files are directly pushed. For automatic buildpack detection we should add GOPACKAGENAME environment variable also for dependencies we need some vendoring(maybe instructions can be added in the readme).
Same for metricscollector, scalingengine and pruner too.

pradyutsarma:
One question: Currently the database credentials and other env settings are injected via the settings/config file generated from an erb file. For a cf app, we could actually inject them directly into the env of the app via the manifest. For the database, shouldn't we have placeholders for a user-provided or a managed service?

boyang9527:
A general comment:  this should be in release project from my understanding ...

manifest.yaml
boyang9527: we can push all the code in src directory while only only install the one we need.

boyang9527:
@ghaih  how we are going to do with this PR...

ghaih:
@boyang9527 , we thought we will make it an epic and have added stories to it. https://www.pivotaltracker.com/n/projects/1566795/epics/3570041
Are you fine with this approach. We had been caught up in some other task sorry about the delay.

boyang9527:
@ghaih  I am good to have several user stories under this epic and have this PR as the starting user story

boyang9527:
LGTM

ghaih:
@rohitsharma04 , Are you fine with the approach?

rohitsharma04:
@ghaih it's fine with me too.

PR #238, opened 2017-05-23
add streaming as an option in metrics collector main function

boyang9527:
[#145902157]

### DIFF
--- a/src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
+++ b/src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
@@ -20,6 +20,7 @@ import (
 
 	"autoscaler/cf"
 	"autoscaler/metricscollector"
+	"autoscaler/metricscollector/config"
 
 	"code.cloudfoundry.org/consuladapter"
 	"github.com/onsi/gomega/gbytes"
@@ -238,7 +239,7 @@ var _ = Describe("MetricsCollector", func() {
 		})
 	})
 
-	Describe("when a request for memory metrics comes", func() {
+	Context("when a request for memory metrics comes", func() {
 		Context("when token is not expired", func() {
 			BeforeEach(func() {
 				eLock.Lock()
@@ -272,7 +273,7 @@ var _ = Describe("MetricsCollector", func() {
 
 	})
 
-	Describe("when a request for metrics history comes", func() {
+	Context("when a request for metrics history comes", func() {
 		BeforeEach(func() {
 			runner.Start()
 		})
@@ -284,4 +285,21 @@ var _ = Describe("MetricsCollector", func() {
 			rsp.Body.Close()
 		})
 	})
+
+	Context("when using streaming for metrics collection", func() {
+		BeforeEach(func() {
+			streamingCfg := cfg
+			streamingCfg.Collector.CollectMethod = config.CollectMethodStreaming
+			runner.configPath = writeConfig(&streamingCfg).Name()
+			runner.Start()
+		})
+
+		It("returns with a 200 for metric query", func() {
+			rsp, err := httpClient.Get(fmt.Sprintf("https://127.0.0.1:%d/v1/apps/an-app-id/metric_histories/a-metric-type", mcPort))
+			Expect(err).NotTo(HaveOccurred())
+			Expect(rsp.StatusCode).To(Equal(http.StatusOK))
+			rsp.Body.Close()
+		})
+	})
+
 })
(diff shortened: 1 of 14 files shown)


boyang9527:
@kanekoh @rohitsharma04  

can you please help review this PR,  this blocks acceptance test for throughput and response time based scaling

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
```diff
@@ -284,4 +285,21 @@ var _ = Describe("MetricsCollector", func() {
 			rsp.Body.Close()
 		})
 	})
+
+	Context("when using streaming for metrics collection", func() {
```
kanekoh: Maybe this should be a Describe and we need another above sections "When using polling for metrics collection". And then, a context "when a request for metrics history comes"on top of "It("returns with a 200 for metric query", func() " done below.

src/autoscaler/metricscollector/config/config.go
```diff
@@ -126,6 +132,9 @@ func (c *Config) Validate() error {
 		return fmt.Errorf("Configuration error: RefreshInterval is 0")
 	}
 
+	if (c.Collector.CollectMethod != CollectMethodPolling) && (c.Collector.CollectMethod != CollectMethodStreaming) {
```
kanekoh: if we expect more collect methods in future, then maybe can use an enum to avoid adding conditions here.

src/integration/integration_suite_test.go
```diff
@@ -451,39 +457,65 @@ func startFakeCCNOAAUAA(instanceCount int) {
...
 }
 
-func fakeMetrics(appId string, memoryValue uint64) {
+func fakeMetricsPolling(appId string, memoryValue uint64) {
 
 	eLock = &sync.Mutex{}
-	fakeCCNOAAUAA.RouteToHandler("GET", ccNOAAUAARegPath,
+	fakeCCNOAAUAA.RouteToHandler("GET", noaaPollingRegPath,
 		func(rw http.ResponseWriter, r *http.Request) {
-			eLock.Lock()
```
kanekoh: Just for our understanding, why you remove this piece of code?

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
boyang9527: We can add a describe for REST API test, and then have two contexts for polling and streaming

src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
```diff
@@ -284,4 +285,21 @@ var _ = Describe("MetricsCollector", func() {
...
+		BeforeEach(func() {
+			streamingCfg := cfg
+			streamingCfg.Collector.CollectMethod = config.CollectMethodStreaming
+			runner.configPath = writeConfig(&streamingCfg).Name()
```
boyang9527: added.

src/autoscaler/metricscollector/config/config.go
boyang9527: We may add firehose in the future...
but I guess we are not going to provide multiple ways .. So far my thinking is they are just for internal use only since we are not sure which is the best way given we have not done performance test yet... 

Polling will eventually be dropped since it can only provide memory ( and cpu ) support.

boyang9527: Given this, I will not use enum

src/integration/integration_suite_test.go
boyang9527: this piece of code is used to test token expire case, we have done it in unit test.. we don't need to test every cases in  integration test ...

boyang9527:
@kanekoh and @ghaih,  websocket_handler.go is a test helper to handle the web socket request and write messages back to the WS client. 

It is a piece of code copied from github.com/cloudfoundry/noaa test package ( we can not import )

boyang9527:
@kanekoh and @ghaih, refactored the test cases. Please check

rohitsharma04:
LGTM

src/autoscaler/metricscollector/config/config.go
ghaih: Yes with one option, enum doesn't make sense. Got it.

src/integration/integration_suite_test.go
ghaih: Ok Thanks.

boyang9527:
@ghaih  rebased from "develop" branch.. while the scheduler-scalingengine integration test of "develop" branch are failing, so it can not pass travis CI. 

Will need someone to fix the defect in "develop" branch first.

boyang9527:
Here the defect 
https://www.pivotaltracker.com/n/projects/1566795/stories/146331975

boyang9527:
@ghaih  It passed the travis CI, while I still have the scheduler-scalingengine integration test failure sometimes. Let us track this in the defect I created.  This has nothing to do with this PR.

PR #239, opened 2017-05-24
change memoryused to memoryutil

boyang9527:
[#144957125]

### DIFF
--- a/api/lib/validation/schemaValidator.js
+++ b/api/lib/validation/schemaValidator.js
@@ -57,7 +57,7 @@ var getDaysInMonthInISOFormat = function() {
 };
 
 var getMetricTypes = function() {
-  var metricTypeEnum = ['memoryused'];
+  var metricTypeEnum = ['memoryutil', 'responsetime', 'throughput'];
   return metricTypeEnum;
 };
 
--- a/src/autoscaler/scalingengine/scalingengine_test.go
+++ b/src/autoscaler/scalingengine/scalingengine_test.go
@@ -56,7 +56,7 @@ var _ = Describe("ScalingEngine", func() {
 				MetricType:            models.MetricNameMemory,
 				BreachDurationSeconds: 100,
 				CoolDownSeconds:       30,
-				Threshold:             222,
+				Threshold:             80,
 				Operator:              ">",
 				Adjustment:            "+1",
 			}
@@ -92,7 +92,7 @@ var _ = Describe("ScalingEngine", func() {
 					Status:       models.ScalingStatusSucceeded,
 					OldInstances: 2,
 					NewInstances: 3,
-					Reason:       "+1 instance(s) because memoryused > 222 for 100 seconds",
+					Reason:       "+1 instance(s) because memoryutil > 80 for 100 seconds",
 				}))
 
 			})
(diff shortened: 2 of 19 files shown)


src/autoscaler/scalingengine/scalingengine_test.go
```diff
@@ -92,7 +92,7 @@ var _ = Describe("ScalingEngine", func() {
 					Status:       models.ScalingStatusSucceeded,
 					OldInstances: 2,
 					NewInstances: 3,
-					Reason:       "+1 instance(s) because memoryused > 222 for 100 seconds",
+					Reason:       "+1 instance(s) because memoryutil > 80 for 100 seconds",
```
rohitsharma04: we can add percentage in the message

> +1 instance(s) because memoryutil > 80 percentage for 100 seconds

api/lib/validation/schemaValidator.js
```diff
@@ -57,7 +57,7 @@ var getDaysInMonthInISOFormat = function() {
 };
 
 var getMetricTypes = function() {
-  var metricTypeEnum = ['memoryused'];
+  var metricTypeEnum = ['memoryutil', 'responsetime', 'throughput'];
```
rohitsharma04: Add the upper limit for the threshold value validation, currently there's no upper limit
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/develop/api/lib/validation/schemaValidator.js#L96

boyang9527:
I will need to hold on this PR since I found CloudFoundry is not returning back Quotas in container metrics, through there is data structure there. I always got a '0' when I connect to real CF.  

So I will need to keep memoryused to make sure there is a memory metric working for now...

Meanwhile I will talk to CF people to confirm and ask why there is Quota field but it is not reported.

src/autoscaler/scalingengine/scalingengine_test.go
boyang9527: this will need change the model of trigger to carry the metric unit information.  I will add this.

api/lib/validation/schemaValidator.js
boyang9527: memoryutil should be >=- and <= 100, will add this.

boyang9527:
@rohitsharma04  "memoryusage" is a general name,  memoryutil in most cases refers to percentage

api/lib/validation/schemaValidator.js
rohitsharma04: @boyang9527 
Correct me if I'm wrong.
We're not getting quota while using polling metrics, quota is only available while streaming metrics so I think we should keep "memoryused" too and apply the upper limit validation selectively.

rohitsharma04: Also ApiServer will need to know about the collection method for the upper limit and metric type validations.

boyang9527: No, we can get quota for both polling and streaming, so we can just replace memoryused with memoryutil.

rohitsharma04: then we're good.

pradyutsarma: There may be a problem here. We have customers who are already using the `memoryused` parameter value. We cannot just remove or change that to `memoryutil`, which will be considered a regression. We have to provide a deprecation path if we want to move to `memoryutil` from `memoryused`.

boyang9527: @pradyutsarma  it is already in production?  I can keep both if your customer are seriously use it.

boyang9527:
@rohitsharma04 @pradyutsarma  I am going to close this PR first given that we need to have both memoryused and memoryutil metric. This will need major change.

PR #240, opened 2017-05-25
145587295: Scheduler changelog, newly added guid columns are non nullable which will break existing deployment

kanekoh:


api/db/api.db.changelog.yml
```diff
@@ -30,4 +30,15 @@ databaseChangeLog:
...
+   - changeSet:
+      id: 2
+      author: fujitsu
+      changes:
+      - update:
+          columns:
+          - column:
+              name: guid
+              type: varchar(50)
+              value: "1"
```
pradyutsarma: Just a question: Wouldn't this update the guid column for all rows to 1 and is that ok?

boyang9527: I guess if the data is synchronized then it probably OK....
While to keep the history data synchronized too, we need to add different ids to force the synchronization.

Need confirmation from @kanekoh and @ghaih

kanekoh: @pradyutsarma Since the guids are already populated in policy table so it will be a hassle to populate the same ones in scheduler tables as there are 2 tables. So it is ok to have it set to 1  which is easier. And this will not effect the normal processing  and will be taken care of when the synchronisation happens. Also these are used by the synchroniser only.

pradyutsarma: Aah ok. I'm getting the logic now. Sorry for the confusion. LGTM.

ghaih: No problem. Thanks!

boyang9527:
adding my review. LGTM

PR #241, opened 2017-05-29
#145819805 : Make TLS configuration (keys, certfiles) optional for the service broker

rohitsharma04:


servicebroker/test/util/apiServerUtil.test.js
```diff
@@ -0,0 +1,99 @@
+'use strict';
```
kanekoh: The file name should use '_' instead of '.' as that is the convention we are following across.

ghaih:
Just a general comment, will be good to have separate stories for the api and service broker. And there is a code refactoring here, so a separate commit for code refactor makes review bit easier. Thanks!

api/lib/utils/schedulerUtils.js
```diff
@@ -5,40 +5,54 @@ module.exports = function(schedulerSettings) {
...
+  var getOptions = function(appId, policy, policyGuid ){
+    var options = { 
+      json: true,
+      timeout: 10000
+    };
+    if(policy && policyGuid){
+      options.body = policy;
+      options.url = schedulerSettings.uri + '/v2/schedules/' + appId + "?guid=" + policyGuid;
+    }else{
+      options.url = schedulerSettings.uri + '/v2/schedules/' + appId;
```
qibobo: As policy and policyGuid are required parameter, why "if else " here?

rohitsharma04:
@kanekoh, incorporated changes,
@ghaih, I'll create a new PR for apiserver.

boyang9527:
two things we need to discuss:
1) do we need to check the client is using https to access our server even though the SSL termination has been done by router/lb in the front

2) Do we force to use https when talking to other components

PR #244, opened 2017-06-01
refactor scheduler-scalingengine integration test

boyang9527:


boyang9527:
@ghaih  sorry what do you mean by "scaling engine test cases introduced for synchronisation are in the scope of synchronisation"?

ghaih:
@boyang9527 , sorry about the confusion, I was referring to scheduler synchronisation w.r.t policy db, which is not part of your code change. I meant there is no need to check synchronisation of data to active schedules table in scaling engine. The synchroniser syncs the data between api and scheduler. The scaling engine is updated as a result of scheduler's behaviour. So no need of **Describe("Synchronized Schedule", func()** in scheduler and scaling engine synchronisation.

boyang9527:
@ghaih  I guess we need test the policy change in policy db as an input to scheduler.  I talked to Yang Qi, this integration test will need to test  the output of scaling engine(say scaling request to cf, and scaling history in DB) instead of checking the active_ schedule db in scaling engine which I think should be an internal thing for scheudler-scalingengine integration test.  

So the test cases will need to be revised while it will be tracked by another user story.

boyang9527:
@qibobo  please review this PR and give you comment as well. thanks,

qibobo:
LGTM

PR #250, opened 2017-06-13
Add debug log for raw events from loggregator

boyang9527:
I found there are some abnormal data points from NOAA. This PR is used to enable the debug log for raw events from loggregator so that issues can be tracked if there are any.


qibobo:
LGTM

PR #256, opened 2017-07-16
Remove toLowerCase from dbUri in apiserver and broker settings

pradyutsarma:
Fixes https://www.pivotaltracker.com/story/show/145872219

### DIFF
--- a/api/lib/config/setting.js
+++ b/api/lib/config/setting.js
@@ -4,7 +4,7 @@ module.exports = function(settingsObj) {
   var settingObj = {};
   var db = function(dbUri) {
     if (dbUri != null) {
-      let uri = dbUri.replace(/\/$/g, "").toLowerCase();
+      let uri = dbUri.replace(/\/$/g, "");
       let name = uri.slice(uri.lastIndexOf("/") + 1, uri.length);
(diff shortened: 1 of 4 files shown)


pradyutsarma:
The failure is in this context/test https://github.com/cloudfoundry-incubator/app-autoscaler/blob/develop/api/test/unit/config/settings_test.js#L235-L236
Can somebody tell me if we really need this test or can I just remove it from both the apiserver and the servicebroker?

pradyutsarma:
I took the liberty of removing the db mixed-to-lowercase tests from both the servicebroker and the apiserver. Will create a separate chore to look discuss https://github.com/cloudfoundry-incubator/app-autoscaler/blob/develop/api/lib/config/setting.js#L20

boyang9527:
cleanUpUri is used by the following:
lib/config/setting.js:  settingsObj.scheduler.uri = cleanUpUri(settingsObj.scheduler.uri);
lib/config/setting.js:  settingsObj.scalingEngine.uri = cleanUpUri(settingsObj.scalingEngine.uri);

since there is no usenrame:password in these two uri, so this should be fine.. while I guess we can remove the toLowerCase() in case it will be used by other places.

boyang9527:
LGTM

PR #258, opened 2017-07-18
Enable database lock for singleton metricscollector without using consul based lock

paltanmoy:
- Implemented database lock to achieve singleton for metricscolletor without using consul (Need to enable this feature by setting enable_db_lock to true)

- Added unit tests

- Modified Integration tests

### DIFF
--- a/src/autoscaler/db/sqldb/lock_sqldb.go
+++ b/src/autoscaler/db/sqldb/lock_sqldb.go
@@ -1,0 +1,241 @@
+package sqldb
+
+import (
+	"database/sql"
+	"time"
+
+	"code.cloudfoundry.org/lager"
+	_ "github.com/lib/pq"
+
+	"autoscaler/db"
+	"autoscaler/models"
+)
+
+type LockSQLDB struct {
+	url    string
+	logger lager.Logger
+	sqldb  *sql.DB
+}
+
+func NewLockSQLDB(url string, logger lager.Logger) (*LockSQLDB, error) {
+	sqldb, err := sql.Open(db.PostgresDriverName, url)
+	if err != nil {
+		logger.Error("open-lock-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	err = sqldb.Ping()
+	if err != nil {
+		sqldb.Close()
+		logger.Error("ping-lock-db", err, lager.Data{"url": url})
+		return nil, err
+	}
+
+	return &LockSQLDB{
+		url:    url,
+		logger: logger,
+		sqldb:  sqldb,
+	}, nil
+}
+
+func (ldb *LockSQLDB) Close() error {
+	err := ldb.sqldb.Close()
+	if err != nil {
+		ldb.logger.Error("close-lock-db", err, lager.Data{"url": ldb.url})
+		return err
+	}
+	return nil
+}
+
+func (ldb *LockSQLDB) Fetch() (lock *models.Lock, err error) {
+	ldb.logger.Info("fetching locks ")
+	var (
+		owner     string
+		timestamp time.Time
+		ttl       time.Duration
+	)
+	tx, err := ldb.sqldb.Begin()
+	if err != nil {
+		ldb.logger.Error("error-fetching-lock-transaction", err)
+		return nil, err
+	}
+	query := "SELECT * FROM locks"
+	err = tx.QueryRow(query).Scan(&owner, &timestamp, &ttl)
+	if err != nil {
+		return &models.Lock{}, err
+	}
+	err = tx.Commit()
+	if err != nil {
+		ldb.logger.Error("failed-to-commit-fetching-lock-transaction", err)
+		return &models.Lock{}, err
+	}
+	fetchedLock := &models.Lock{Owner: owner, LastModifiedTimestamp: timestamp, Ttl: ttl}
+	return fetchedLock, nil
+}
+
+func (ldb *LockSQLDB) Acquire(lockDetails *models.Lock) error {
+	ldb.logger.Info("acquiring-the-lock", lager.Data{"Owner": lockDetails.Owner, "LastModifiedTimestamp": lockDetails.LastModifiedTimestamp, "Ttl": lockDetails.Ttl})
+	tx, err := ldb.sqldb.Begin()
+	if err != nil {
+		ldb.logger.Error("error-starting-acquire-lock-transaction", err)
+		return err
+	}
+	defer func() {
+		if err != nil {
+			ldb.logger.Error("rolling-back-acquire-lock-transaction!", err)
+			err = tx.Rollback()
+			if err != nil {
+				ldb.logger.Error("failed-to-rollback-acquire-lock-transaction", err)
+			}
+			return
+		}
+		err = tx.Commit()
+		if err != nil {
+			ldb.logger.Error("failed-to-commit-acquire-lock-transaction", err)
+		}
+	}()
+	if _, err = tx.Exec("SELECT * FROM locks FOR UPDATE NOWAIT"); err != nil {
+		ldb.logger.Error("failed-to-select-for-update", err)
+		return err
+	}
+	currentTimestamp, err := ldb.GetDatabaseTimestamp()
+	if err != nil {
+		ldb.logger.Error("error-getting-timestamp-while-acquiring-lock", err)
+		return err
+	}
+	query := "INSERT INTO locks (owner,lock_timestamp,ttl) VALUES ($1,$2,$3)"
+	if _, err = tx.Exec(query, lockDetails.Owner, currentTimestamp, int64(lockDetails.Ttl/time.Second)); err != nil {
+		ldb.logger.Error("failed-to-insert-lock-details-during-acquire-lock", err)
+		return err
+	}
+	return err
+}
+
+func (ldb *LockSQLDB) Renew(owner string) error {
+	ldb.logger.Debug("renewing-lock", lager.Data{"Owner": owner})
+	tx, err := ldb.sqldb.Begin()
+	if err != nil {
+		ldb.logger.Error("error-starting-renew-lock-transaction", err)
+		return err
+	}
+	defer func() {
+		if err != nil {
+			ldb.logger.Error("rolling-back-renew-lock-transaction", err)
+			err = tx.Rollback()
+			if err != nil {
+				ldb.logger.Error("failed-to-rollback-renew-lock-transaction", err)
+			}
+			return
+		}
+		err = tx.Commit()
+		if err != nil {
+			ldb.logger.Error("failed-to-commit-renew-lock-transaction", err)
+		}
+	}()
+	query := "SELECT * FROM locks where owner=$1 FOR UPDATE NOWAIT"
+	if _, err = tx.Exec(query, owner); err != nil {
+		ldb.logger.Error("failed-to-select-for-update", err)
+		return err
+	}
+	currentTimestamp, err := ldb.GetDatabaseTimestamp()
+	if err != nil {
+		ldb.logger.Error("error-getting-timestamp-while-renewing-lock", err)
+		return err
+	}
+	updatequery := "UPDATE locks SET lock_timestamp=$1 where owner=$2"
+	if _, err = tx.Exec(updatequery, currentTimestamp, owner); err != nil {
+		ldb.logger.Error("failed-to-update-lock-details-during-lock-renewal", err)
+		return err
+	}
+	return err
+}
+
+func (ldb *LockSQLDB) Release(owner string) error {
+	ldb.logger.Debug("releasing-lock", lager.Data{"Owner": owner})
+	tx, err := ldb.sqldb.Begin()
+	if err != nil {
+		ldb.logger.Error("error-starting-release-lock-transaction", err)
+		return err
+	}
+	defer func() {
+		if err != nil {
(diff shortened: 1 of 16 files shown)


paltanmoy:
@boyang9527 We can have a shared lock tables across the components. But that requires a separate liquibase changelog file (as part of a separate dblock module)

boyang9527:
@paltanmoy  I guess there are some misunderstanding... 

The current one  is a single lock implementation.  The table only serves one lock.  I am hoping there will be database impl of distributed locks with an abstraction interface, which can support creating multiple locks,  and also different lock types will be even better... but we can start with non-blocking mutex

src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -208,3 +284,16 @@ func generateGUID(logger lager.Logger) string {
...
 	return uuid.String()
 }
+
+func getOwner(logger lager.Logger, conf *config.Config) string {
+	var owner string
+	if strings.TrimSpace(os.Getenv("VCAP_APPLICATION")) != "" {
+		appEnv, _ := cfenv.Current()
+		owner = appEnv.ID
+		logger.Info("ownership found in VCAP_APPLICATION", lager.Data{"owner": owner})
+	} else if conf.DBLock.Owner != "" {
```
paltanmoy: @kanekoh It's required parameter only when conf.EnableDBLock is set to true.

kanekoh: I wonder that getOwner function is called when conf.EnableDBLock is set to true. 
Am I missing something?

src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -159,6 +176,65 @@ func main() {
 	logger.Info("exited")
 }
 
+func initDBLockRunner(conf *config.Config, logger lager.Logger, lockDB db.LockDB) ifrit.Runner {
+	dbLockMaintainer := ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
+		ttl := conf.DBLock.LockTTL
+		lockTicker := time.NewTicker(ttl)
+		readyFlag := true
+		owner := getOwner(logger, conf)
```
boyang9527: If it is not deployed as a cf app,  we can use IP address and process ID, instead of using the one in config

If the component is deployed as one instance of a cf app, this will have issue.

src/autoscaler/db/sqldb/lock_sqldb_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+		Context("when no lock owner exist", func() {
+			BeforeEach(func() {
+				newlock := &models.Lock{Owner: "123456", Ttl: testTTL}
+				err = ldb.Acquire(newlock)
+			})
+			It("should successfully acquire the lock", func() {
+				Expect(err).NotTo(HaveOccurred())
+			})
+		})
```
boyang9527: what about test case when the lock owner exists?

src/autoscaler/db/sqldb/lock_sqldb_test.go
```diff
@@ -0,0 +1,270 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			cleanLockTable()
+		})
+
+		AfterEach(func() {
+			err = ldb.Close()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		Context("when lock owned by an instance", func() {
```
boyang9527: what about  nobody owns the lock, and the lock is owned by others.

src/autoscaler/db/sqldb/lock_sqldb_test.go
```diff
@@ -0,0 +1,270 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			cleanLockTable()
+		})
+
+		AfterEach(func() {
+			err = ldb.Close()
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		Context("when owner instance wants to release the lock", func() {
```
boyang9527: missing test cases when lock table is empty,  and the lock is owned by others

src/autoscaler/metricscollector/cmd/metricscollector/main.go
```diff
@@ -159,6 +176,65 @@ func main() {
...
+				if isLockAcquired && readyFlag {
+					close(ready)
+					readyFlag = false
+					logger.Info("successfully-acquired-lock", lager.Data{"owner": owner})
+				}
+			}
+		}
+	})
+	return dbLockMaintainer
+}
```
boyang9527: if only  Lock() and Release() are used,  why we have 	following as Public
 Fetch() (*models.Lock, error)
 Acquire(lock *models.Lock) error
 Renew(owner string) error

boyang9527: by the way,  I guess we can define a lock interface, and then use DBLock as one of the implementation.

paltanmoy:
@boyang9527 , @kanekoh  most of your review comments has been addressed in the updated PR. Please have a look.

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+func NewDatabaseLock(logger lager.Logger) *DatabaseLock {
+	return &DatabaseLock{
+		logger: logger,
+	}
+}
+
+func (dblock *DatabaseLock) InitDBLockRunner(ttl time.Duration, owner string, lockDB db.LockDB) ifrit.Runner {
+	dbLockMaintainer := ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
+		lockTicker := time.NewTicker(ttl)
+		readyFlag := true
```
boyang9527: a little bit misleading, perhaps initialize with false, when it is ready then assign true value

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+			case <-signals:
+				dblock.logger.Info("received-interrupt-signal", lager.Data{"owner": owner})
+				lockTicker.Stop()
+				releaseErr := lockDB.Release(owner)
+				if releaseErr != nil {
+					dblock.logger.Error("failed-to-release-lock ", releaseErr)
+				} else {
+					dblock.logger.Info("successfully-released-lock", lager.Data{"owner": owner})
+				}
+				readyFlag = true
```
boyang9527: why set readyFalg, given you immediately return nil

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+			dblock.logger.Info("lock-acquired-in-first-attempt", lager.Data{"owner": owner, "isLockAcquired": isLockAcquired})
+			close(ready)
+			readyFlag = false
+		}
+		for {
+			select {
+			case <-signals:
+				dblock.logger.Info("received-interrupt-signal", lager.Data{"owner": owner})
+				lockTicker.Stop()
+				releaseErr := lockDB.Release(owner)
```
boyang9527: don't need to release if the lock is not acquired.

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+
+func NewDatabaseLock(logger lager.Logger) *DatabaseLock {
+	return &DatabaseLock{
+		logger: logger,
+	}
+}
+
+func (dblock *DatabaseLock) InitDBLockRunner(ttl time.Duration, owner string, lockDB db.LockDB) ifrit.Runner {
+	dbLockMaintainer := ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
+		lockTicker := time.NewTicker(ttl)
```
boyang9527: should not we use LockRetryInterval instead of ttl ?

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+				readyFlag = true
+				return nil
+
+			case <-lockTicker.C:
+				dblock.logger.Info("retry-acquiring-lock", lager.Data{"owner": owner})
+				lock := &models.Lock{Owner: owner, Ttl: ttl}
+				isLockAcquired, lockErr := lockDB.Lock(lock)
+				if lockErr != nil {
+					dblock.logger.Error("failed-to-acquire-lock", lockErr)
+					releaseErr := lockDB.Release(owner)
```
boyang9527: why bothered by release the lock if lock failed?

src/autoscaler/db/sqldb/lock_sqldb.go
```diff
@@ -0,0 +1,242 @@
...
+	})
+	return err
+}
+
+func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
+	ldb.logger.Debug("acquiring-lock", lager.Data{"Owner": lock.Owner})
+	isLockAcquired := true
+	err := ldb.transact(ldb.sqldb, func(tx *sql.Tx) error {
+		newLock := false
+		fetchedLock, err := ldb.fetch(tx)
```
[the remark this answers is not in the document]
paltanmoy: done

src/autoscaler/db/sqldb/lock_sqldb.go
```diff
@@ -0,0 +1,242 @@
...
+
+			}
+			return err
+		}()
+
+		// golang sql package does not always retry query on ErrBadConn
+		if attempts >= 2 || (err != driver.ErrBadConn) {
+			break
+		} else {
+			ldb.logger.Error("deadlock-transaction", err, lager.Data{"attempts": attempts})
```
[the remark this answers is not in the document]
paltanmoy: for debugging purpose

src/autoscaler/metricscollector/config/config.go
```diff
@@ -70,27 +70,35 @@ type LockConfig struct {
...
 }
 
 type Config struct {
-	Cf        cf.CfConfig     `yaml:"cf"`
-	Logging   LoggingConfig   `yaml:"logging"`
-	Server    ServerConfig    `yaml:"server"`
-	Db        DbConfig        `yaml:"db"`
-	Collector CollectorConfig `yaml:"collector"`
-	Lock      LockConfig      `yaml:"lock"`
+	Cf           cf.CfConfig     `yaml:"cf"`
```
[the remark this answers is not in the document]
paltanmoy: How about refactoring the config through a separate PR ?

src/autoscaler/sync/dblock.go
paltanmoy: we are renewing the lock every ttl timeframe. Database lock does not have any retry interval attribute.

src/autoscaler/sync/dblock.go
paltanmoy: the readyFlag is being used to indicate whether it's ready to get the lock, initially set to true to indicate that the instance is ready to get the lock, as soon as it got the lock we made it to false.

src/autoscaler/sync/dblock.go
```diff
@@ -0,0 +1,79 @@
...
+	return &DatabaseLock{
+		logger: logger,
+	}
+}
+
+func (dblock *DatabaseLock) InitDBLockRunner(ttl time.Duration, owner string, lockDB db.LockDB) ifrit.Runner {
+	dbLockMaintainer := ifrit.RunFunc(func(signals <-chan os.Signal, ready chan<- struct{}) error {
+		lockTicker := time.NewTicker(ttl)
+		readyFlag := true
+		if owner == "" {
```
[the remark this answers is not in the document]
paltanmoy: the generateGUID() has been modified to send empty string during uuid generation error.

src/autoscaler/sync/dblock.go
paltanmoy: Release is a delete operation on owner id, will succeed only if lock exists.

src/autoscaler/sync/dblock.go
paltanmoy: it might fail to renew the lock for some reason (VM Crash/shutdown etc). In that case rather than waiting for the instance, we should forcefully release the existing lock, so that other instance could get the lock.

src/autoscaler/db/sqldb/lock_sqldb.go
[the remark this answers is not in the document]
boyang9527: sorry i mean why this indicates 'deadlock'

src/autoscaler/sync/dblock.go
boyang9527: will this cause loss of the lock? I am assuming we need to refresh the lock less than ttl 
why not we add this? like consul lock?

src/autoscaler/sync/dblock.go
boyang9527: it is confusing since there is also ready channel while it has different meaning... If so can we change the variable name?

src/autoscaler/sync/dblock.go
boyang9527: other instance can still get the lock after ttl ...so that is fine not to do it...

paltanmoy: If we release the lock as and when it fails to renew, then it does not require to wait till the ttl time to release forcefully. Lock switch will be limited within retry-interval timeframe in this scenario.

paltanmoy:
@kanekoh & @boyang9527 any more comments ?

ghaih:
No more comments from our side.Thanks.

PR #262, opened 2017-08-08
Unify api-server rest api path

qibobo:
[#150033265]


cdlliuy:
but the travis build looks hang. althrough it passed as green, but git still shows the "some checks haven't completed yet"

PR #266, opened 2017-08-28
#145124007 : Provide default values for mandatory attributes(stat_window_secs, breach_duration_secs and cool_down_secs) in scaling_rules in policy document

rohitsharma04:


pradyutsarma:
LGTM

PR #268, opened 2017-08-30
cli command design

boyang9527:
[#150697341]

### DIFF
--- a/src/cli/readme.MD
+++ b/src/cli/readme.MD
@@ -1,0 +1,94 @@
+# Cloud Foundry CLI AutoScaler Plugin
+
+AutoScaler plugin provides the command line interface to manage AutoScaler service policies, retrieve metrics and scaling histories.
+
+## Command List
+
+| Command | Description |
+|---------|-------------|
+| [autoscaler-api, asa](#cf-autoscaler-api) | Set or view AutoScaler service API endpoint |
+| [autoscaler-policy, asp](#cf-autoscaler-policy) | Retrieve the scaling policy of an application |
+| [attach-autoscaler-policy, aasp](#cf-attach-autoscaler-policy) | Attach a scaling policy to an application |
+| [detach-autoscaler-policy, dasp](#cf-detach-autoscaler-policy) | Detach the scaling policy from an application |
+| [autoscaler-metrics, asm](#cf-autoscaler-metrics) | Retrieve the metrics of an application |
+| [autoscaler-histories, ash](#cf-autoscaler-histories) | Retrieve the scaling histories of an application|
+
+## Command usage
+
+### `cf autoscaler-api`
+
+Set or view AutoScaler service API endpoint
+
+```
+cf autoscaler-api [URL] [--unset] 
+```
+
+**ALIAS**: asa
+
+**OPTIONS**
+- `--unset`: Unset the api endpoint
+
+### `cf autoscaler-policy` 
+
+Retrieve the scaling policy of an application
+
+```
+cf autoscaler-policy APP_NAME [-f PATH_TO_FILE]
+```
+
+**ALIAS**: asp
+
+
+**OPTIONS**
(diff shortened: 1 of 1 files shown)


src/cli/readme.MD
```diff
@@ -0,0 +1,80 @@
...
+
+```
+cf as-policy APP_NAME [-f PATH_TO_FILE]
+```
+
+**OPTIONS**
+- `-f` : dump the policy to a file in JSON format
+
+### `cf attach-as-policy` 
+
```
ghaih: Just a suggestion, good to have "aas" (app-autoscaler) instead of "as". When  using it in "attach-as-policy", it sounds like having policy as an option. Also is it good to have "aas" prepended in all commands as in
cf aas-api
cf aas-attach-policy
cf aas-detach-policy
and so on

cdlliuy: I think "as" is concise, but it is a good idea to have "as" as the prefix.  how about 
cf as-policy-show
cf as-policy-attach
cf as-policy-detach

julz: is it odd to abbreviate here though when cli commands can also have abbreviations? i.e. maybe it should be `cf attach-autoscaler-policy` and then we can provide a short-form like `cf aap` too. (which fits with things like `cf create-service-instance` + `cf csi` rather than `cf create-s-instance` iyswim)

julz: (either way I think we need to avoid `as` definitely because there's a lot of confusion there -- I read `detach-as-policy` as "detach as policy" as in the English word "as")

ghaih: Yes thats the point I was trying to make about "as".

boyang9527: agreed.  'as' may cause confusion.  Let us use full name autoscaler, and also provide short name for each command

boyang9527:
@kanekoh src/github.com/ramya-rao-a/go-outline should not be there. I just removed it

boyang9527:
@ghaih @julz @cdlliuy  I revised the command design. Please review again.  Basically 
1) use full name 'autoscaler' in command name
2) give each command an alias as short name

pradyutsarma:
I'm fine with "autoscaler" if everybody else is. but somehow "scaling" sounds better. I'm more for the following,
`cf attach-scaling-policy or cf asp`
`cf detach-scaling-policy or cf dsp`
`cf scaling-policy or cf sp`
`cf scaling-metrics or cf sm`
`cf scaling-history or cf sh` (I'm not sure if histories is correct)

boyang9527:
cf sp is already occupied....
So I guess we can do 
```
cf attach-autoscaling-policy, cf aasp
cf detach-autoscaling-policy, cf dasp
cf autoscaling-policy, cf asp
cf autoscaling-metrics ,  cf asm
cf autoscaling-history,  cf ash
```

boyang9527:
@pradyutsarma @ghaih  @cdlliuy @kanekoh @julz 
any additional comments for the latest cli design?

ghaih:
A suggestion from consistency point of view,

cf attach-autoscaling-policy ==> cf autoscaling-attach-policy
cf detach-autoscaling-policy ==> cf autoscaling-detach-policy
Since rest have autoscaling in the beginning. Rest all looks and sounds good :)

boyang9527:
@ghaih @kanekoh 
The cf cli naming convention are 
- none for showing a resource 
-  plural noun for list the resources
- verb-none for take action on a resource 

So I would keep the original design. Let me know if you have additional comments.

kanekoh:
Understand Thanks.
No more comments from our side (kaneko, Harpreet).

boyang9527:
@julz @pradyutsarma 
Let me know if you are good with the latest version of the CLI design.

pradyutsarma:
LGTM

cdlliuy:
LGTM

julz:
👍

PR #273, opened 2017-09-14
Adding Varchar Column Length

kmj940101:
I have noticed that some of the columns are missing sizes. 
It is always good to follow consistency.

### DIFF
--- a/src/autoscaler/scalingengine/db/scalingengine.db.changelog.yml
+++ b/src/autoscaler/scalingengine/db/scalingengine.db.changelog.yml
@@ -38,17 +38,17 @@ databaseChangeLog:
                     nullable: false
               - column:
                   name: reason
-                  type: varchar(255)
+                  type: text
                   constraints:
                     nullable: false
               - column:
                   name: message
-                  type: varchar(255)
+                  type: text
                   constraints:
                     nullable: true
               - column:
                   name: error
-                  type: varchar(255)
+                  type: text
                   constraints:
                     nullable: true
         - createTable:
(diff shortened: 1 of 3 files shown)


src/autoscaler/scalingengine/db/scalingengine.db.changelog.yml
```diff
@@ -38,25 +38,25 @@ databaseChangeLog:
                     nullable: false
               - column:
                   name: reason
-                  type: varchar
+                  type: varchar(255)
```
boyang9527: we probably will need larger size for reason, message and error

kmj940101: @boyang9527 Thank you for your comment. How about changing the data type to a longer type such as TEXT?

screenshot from 2017-09-20 09-29-09

kmj940101:
@rohitsharma04 Thank you for your review. Why do we need another changeset?
I did not add new columns to the existing table neither update the value.

paltanmoy:
@kmj940101 It's an alteration of table structure, if you don't add this in a new changeset, Liquibase wont consider this changes (in existing deployments) assuming it has all changes in the databasechangelog table

boyang9527:
LGTM

kmj940101:
I’m closing this pull request because of inactivity. One of the reasons why I made this pull request was mysql needs to define a maximum size for varchar. I will submit another pull request soon.
Related Issue: https://github.com/cloudfoundry-incubator/app-autoscaler/issues/174

PR #275, opened 2017-09-14
Write 0 responsetime due to no requests.

7CHANHO:
## Issue
If there is no request, 'scaled in' by 'response time' policy never happens.
I think this is the intended behavior.
But I would like to suggest a new opinion.

## Problem
Suppose your app is scaled out. (App intances: 1 --> 2)
If there is no request to your app, eventgenerator does not send a trigger alarm because there is no value in the DB table.
When there is no request, there is no value in the DB table as follows:
```
autoscaler=> SELECT * FROM app_metric ORDER BY timestamp DESC LIMIT 3;
                app_id                | metric_type  | unit |      timestamp      | value 
--------------------------------------+--------------+------+---------------------+-------
 e569062a-fcbe-4660-a8a1-8dd00630a17d | responsetime |      | 1505365079880967557 | 
 e569062a-fcbe-4660-a8a1-8dd00630a17d | responsetime |      | 1505365079880918445 | 
 e569062a-fcbe-4660-a8a1-8dd00630a17d | responsetime |      | 1505365039881445857 | 
```
If there is no value like the above, the following message is displayed in `eventgenerator`.
```
{"timestamp":"1505366359.882145643","source":"eventgenerator","message":"eventgenerator.Evaluator.should not send trigger alarm to scaling engine because there is empty value metric","log_level":0,"data":{"appMetric":{"AppId":"e569062a-fcbe-4660-a8a1-8dd00630a17d","MetricType":"responsetime","Value":"","Unit":"","Timestamp":1505366319881877778},"session":"17","trigger":{"app_id":"e569062a-fcbe-4660-a8a1-8dd00630a17d","metric_type":"responsetime","metric_unit":"","breach_duration_secs":60,"threshold":5,"operator":"\u003c","cool_down_secs":300,"adjustment":"-1"}}}
```
As a result, the app remains scaled out with no requests at all.
If you are on a usage-based billing basis, these results may be considered unreasonable.

## Suggestion
I solved the above problem by storing the value of response time as 0, like the 'throughput' metric.
If the cause of delaying the response time has not been resolved, the app will scale out again on the next request.
Please review this PR.


cdlliuy:
@7CHANHO , thanks for the submission. but when you provide the fix, could you also add related test case in app-autoscaler/src/autoscaler/metricscollector/collector/app_streamer_test.go to validate the code change?

7CHANHO:
Hello @cdlliuy ,
I have added a test case.

7CHANHO:
I think your opinion is right. @boyang9527 
I have verified through testing that this problem is resolved if another metric other than responsetime is specified in the policy. Thank you :)

PR #276, opened 2017-09-19
return CLI metadata

boyang9527:
[#151235404]
- return cli meta data
- add how to build/install cli in readme
- change option name '-f' to '--output' since '-f' usually means force execution


qibobo:
LGTM

PR #279, opened 2017-09-29
add include=all parameter in scaling history api, only return succedd…

boyang9527:
…ed and failed history by default

[#151547679]

- db pacakge changes to add includeAll flag when retrieving scaling history
- add include parameter in scaling history api
- regenerate fakes due to interface change

### DIFF
--- a/src/autoscaler/db/sqldb/scalingengine_sqldb.go
+++ b/src/autoscaler/db/sqldb/scalingengine_sqldb.go
@@ -2,7 +2,6 @@ package sqldb
 
 import (
 	"database/sql"
-	"fmt"
 
 	"code.cloudfoundry.org/lager"
 	_ "github.com/lib/pq"
@@ -69,16 +68,11 @@ func (sdb *ScalingEngineSQLDB) RetrieveScalingHistories(appId string, start int6
 	} else {
 		orderStr = db.ASCSTR
 	}
-	var includeStr string
-	if !includeAll {
-		includeStr = fmt.Sprintf(" AND status != %d", models.ScalingStatusIgnored)
-	}
 
 	query := "SELECT timestamp, scalingtype, status, oldinstances, newinstances, reason, message, error FROM scalinghistory WHERE" +
 		" appid = $1 " +
 		" AND timestamp >= $2" +
 		" AND timestamp <= $3" +
-		includeStr +
 		" ORDER BY timestamp " + orderStr
 
 	if end < 0 {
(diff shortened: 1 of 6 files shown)


cdlliuy:
LGTM

src/autoscaler/db/sqldb/scalingengine_sqldb.go
```diff
@@ -60,17 +62,24 @@ func (sdb *ScalingEngineSQLDB) SaveScalingHistory(history *models.AppScalingHist
...
+func (sdb *ScalingEngineSQLDB) RetrieveScalingHistories(appId string, start int64, end int64, orderType db.OrderType, includeAll bool) ([]*models.AppScalingHistory, error) {
 	var orderStr string
 	if orderType == db.DESC {
 		orderStr = db.DESCSTR
 	} else {
 		orderStr = db.ASCSTR
 	}
+	var includeStr string
+	if !includeAll {
+		includeStr = fmt.Sprintf(" AND status != %d", models.ScalingStatusIgnored)
```
qibobo: It is not a good way to use "!=" in a sql query as it will scan all the records. It is better to use "in()" or "=".

boyang9527: in() will still have the issue, so I moved the filtering out of db query

boyang9527:
@qibobo  I pushed a new commit to move the status filtering out of db query, which will address the issue that you mentioned.

qibobo:
@boyang9527  conflicting with latest.

boyang9527:
@qibobo  conflict solved.

qibobo:
@boyang9527 @cdlliuy  I did a performance test for select sql "in()" and "!=" (https://github.com/qibobo/dbperformance). Two test cases:

### select 200,000 from 1,000,000 rows (unit : ms)
|       | 1     | 2   | 3   | 4  |
| --- | --- |--- |--- |--- |
| !=  | 1113 |1082  |1049  | 1070 |
| in  | 1143 | 1135 | 1128  | 1157 |

### select 2,000,000 from 10,000,000 rows (unit : ms)
|       | 1     | 2   | 3   | 4  |
| --- | --- |--- |--- |--- |
| !=  | 10915 |10401  |10465  | 10363 |
| in  | 11033 | 11153 | 12014  | 12664 |

According to the test result, we can see that "!=" is a little faster than "in", but the gap is small.

qibobo:
LGTM

PR #280, opened 2017-10-03
Add in CPU percentage data for scaling.

andrewdriver123:
## What

We found that the existing set of metrics supported by the autoscaler was not suitable for our workloads. Most of our apps are java containers and they do not release the memory back to the operating system, so attempting to scale applications based on memory consumption resulted in all apps always being scaled to the max instances and never being scaled back down again.

We noticed that you had a spike in your Pivotal Tracker for adding the CPU scaling back in again.
 https://www.pivotaltracker.com/story/show/144124793 - The ticket in your tracker makes some noises about Diego being able to enforce CPU quotas - We checked and it still doesn't do that - It just uses a best-effort fairness algorithm to try and ensure that all containers get some CPU from time to time, so noisy neighbours may still have an impact on the CPU metrics emitted (i.e. never able to reach 100% CPU usage in a container because other containers are already using the CPU)

We understand that the CPU metric was removed in 2015 as the CPU metrics emitted from diego cells are not normalised to take account of the number of CPUs present - https://developer.ibm.com/answers/questions/211364/changes-to-bluemix-auto-scaling-service-metric-cpu/ - However, for our particular use case, all of our diego cells are identically sized so normalisation of the CPU metrics is not essential.

## Features:
- Store the cm.GetCpuPercentage value into the database as cpuPercentage
- Allow cpuPercentage to be a valid entry in the schema
- Add tests for the new code.
- update the readme

We thought we'd fire a PR off over to you to see if you'd like to merge in these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@
 /.project
 /.settings/
 *.coverprofile
+npm-shrinkwrap.json
+src/autoscaler/metricscollector/metricscollector
--- a/src/autoscaler/models/metrics.go
+++ b/src/autoscaler/models/metrics.go
@@ -8,6 +8,7 @@ const (
 	UnitRPS          = "rps"
 )
 
+const MetricNameCpuPercentage = "cpuPercentage"
 const MetricNameMemoryUtil = "memoryutil"
 const MetricNameMemoryUsed = "memoryused"
 const MetricNameThroughput = "throughput"
--- a/api/lib/validation/schemaValidator.js
+++ b/api/lib/validation/schemaValidator.js
@@ -57,7 +57,7 @@ var getDaysInMonthInISOFormat = function() {
 };
 
 var getMetricTypes = function() {
-  var metricTypeEnum = ['memoryused', 'memoryutil', 'responsetime', 'throughput'];
+  var metricTypeEnum = ['memoryused', 'memoryutil', 'responsetime', 'throughput', 'cpuPercentage'];
   return metricTypeEnum;
 };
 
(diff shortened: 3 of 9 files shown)


src/autoscaler/models/metrics.go
```diff
@@ -8,6 +8,7 @@ const (
 	UnitRPS          = "rps"
 )
 
+const MetricNameCpuPercentage = "cpuPercentage"
```
kmj940101: How about "cpuused"?
```
const MetricNameCpuUsed = "cpuused"
```

.gitignore
```diff
@@ -7,3 +7,5 @@
 /.project
 /.settings/
 *.coverprofile
+npm-shrinkwrap.json
+src/autoscaler/metricscollector/metricscollector
```
kmj940101: Why did you add this to .gitignore file?

boyang9527:
we prefer to hold this PR since it is hard to define a cpu policy since 
1)  the number of cores of the cell is a hardware parameter. The end user has no idea about it. The cpu usage reported can be more than 100% depending on how many cores the cell has.

2) you never know what is the CPU share since it depends on other apps in the same cell.  So the cpu  share changes dynamically 

By the way, the cpu maximum is introduced in in garden 1.3.0, while most of the deployment does not enable that yet.... https://github.com/cloudfoundry/garden-runc-release/releases/tag/v1.3.0

boyang9527:
@andrewdriver123 I am going to close this PR since we need more discussion on this.... 
please give your comments here: https://docs.google.com/document/d/1PBOati9Opusfss2s7HX_Hl3D0A-SRTmmMhNk2NRCUSo/edit#

PR #283, opened 2017-10-10
remove -p in integration test in readme

boyang9527:
we don't support run scheduler-scalingengine integration test in parallel


qibobo:
LGTM

PR #284, opened 2017-10-11
Public api should only take effective on bound apps, and reject to se…

qibobo:
…rve for the unbound apps.

[#151801082]

### DIFF
--- a/api/lib/oauth/oauth.js
+++ b/api/lib/oauth/oauth.js
@@ -31,7 +31,7 @@ module.exports = function(configFilePath) {
           method: "GET",
           json: true,
           timeout: 10000,
-          rejectUnauthorized: false,   
+          rejectUnauthorized: false,
           headers: {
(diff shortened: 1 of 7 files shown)


api/lib/oauth/oauth.js
```diff
@@ -69,6 +69,58 @@ module.exports = function(configFilePath) {
...
+      url: ccEndpoint + "/v2/apps/" + appId + "/env",
+      method: "GET",
+      json: true,
+      timeout: 10000,
+      headers: {
+        "Authorization": userToken,
+        "Content-Type": "application/json"
+      }
+    };
+    request(options, function(error, response, body) {
```
boyang9527: why not checking it from our own database?

qibobo: By checking from database, it will add broker's database as dependency of apiserver.

PR #286, opened 2017-10-12
Service broker provides a internal api for checking an application's …

qibobo:
…binding

[#151906019]


boyang9527:
let us hold this PR before we make a decision whether we go this way

PR #287, opened 2017-10-17
#150843282 : Delegate DB calls to separate goroutine in the metrics collector

rohitsharma04:


src/autoscaler/metricscollector/collector/app_streamer_test.go
```diff
@@ -112,50 +111,17 @@ var _ = Describe("AppStreamer", func() {
 					Value:         "67",
 					Timestamp:     222222,
 				}))
-
-				By("collecting and computing throughput")
-				Consistently(database.SaveMetricCallCount).Should(Equal(4))
```
boyang9527: why remove the test of collecting throughput metrics? 
Even there is only container metrics, we need to save throughput = 0  periodically

src/autoscaler/metricscollector/collector/app_streamer_test.go
```diff
@@ -174,20 +140,12 @@ var _ = Describe("AppStreamer", func() {
...
-				By("collecting and computing throughput and responsetime for second interval")
-				Consistently(database.SaveMetricCallCount).Should(Equal(2))
-
-				By("save throughput and responsetime metric after the second collect interval")
-				fclock.Increment(TestCollectInterval)
-				Eventually(database.SaveMetricCallCount).Should(Equal(4))
+				msgChan <- noaa.NewHttpStartStopEnvelope(333333, 100000000, 300000000, 1)
+				msgChan <- noaa.NewHttpStartStopEnvelope(555555, 300000000, 600000000, 1)
+				msgChan <- noaa.NewHttpStartStopEnvelope(666666, 300000000, 700000000, 1)
+				fclock.WaitForWatcherAndIncrement(TestCollectInterval)
```
boyang9527: check the channel is empty before you move the flock.. this will make sure the new metric is generated in the next interval

src/autoscaler/metricscollector/collector/collector.go
```diff
@@ -110,3 +122,23 @@ func (c *Collector) GetCollectorAppIds() []string {
...
 }
+
+func (c *Collector) SaveMetricsInDB() {
+	ticker := c.cclock.NewTicker(c.collectInterval)
+	metrics := make([]*models.AppInstanceMetric, 0)
+	for {
+		select {
+		case metric := <-c.dataChan:
+			metrics = append(metrics, metric)
+		case <-ticker.C():
```
boyang9527: I guess we can do it more faster than collection time,  to make data available to event generator as soon as possible...

src/autoscaler/metricscollector/collector/app_streamer.go
```diff
@@ -26,9 +26,10 @@ type appStreamer struct {
 	numRequests     map[int32]int64
 	sumReponseTimes map[int32]int64
 	ticker          clock.Ticker
+	dataChan        chan *models.AppInstanceMetric
 }
 
-func NewAppStreamer(logger lager.Logger, appId string, interval time.Duration, cfc cf.CfClient, noaaConsumer noaa.NoaaConsumer, database db.InstanceMetricsDB, sclock clock.Clock) AppCollector {
+func NewAppStreamer(logger lager.Logger, appId string, interval time.Duration, cfc cf.CfClient, noaaConsumer noaa.NoaaConsumer, database db.InstanceMetricsDB, sclock clock.Clock, dataChan chan *models.AppInstanceMetric) AppCollector {
```
qibobo: Currently the metric data will put into dataChan, so the parameter "database" is useless.

rohitsharma04:
@boyang9527 and @qibobo updated the PR.

src/autoscaler/metricscollector/collector/collector.go
rohitsharma04: @boyang9527 I introduced a new parameter saveInterval which is by default set to 5 secs and is also configurable.

boyang9527:
LGTM

qibobo:
LGTM

PR #288, opened 2017-10-25
149082123: VMs start running even if the db url/credentials is incorrect or db is unreachable

ghaih:
@boyang9527 As discussed on slack, please review, Rest of the discussion can be done on tracker story.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/SchedulerApplication.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/SchedulerApplication.java
@@ -1,7 +1,9 @@
 package org.cloudfoundry.autoscaler.scheduler;
 
+import org.apache.commons.lang.exception.ExceptionUtils;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.hibernate.exception.GenericJDBCException;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.actuate.autoconfigure.AuditAutoConfiguration;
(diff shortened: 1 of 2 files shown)


cdlliuy:
@ghaih , the tracker issue said :when the individual components are deployed, each of them should check whether the DB is reachable and should fail to start in case of an unreachable DB or incorrect credentials.

so, the PR is only address the scheduler part, right ? are you planning to go forward to cover each component?

ghaih:
@cdlliuy , Thats right scheduler only handled here, thats why didnot change story status. I guess go components are fine. Rest of components can be take up by anyone.

boyang9527:
I saw you are detecting whether there is db initialization issue by capturing the jdbc exceptions. Are there any other possibilities than jdbc exception  that we need to exit  for scheduler? 

Did you observe the process exit when bosh deploy scheduler with a wrong db setting?

ghaih:
The unit test worked fine with the incorrect db setting, but the bosh deploy worked ok. That was my concern which I wanted to understand why the depoly does not fail.

boyang9527:
@ghaih  I will try to deploy with a wrong db setting and see what happens...

ghaih:
Cannot think of any other exception that would require an exit.

boyang9527:
@ghaih I tested this PR with cf-deployment.  The scheduler job only failed when spring.datasource.url / username / password is not set correctly...  it still starts successfully when spring.policyDbDataSource is incorrect. 

So looks like scheduler only checks db settings for application and quartz, not for policy

ghaih:
@boyang9527 , Thanks for letting me know. I will look at it once I get a chance.

cdlliuy:
Close this PR as it is an incomplete solution, and will drive it further in another PR.

PR #290, opened 2017-11-08
#151855308 : Redacting sensitive data in the logs

rohitsharma04:
example:
```
{"data":{"authUrl":"https://login.bosh-lite.com/oauth/token","form":{"client_id":["admin"],"client_secret":"*REDACTED*","grant_type":["client_credentials"]},"session":"1"},"log_level":1,"message":"metricscollector.cf.request-token-grant","source":"metricscollector","timestamp":"1510306805.491698027"}
```

### DIFF
--- a/src/autoscaler/eventgenerator/cmd/eventgenerator/main.go
+++ b/src/autoscaler/eventgenerator/cmd/eventgenerator/main.go
@@ -150,7 +150,14 @@ func initLoggerFromConfig(conf *config.LoggingConfig) lager.Logger {
 		os.Exit(1)
 	}
 	logger := lager.NewLogger("eventgenerator")
-	logger.RegisterSink(lager.NewWriterSink(os.Stdout, logLevel))
+
+	keyPatterns := []string{"[Pp]wd", "[Pp]ass", "[Ss]ecret", "[Tt]oken", "dbur[il]"}
+
+	redactedSink, err := lager.NewRedactingWriterSink(os.Stdout, logLevel, keyPatterns, nil)
+	if err != nil {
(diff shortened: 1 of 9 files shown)


src/autoscaler/cf/client.go
```diff
@@ -122,7 +122,7 @@ func (c *cfClient) retrieveEndpoints() error {
 }
 
 func (c *cfClient) requestTokenGrant(formData *url.Values) error {
-	c.logger.Info("request-token-grant", lager.Data{"authUrl": c.authUrl, "form": *formData})
```
boyang9527: I would prefer to have a function to replace passwords, credentials to be something like [hidden]. It can detect the type, and have several keywords mapping... 

I am hoping this function can handle map, db url,  http request/response ( to hide authorization header if you log the request headers)

boyang9527: and I believe there are other places we need to hide passwords, credentials and tokens... I pointed you here just giving one example.

Another case is dburl which will have password in it.  

We need to go over the code and have a sanity check ...

rohitsharma04:
@boyang9527 I've updated the PR for hiding sensitive data for all golang components. 
For the node components I've also implemented something similar in node component's logger but as we're currently planning for audit logs in the node components, there might be some major changes in the logger so I'd like to do a separate PR for node components once we've decided on the approach.
Please review this PR.

boyang9527:
@rohitsharma04  looks like it will hide all the dburl, while we only want to hide password part in the db url.

boyang9527:
@rohitsharma04  how you are going to do with this PR?  We really don't want to redact the whole DB URL.

boyang9527:
LGMT

pradyutsarma:
LGTM

PR #291, opened 2017-11-13
Making rejectUnauthorized configurable to connect to CC & UAA

paltanmoy:
[#152672776]


cdlliuy:
Looks good for me , but  agree with the Ben's comments.    'skip-ssl-validation'  looks more straightforward.

boyang9527:
@paltanmoy  you are right...

PR #304, opened 2018-02-01
VMs start running even if the db url/credentials is incorrect or db i…

qibobo:
…s unreachable for apiserver

[#154807174]


api/app.js
```diff
@@ -9,14 +9,10 @@ module.exports = function(configFilePath) {
...
-
-  if (!configFilePath || !fs.existsSync(configFilePath)) {
-      logger.error("Invalid configuration file path: " + configFilePath);
-      throw new Error('configuration file does not exist:' + configFilePath);
-  }
-  var settings = require(path.join(__dirname, './lib/config/setting.js'))((JSON.parse(
-        fs.readFileSync(configFilePath, 'utf8'))));
+  
+  var oauth = require('./lib/oauth/oauth')(settings);
+  var models = require('./lib/models')(settings.db, callback);
```
cdlliuy: It is recommend to do setting.validate first, then using the setting to initialize the models

api/index.js
```diff
@@ -1,7 +1,22 @@
...
+if (!configFilePath || !fs.existsSync(configFilePath)) {
+      logger.error("Invalid configuration file path: " + configFilePath);
+      throw new Error('configuration file does not exist:' + configFilePath);
+  }
+var settings = require(path.join(__dirname, './lib/config/setting.js'))((JSON.parse(
+        fs.readFileSync(configFilePath, 'utf8'))));
+var dbErrorCallback = function(err) {
+  if (err) {
+    logger.error('DB configuration is incorrect, server will exit', err);
+    process.exit(1);
```
cdlliuy: when setting is invalid, api server exit with " throw new Error("Invalid public TLS ca certificate path: " + settings.publicTls.caCertFile);" instead of "process.exit(1)
should we handle these kinds of  FATAL error with a consistent approach?

cdlliuy:
LGTM

PR #307, opened 2018-02-09
Incorrect time range overlapping detection in api server

cdlliuy:
#155061669.

This PR is used to fix #155061669.

Also, enhance the unit test of policy validation.  Expands API server's  total unit test case from 353 to 895.

### DIFF
--- a/api/lib/validation/attributeValidator.js
+++ b/api/lib/validation/attributeValidator.js
@@ -20,6 +20,19 @@ var compareDateTimes = function(startDateTime,endDateTime) {
   return moment(startDateTime,'YYYY-MM-DDTHH:mm')
     .isSameOrAfter(moment(endDateTime,'YYYY-MM-DDTHH:mm'));
 }
+var normalizeStartDate = function(object) {
+  if ( !object.hasOwnProperty("start_date") ) {
+    return '0001-01-01'
+  }  
+  return object.start_date;
+}
+var normalizeEndDate = function(object) {
+  if ( !object.hasOwnProperty("end_date") ) {
+    return '9999-12-31'
+  }  
+  return object.end_date;
+}
+
 var validateOverlappingDaysInRecurringSchedules = function(inputRecurringSchedules,property) {
   var errors = [];
   var errorCount = 0;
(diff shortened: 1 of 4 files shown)


cdlliuy:
For reviewers, all testcases in  api/test/unit/validation/validationMiddleware.test.js are refactored with mocha "dynamic test case generating" feature.  No need to view diff, please just check the committed code ..

api/lib/validation/attributeValidator.js
```diff
@@ -20,6 +20,19 @@ var compareDateTimes = function(startDateTime,endDateTime) {
   return moment(startDateTime,'YYYY-MM-DDTHH:mm')
     .isSameOrAfter(moment(endDateTime,'YYYY-MM-DDTHH:mm'));
 }
+var normalizeStartDate = function(object) {
```
boyang9527: why we do normalization, start_date and end_date are required fields, aren't they?

cdlliuy: @boyang9527 , start_date and end_date are optional fields, so it is allowed to set "start_date" or "end_date" only, the function "normalizeStartDate" and "normalizeEndDate" facilitates  the time overlapping detection if "start_date" or "end_date" not set.

api/test/unit/validation/attributeValidator.test.js
```diff
@@ -13,31 +13,56 @@ describe('Validating Policy JSON properties',function(){
...
+        it('Should fail to validate the policy as instance_min_count is greater than instance_max_count without schedule defined',function(){
+          fakePolicy.schedules = null
+          attributeValidator.validatePolicy(fakePolicy,function(result){
+            expect(result[0]).to.have.property('stack').and.equal('instance_min_count 10 is higher or equal to instance_max_count 5 in policy_json');
+            expect(result[0]).to.have.property('message').and.equal('instance_min_count and instance_max_count values are not compatible');
+            expect(result[0]).to.have.property('property').and.equal('instance_min_count');
+          });
+        });
+
+        it('Should fail to validate the policy as instance_min_count is greater than instance_max_count without schedule defined',function(){
```
qibobo: without scaling_rules defined?

api/test/unit/validation/attributeValidator.test.js
```diff
@@ -13,31 +13,56 @@ describe('Validating Policy JSON properties',function(){
...
+      it('Should fail to validate the policy as instance_min_count is greater than instance_max_count with schedule defined',function(){
+          attributeValidator.validatePolicy(fakePolicy,function(result){
+            expect(result[0]).to.have.property('stack').and.equal('instance_min_count 10 is higher or equal to instance_max_count 5 in policy_json');
+            expect(result[0]).to.have.property('message').and.equal('instance_min_count and instance_max_count values are not compatible');
+            expect(result[0]).to.have.property('property').and.equal('instance_min_count');
+          });
+        });
+
+        it('Should fail to validate the policy as instance_min_count is greater than instance_max_count without schedule defined',function(){
+          fakePolicy.schedules = null
```
qibobo: do we need to test "undefined"?

cdlliuy: it is a regression case. 
we did have validate error when schedules is undefined previously.
These case didn't change, just reformatted with "context"

cdlliuy: change to "delete fakePolicy.schedules" to test the situation of "undefined"..

Also, add more testcase to validate the policy when properties are set to null.

boyang9527:
The travis failed with "The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over). The job has been terminated"

I guess this is caused by the schema validation...

qibobo:
LGTM

PR #309, opened 2018-03-02
Fix the inconsistent response from API server (internal/public)

cdlliuy:
#155646346. See tracker issue description for the proposed changes

### DIFF
--- a/api/app.js
+++ b/api/app.js
@@ -76,7 +76,7 @@ module.exports = function(settings, callback) {
         if(result.statusCode == HttpStatus.OK){
           next();
         }else if(result.statusCode == HttpStatus.NOT_FOUND){
-          res.status(HttpStatus.FORBIDDEN).send({"message": "The application is not bound to Auto-Scaling service"});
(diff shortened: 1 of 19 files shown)


api/test/unit/routes/policies.test.js
```diff
@@ -141,6 +145,7 @@ describe('Routing Policy Creation', function() {
           .set("Authorization", "fake-token")
           .send(fakePolicy)
           .end(function(error, result) {
+            console.log(result);
```
qibobo: remove it?

api/app.js
```diff
@@ -157,9 +155,7 @@ module.exports = function(settings, callback) {
     var errorResponse = {};
     if (err) {
       errorResponse = {
-        'success': false,
         'error': err,
-        'result': null
```
boyang9527: what is the reason why we have 'success' and 'result' defined in error message previously?

boyang9527: @qibobo  may be a question for you...

qibobo: @boyang9527 I did not change the result content of policy when I did the public api relevant PRs. It should be created in the origin commits of apiserver.

PR #310, opened 2018-03-04
circuit breaker for event generator

boyang9527:
[#151039644]

- add circuit breaker config
- use goboker when invoking scaling engine
- test cases for evaluator manager and evaluator
- corresponding changes in main and mian test

### DIFF
--- a/src/autoscaler/eventgenerator/config/config.go
+++ b/src/autoscaler/eventgenerator/config/config.go
@@ -13,18 +13,21 @@ import (
 )
 
 const (
-	DefaultLoggingLevel              string        = "info"
-	DefaultPolicyPollerInterval      time.Duration = 40 * time.Second
-	DefaultAggregatorExecuteInterval time.Duration = 40 * time.Second
-	DefaultMetricPollerCount         int           = 20
-	DefaultAppMonitorChannelSize     int           = 200
-	DefaultEvaluationExecuteInterval time.Duration = 40 * time.Second
-	DefaultEvaluatorCount            int           = 20
-	DefaultTriggerArrayChannelSize   int           = 200
-	DefaultLockTTL                   time.Duration = locket.DefaultSessionTTL
-	DefaultRetryInterval             time.Duration = locket.RetryInterval
-	DefaultDBLockRetryInterval       time.Duration = 5 * time.Second
-	DefaultDBLockTTL                 time.Duration = 15 * time.Second
+	DefaultLoggingLevel                   string        = "info"
+	DefaultPolicyPollerInterval           time.Duration = 40 * time.Second
+	DefaultAggregatorExecuteInterval      time.Duration = 40 * time.Second
+	DefaultMetricPollerCount              int           = 20
+	DefaultAppMonitorChannelSize          int           = 200
+	DefaultEvaluationExecuteInterval      time.Duration = 40 * time.Second
+	DefaultEvaluatorCount                 int           = 20
+	DefaultTriggerArrayChannelSize        int           = 200
+	DefaultLockTTL                        time.Duration = locket.DefaultSessionTTL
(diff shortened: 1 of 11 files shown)


qibobo:
@boyang9527 The circuit breaker relevant configuration should be added to exampleconfig/example.yml and integration test configuration.

src/autoscaler/eventgenerator/config/config.go
```diff
@@ -13,18 +13,20 @@ import (
...
+	DefaultMetricPollerCount              int           = 20
+	DefaultAppMonitorChannelSize          int           = 200
+	DefaultEvaluationExecuteInterval      time.Duration = 40 * time.Second
+	DefaultEvaluatorCount                 int           = 20
+	DefaultTriggerArrayChannelSize        int           = 200
+	DefaultLockTTL                        time.Duration = locket.DefaultSessionTTL
+	DefaultRetryInterval                  time.Duration = locket.RetryInterval
+	DefaultDBLockRetryInterval            time.Duration = 5 * time.Second
+	DefaultDBLockTTL                      time.Duration = 15 * time.Second
+	DefaultBreakerTimeout                 time.Duration = 10 * time.Minute
```
qibobo: 10 minutes as default is too long?

src/autoscaler/eventgenerator/config/config_test.go
```diff
@@ -155,7 +162,7 @@ db:
   policy_db_url: postgres://postgres:password@localhost/autoscaler?sslmode=disable
   app_metrics_db_url: postgres://postgres:password@localhost/autoscaler?sslmode=disable
 aggregator: 
-  aggregator_execute_interval: NOT-INTEGER-VALUE
+  aggregator_execute_interval: 5k
```
qibobo: why change the value? 5k is not as clear as NOT-INTEGER-VALUE.

src/autoscaler/eventgenerator/config/config_test.go
```diff
@@ -189,7 +196,7 @@ db:
   app_metrics_db_url: postgres://postgres:password@localhost/autoscaler?sslmode=disable
 aggregator: 
   aggregator_execute_interval: 30s
-  policy_poller_interval: NOT-INTEGER-VALUE
+  policy_poller_interval: 7u
```
qibobo: same as above.

src/autoscaler/eventgenerator/config/config_test.go
```diff
@@ -398,7 +405,7 @@ scalingEngine:
 metricCollector:
   metric_collector_url: http://localhost:8083
 lock:
-  lock_ttl: NOT-INTEGER-VALUE
+  lock_ttl: 9R
```
qibobo: same as above.

src/autoscaler/eventgenerator/config/config_test.go
```diff
@@ -434,7 +441,7 @@ metricCollector:
   metric_collector_url: http://localhost:8083
 lock:
   lock_ttl: 15s
-  lock_retry_interval: NOT-INTEGER-VALUE
+  lock_retry_interval: 8j
```
qibobo: same as above.

src/autoscaler/eventgenerator/config/config_test.go
boyang9527: it is used to check whether it can parse a time duration, not an integer. The value can be 5s/5m etc, they are not integer but valid time duration

src/autoscaler/eventgenerator/config/config.go
boyang9527: I will change to 5 minutes

boyang9527:
Per discussion with @paltanmoy, I will change the breaker impl to github.com/rubyist/circuitbreaker which has the backoff timeout interval support.

PR #312, opened 2018-03-08
As an operator, I expect to do service registry in deployment rather …

qibobo:
…than using consul in components' code

[#155766592]


boyang9527:
Let us confirm the change in team meeting as it is important change.

boyang9527:
looks like consul is removed completely, don't we have consul lock in scheduler?

boyang9527:
@qibobo  can you add consul lock for scheduler?

boyang9527:
LGTM ... while we need another PR to add lock to scheduler

PR #317, opened 2018-03-20
Adopted OSB API by adding service-binding schema

paltanmoy:
#156123844


boyang9527:
@paltanmoy  how we can validate the schema is correct?

paltanmoy:
@boyang9527 Schema field is used in the catalog definition to declare the configuration parameters the service accepts for creating a service binding. It's an way to open up the configuration parameters to the user so that they are aware of the policyjson structure before binding the service instance. As the schema is validated by the APIServer, it’s expected from the operators to define correct schema through deployment manifest. But it makes sense to have a test case to validate whether the policyjson conforms with the schema or not.

cdlliuy:
@paltanmoy ,  
It is nice to have a schema definition catalog.  
But if the schema has small differences with the one defined api server, then it is hard to detect the difference without an identical test coverage.
From this point of view, I think current invalid cases are not sufficient ...   I think we need go through at least the full schema test coverage as API server instead of just checking min/max_instance_count

paltanmoy:
@cdlliuy Though the entire schema is being validated in APIServer but I agree, it would be safer to have another layer of validation here.  Added tests for schema validation.

cdlliuy:
LGTM

boyang9527:
LGTM

PR #318, opened 2018-03-26
The eventgenerator should not send trigger alarm when the appmetrics …

qibobo:
…of an application  do not cover a whole breach duration.

[[#156261863]]

### DIFF
--- a/src/autoscaler/eventgenerator/generator/evaluator_test.go
+++ b/src/autoscaler/eventgenerator/generator/evaluator_test.go
@@ -145,7 +145,20 @@ var _ = Describe("Evaluator", func() {
 					Eventually(database.RetrieveAppMetricsCallCount).Should(Equal(1))
 				})
 			})
-
+			Context("when the appMetrics are not enough", func() {
+				BeforeEach(func() {
+					scalingEngine.RouteToHandler("POST", urlPath, ghttp.RespondWith(http.StatusOK, "successful"))
+					Expect(triggerChan).To(BeSent(triggerArrayGT))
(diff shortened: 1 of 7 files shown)


src/autoscaler/eventgenerator/generator/evaluator_test.go
```diff
@@ -315,26 +155,59 @@ var _ = Describe("Evaluator", func() {
...
 					Context("when the appMetrics breach the trigger", func() {
-						BeforeEach(func() {
-							database.RetrieveAppMetricsStub = func(appId string, metricType string, start int64, end int64) ([]*models.AppMetric, error) {
-								return appMetricGTBreach, nil
-							}
-						})
-						It("should send trigger alarm to scaling engine", func() {
-							Eventually(scalingEngine.ReceivedRequests).Should(HaveLen(1))
-							Eventually(logger.LogMessages).Should(ContainElement(ContainSubstring("send trigger alarm to scaling engine")))
+						Context("when the appMetrics are enough", func() {
```
boyang9527: we do not need to test not enough metrics for each operator.  The test case with not enough metrics should not be under "when the appMetrics brach the trigger"

qibobo:
@boyang9527 updated.

boyang9527:
I am thinking perhaps we can just retrieve metrics in 2*aggregation_interval + breach_duration  as opposed to 2* breach_duration, this will prevent too many metrics if breach_duration is very large.. Anyway this could be another PR for optimization

cdlliuy:
LGTM

PR #319, opened 2018-03-28
Insert appmetrics to database by batch

qibobo:
[#156261970]


src/autoscaler/db/sqldb/sqldb_suite_test.go
```diff
@@ -100,9 +100,9 @@ func cleanAppMetricTable() {
 	}
 }
 
-func hasAppMetric(appId, metricType string, timestamp int64) bool {
-	query := "SELECT * FROM app_metric WHERE app_id = $1 AND metric_type = $2 AND timestamp = $3"
-	rows, e := dbHelper.Query(query, appId, metricType, timestamp)
+func hasAppMetric(appId, metricType string, timestamp int64, value string) bool {
+	query := "SELECT * FROM app_metric WHERE app_id = $1 AND metric_type = $2 AND timestamp = $3 and value = $4"
```
cdlliuy: how about using "AND"?

cdlliuy:
LGTM

src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -14,21 +15,28 @@ type Aggregator struct {
 	appChan                   chan *models.AppMonitor
 	metricPollerArray         []*MetricPoller
 	cclock                    clock.Clock
+	saveClock                 clock.Clock
```
boyang9527: where saveClock is used?  why not have one single clock?

src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -80,3 +89,24 @@ func (a *Aggregator) startWork() {
...
+	for {
+		select {
+		case <-a.doneChan:
+			return
+		case appMetric := <-a.appMetricChan:
+			appMetricArray = append(appMetricArray, appMetric)
+		case <-ticker.C():
+			go func(appMetricDB db.AppMetricDB, metrics []*models.AppMetric) {
+				appMetricDB.SaveAppMetricsInBulk(metrics)
+				metrics = nil
```
boyang9527: why set metrics to be nil?

src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -80,3 +89,24 @@ func (a *Aggregator) startWork() {
...
+				return
+			}(a.appMetricDB, appMetricArray)
+			appMetricArray = nil
```
boyang9527: appMetricArray = []*models.AppMetric{} ?

src/autoscaler/eventgenerator/aggregator/aggregator_test.go
```diff
@@ -51,50 +58,73 @@ var _ = Describe("Aggregator", func() {
...
 
 		clock = fakeclock.NewFakeClock(time.Now())
+		saveClock = fakeclock.NewFakeClock(time.Now())
 		logger = lager.NewLogger("Aggregator-test")
 
 		appMonitorsChan = make(chan *models.AppMonitor, 10)
-		if testEvaluateInteval > testAggregatorExecuteInterval {
-			fakeWaitDuration = testEvaluateInteval
-		} else {
+		if testEvaluateInterval > fakeWaitDuration {
```
boyang9527: what testEvaluateInterval is used for?

src/autoscaler/eventgenerator/config/config.go
```diff
@@ -16,8 +16,10 @@ const (
 	DefaultLoggingLevel                   string        = "info"
 	DefaultPolicyPollerInterval           time.Duration = 40 * time.Second
 	DefaultAggregatorExecuteInterval      time.Duration = 40 * time.Second
+	DefaultSaveInterval                   time.Duration = 30 * time.Second
```
boyang9527: the default values looks to long, we need to make the aggregated data available as soon as possible

src/autoscaler/eventgenerator/aggregator/aggregator.go
qibobo: @boyang9527  After "metrics" has been saved to db, it should be set to nil so that gc can work.

boyang9527: metrics is passed in with appMetricArray. You don't need to set to nil

boyang9527:
LGTM

PR #326, opened 2018-04-04
When using percentage as the adjustment ,  scaling engine fail to scale out if current instance number is small.

cdlliuy:
#156494573

### DIFF
--- a/src/autoscaler/scalingengine/scalingengine.go
+++ b/src/autoscaler/scalingengine/scalingengine.go
@@ -4,7 +4,6 @@ import (
 	"autoscaler/cf"
 	"autoscaler/db"
 	"autoscaler/models"
-	"math"
 
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


src/autoscaler/scalingengine/scalingengine.go
```diff
@@ -179,7 +180,12 @@ func (s *scalingEngine) ComputeNewInstances(currentInstances int, adjustment str
 			s.logger.Error("failed-to-parse-percentage", err, lager.Data{"adjustment": adjustment})
 			return -1, err
 		}
-		newInstances = int(float64(currentInstances)*(1+percentage/100) + 0.5)
+		if percentage >= 0 {
```
boyang9527: I think this is different with what we discussed.  We need to add/delete at least 1 instance, so the logic is 

if newInstances == currentInstances 
     if precentage > 0, then newInstances = currentInstances + 1
     if percentage < 0, then newInstances = currentIntances -1

src/autoscaler/scalingengine/scalingengine.go
```diff
@@ -180,10 +179,14 @@ func (s *scalingEngine) ComputeNewInstances(currentInstances int, adjustment str
...
 			return -1, err
 		}
-		if percentage >= 0 {
-			newInstances = int(math.Ceil(float64(currentInstances) * (1 + percentage/100)))
-		} else {
-			newInstances = int(math.Floor(float64(currentInstances) * (1 + percentage/100)))
+		newInstances = int(float64(currentInstances)*(1+percentage/100) + 0.5)
+
+		if newInstances == currentInstances {
+			if percentage >= 0 {
```
boyang9527: if percentage == 0, then we don't need to scale

cdlliuy:
@boyang9527 , fixed. anyway, +0% is invalid per policy validation, so I didn't add new test case for adjustment = +0% in scaling engine

boyang9527:
LGTM

PR #329, opened 2018-04-10
Enhanced error details for scaling failure and added log trace for ci…

paltanmoy:
…rcuit trip

[#156649637]

### DIFF
--- a/src/autoscaler/scalingengine/scalingengine.go
+++ b/src/autoscaler/scalingengine/scalingengine.go
@@ -78,7 +78,7 @@ func (s *scalingEngine) Scale(appId string, trigger *models.Trigger) (*models.Ap
 	if err != nil {
 		logger.Error("failed-to-get-app-info", err)
 		history.Status = models.ScalingStatusFailed
-		history.Error = "failed to get app info"
+		history.Error = "failed to get app info: " + err.Error()
 		return nil, err
 	}
 	history.OldInstances = appEntity.Instances
@@ -162,17 +162,17 @@ func (s *scalingEngine) Scale(appId string, trigger *models.Trigger) (*models.Ap
 		result.Status = history.Status
 		result.Adjustment = 0
 		result.CooldownExpiredAt = now.Add(trigger.CoolDown(s.defaultCoolDownSecs)).UnixNano()
-		return result,nil
+		return result, nil
 	}
-	
+
 	err = s.cfClient.SetAppInstances(appId, newInstances)
 	if err != nil {
 		logger.Error("failed-to-set-app-instances", err, lager.Data{"newInstances": newInstances})
 		history.Status = models.ScalingStatusFailed
-		history.Error = "failed to set app instances"
+		history.Error = "failed to set app instances: " + err.Error()
 		return nil, err
 	}
-	
+
 	history.Status = models.ScalingStatusSucceeded
 	result.Status = history.Status
 	result.Adjustment = newInstances - appEntity.Instances
@@ -196,9 +196,9 @@ func (s *scalingEngine) ComputeNewInstances(currentInstances int, adjustment str
 		newInstances = int(float64(currentInstances)*(1+percentage/100) + 0.5)
 
 		if newInstances == currentInstances {
-			if percentage > 0 { 
+			if percentage > 0 {
 				newInstances = currentInstances + 1
-			} else if percentage < 0 { 
+			} else if percentage < 0 {
 				newInstances = currentInstances - 1
 			}
 		}
@@ -256,7 +256,7 @@ func (s *scalingEngine) SetActiveSchedule(appId string, schedule *models.ActiveS
 	if err != nil {
 		logger.Error("failed-to-get-app-info", err)
 		history.Status = models.ScalingStatusFailed
-		history.Error = "failed to get app info"
+		history.Error = "failed to get app info: " + err.Error()
 		return err
 	}
 	history.OldInstances = appEntity.Instances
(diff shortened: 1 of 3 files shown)


src/autoscaler/scalingengine/scalingengine.go
```diff
@@ -72,7 +72,7 @@ func (s *scalingEngine) Scale(appId string, trigger *models.Trigger) (int, error
 	if err != nil {
 		logger.Error("failed-to-get-app-info", err)
 		history.Status = models.ScalingStatusFailed
-		history.Error = "failed to get app info"
+		history.Error = "failed to get app info: " + err.Error()
```
boyang9527: given the history will be exposed to user, not sure whether it is good to show the detailed / internal errors

src/autoscaler/cf/app.go
```diff
@@ -83,10 +84,22 @@ func (c *cfClient) SetAppInstances(appId string, num int) error {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusCreated {
-		err = fmt.Errorf("failed setting application instances: %s [%d] %s", url, resp.StatusCode, resp.Status)
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			c.logger.Error("failed to read response body while setting app instance", err)
```
boyang9527: add app id info to log.  
It is better the first argument does not include any space, so that index can be built on the whole string.

src/autoscaler/cf/app.go
```diff
@@ -83,10 +84,22 @@ func (c *cfClient) SetAppInstances(appId string, num int) error {
...
+			return err
+		}
+		var bodydata map[string]interface{}
+		err = json.Unmarshal([]byte(respBody), &bodydata)
+		if err != nil {
+			c.logger.Error("failed to unmarshal response body while setting app instance", err)
```
boyang9527: same as above

src/autoscaler/scalingengine/scalingengine.go
cdlliuy: Yes, similar concern.. Autoscaler cli display every field of scaling history as a column in table. Then, with the detailed error msg, the error column will be quite long ..

src/autoscaler/eventgenerator/generator/evaluator.go
```diff
@@ -128,8 +128,8 @@ func (e *Evaluator) doEvaluate(triggerArray []*models.Trigger) {
 		if isBreached {
 			trigger.MetricUnit = appMetricList[0].Unit
 			e.logger.Info("send trigger alarm to scaling engine", lager.Data{"trigger": trigger})
-
 			if appBreaker := e.getBreaker(trigger.AppId); appBreaker != nil {
+				e.logger.Info("circuit-tripped", lager.Data{"appId": trigger.AppId, "consecutiveFailures": appBreaker.ConsecFailures()})
```
cdlliuy: how about do the logging only when  ConsecFailures() != 0 ?

src/autoscaler/scalingengine/scalingengine.go
paltanmoy: @boyang9527  Scaling can fail for many reasons, user should be informed about the reason so that they can take necessary action to resolve such failure if possible. ex. failure due to any quota related issue like SpaceQuotaMemoryLimitExceeded or SpaceQuotaInstanceMemoryLimitExceeded or SpaceQuotaTotalRoutesExceeded. Until and unless user is not informed about the root cause of such failure it will be difficult to handle such.

src/autoscaler/eventgenerator/generator/evaluator.go
paltanmoy: @cdlliuy Updated

src/autoscaler/scalingengine/scalingengine.go
paltanmoy: @boyang9527  It makes sense not to expose all the errors ( specially the internal errors) to the users. So updated the PR by removing the internal errors.

src/autoscaler/cf/app_test.go
```diff
@@ -155,7 +155,7 @@ var _ = Describe("App", func() {
 			BeforeEach(func() {
 				responseMap := make(map[string]interface{})
 				responseMap["description"] = "You have exceeded the total routes for your space's quota"
-				responseMap["error_code"] = "SpaceQuotaTotalRoutesExceeded"
+				responseMap["error_code"] = "SpaceQuotaMemoryLimitExceeded"
```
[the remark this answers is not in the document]
paltanmoy: Fixed

cdlliuy:
@paltanmoy , LGTM, but the travis build seems hang ..

src/autoscaler/cf/app_test.go
```diff
@@ -153,15 +153,18 @@ var _ = Describe("App", func() {
 
 		Context("when updating app instances returns non-200 status code", func() {
 			BeforeEach(func() {
+				responseMap := make(map[string]interface{})
+				responseMap["description"] = "You have exceeded the instance limit for your organization's quota."
+				responseMap["error_code"] = "QuotaInstanceLimitExceeded"
 				fakeCC.AppendHandlers(
 					ghttp.CombineHandlers(
-						ghttp.RespondWithJSONEncoded(http.StatusNotFound, ""),
```
qibobo: Is it better that we keep this removed test case with the response which can not be unmarshalled?

paltanmoy: We already have such test for "when cloud controller is not reachable".

boyang9527:
LGTM   @paltanmoy  pls resolve the conflict

paltanmoy:
it seems the branch is corrupted. I will create a separate PR soon.

PR #330, opened 2018-04-11
Remove cli from app-autoscaler repo and add cli-repo description in readme

cdlliuy:


.gitmodules
```diff
@@ -73,9 +73,6 @@
 [submodule "src/code.cloudfoundry.org/cli"]
```
rohitsharma04: code.cloudfoundry.org/cli
github.com/jessevdk/go-flags
can also be removed.

qibobo:
LGTM

PR #331, opened 2018-04-11
add omitempty for optional parameters in ScalingPolicy Model and Schedules in ScalingPolicy Model

rohitsharma04:
[https://github.com/cloudfoundry-incubator/app-autoscaler-cli-plugin/pull/5#issuecomment-380366376](https://github.com/cloudfoundry-incubator/app-autoscaler-cli-plugin/pull/5#issuecomment-380366376)
This also fixes a bug which occurs when cli plugin is installed from this repo.
`cf autoscaling-policy` doesn't show schedules in policy.

@cdlliuy @boyang9527 
I think we should remove the cli plugin from this repo ?


cdlliuy:
@rohitsharma04 ,  we are on the same page, see PR #330

cdlliuy:
LGTM

paltanmoy:
LGTM

PR #339, opened 2018-05-16
Add health point with prometheus client lib on scaling engine

cdlliuy:
#156684322


src/autoscaler/scalingengine/healthendpoint/registrar.go
```diff
@@ -0,0 +1,76 @@
+package healthendpoint
```
boyang9527: move this package to a common folder

src/autoscaler/models/health.go
```diff
@@ -0,0 +1,7 @@
+package models
+
```
boyang9527: move this interface definition to a common folder for health

src/autoscaler/db/sqldb/scheduler_sqldb.go
```diff
@@ -86,3 +88,11 @@ func (sdb *SchedulerSQLDB) GetActiveSchedules() (map[string]*models.ActiveSchedu
 	return schedules, nil
 
 }
+
+func (sdb *SchedulerSQLDB) EmitHealthMetrics(h models.Health, interval time.Duration) {
```
boyang9527: pass in a clock interface  so that  you can move the timer to test

src/autoscaler/scalingengine/server/scaling_handler.go
```diff
@@ -19,17 +19,22 @@ type ScalingHandler struct {
...
 	return &ScalingHandler{
 		logger:          logger.Session("scaling-handler"),
 		scalingEngineDB: scalingEngineDB,
 		scalingEngine:   scalingEngine,
+		health:          health,
 	}
 }
 
 func (h *ScalingHandler) Scale(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	h.health.Inc("concurrentHTTPReq")
```
boyang9527: do we want to collect separately for request coming from eventgen and scheduler?

boyang9527:
test  needs to be enhanced for EmitHealthMetrics(), while I would give it a pass and let us have another PR to do that.

LGTM

cdlliuy:
@boyang9527 , the  test for EmitHealthMetric() was added into each *_sqldb_test.  For example: 
https://github.com/cdlliuy/app-autoscaler/blob/dev_healthendpoint/src/autoscaler/db/sqldb/policy_sqldb_test.go#L234-L255

boyang9527:
I mean the  EmitHealthMetrics() can pass in a Clock interface so that you can use fakeclock to control the timer for testing, instead of waiting for the timer to time out

cdlliuy:
@boyang9527  , updated for EmmitHealthMetrics

qibobo:
LGTM

PR #340, opened 2018-05-22
[#157651220] Added table level lock during database lock fetch

paltanmoy:


src/autoscaler/metricscollector/cmd/metricscollector/metricscollector_test.go
```diff
@@ -357,6 +357,26 @@ var _ = Describe("MetricsCollector", func() {
...
+				secondRunner.configPath = writeConfig(&consulConfig).Name()
+				secondRunner.Start()
+			})
+
+			AfterEach(func() {
+				secondRunner.KillWithFire()
+			})
+
+			It("Only one instance should get the lock", func() {
+				Eventually(func() int { return runner.GetLockDetails() }, 5*time.Second, 1*time.Second).Should(Equal(1))
```
boyang9527: query the number of rows in the lock table looks not a good way to test this behavior, it does not necessarily indicate actually only one instance get the lock.  You may check with the logs to make sure there is exactly one instance started.

paltanmoy: @boyang9527  Any instance can get the lock and other keep on retrying. But which one will get the lock is not certain and can be changed in each test run. I was looking for some way to test if either of this two scenarios occurs - 
```
Consistently(firstRunner.Session.Buffer, 5*time.Second).Should(gbytes.Say("eventgenerator.lock-acquired-in-first-attempt"))
And				
Consistently(secondRunner.Session.Buffer, 5*time.Second).ShouldNot(gbytes.Say("eventgenerator.lock-acquired-in-first-attempt"))
```
OR 
```
Consistently(secondRunner.Session.Buffer, 5*time.Second).Should(gbytes.Say("eventgenerator.lock-acquired-in-first-attempt"))
And				
Consistently(firstRunner.Session.Buffer, 5*time.Second).ShouldNot(gbytes.Say("eventgenerator.lock-acquired-in-first-attempt"))
```

Did not get a good way to achieve this using Gomega. Any suggestion here ?

cdlliuy: @paltanmoy , I handled similar situation in my scaling engine db Lock PR (not raised yet)  as:
https://github.com/cdlliuy/app-autoscaler/blob/dev_dblock_scalingengine/src/autoscaler/scalingengine/cmd/scalingengine/scalingengine_test.go#L106-L169

paltanmoy: @boyang9527  Updated the test cases.

PR #343, opened 2018-05-28
 Enable db lock in scaling engine

cdlliuy:
#157727157

### DIFF
--- a/src/autoscaler/scalingengine/cmd/scalingengine/main.go
+++ b/src/autoscaler/scalingengine/cmd/scalingengine/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"autoscaler/helpers"
+	"autoscaler/sync"
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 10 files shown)


src/autoscaler/scalingengine/cmd/scalingengine/main.go
```diff
@@ -98,18 +99,47 @@ func main() {
...
+		defer lockDB.Close()
+		sedl := sync.NewDatabaseLock(logger)
+		dbLockMaintainer := sedl.InitDBLockRunner(conf.DBLock.LockRetryInterval, conf.DBLock.LockTTL, guid, lockDB)
+		lockMembers = append(grouper.Members{{"db-lock-maintainer", dbLockMaintainer}}, lockMembers...)
+	}
+
+	nonLockMonitor := ifrit.Invoke(sigmon.New(grouper.NewOrdered(os.Interrupt, nonLockMembers)))
+	lockMonitor := ifrit.Invoke(sigmon.New(grouper.NewOrdered(os.Interrupt, lockMembers)))
 
 	logger.Info("started")
```
boyang9527: "started" will not show if synchronizer does not get the lock

src/autoscaler/scalingengine/cmd/scalingengine/main.go
```diff
@@ -98,18 +99,47 @@ func main() {
...
 
-	err = <-monitor.Wait()
+	err = <-nonLockMonitor.Wait()
```
boyang9527: this is not correct, you should watch on two channels using select

cdlliuy:
By watching the 2 process->wait() with select,  I must add a defer implementation to release db lock before close it.  If I don't do so,  when os.interrupt happens,  the whole process may exit and db connection  may close before db lock is  released.

src/autoscaler/scalingengine/cmd/scalingengine/main.go
boyang9527: I mean you should show "started" even the synchronizer does not get the lock

boyang9527:
LGTM

src/autoscaler/scalingengine/cmd/scalingengine/main.go
```diff
@@ -98,19 +99,53 @@ func main() {
...
+	}
 
+	nonLockMonitor := ifrit.Invoke(sigmon.New(grouper.NewOrdered(os.Interrupt, nonLockMembers)))
 	logger.Info("started")
 
-	err = <-monitor.Wait()
-	if err != nil {
-		logger.Error("exited-with-failure", err)
-		os.Exit(1)
+	lockMonitor := ifrit.Invoke(sigmon.New(grouper.NewOrdered(os.Interrupt, lockMembers)))
```
boyang9527: there is still issue  for the instance that does not have the lock to run synchronizer.  It will hold in line 136,  and will not be able to monitor the errors happened in non-locked processes.

pradyutsarma:
LGTM. Let's ensure that we don't create any more singleton processes and either work towards consolidating the existing ones into one singleton app or remove them (more preferable).

boyang9527:
There is still issue. See my comment at line 136.  Given it is already merged. Let us fix it with another PR

PR #354, opened 2018-06-27
remove stat_window setting from API server, service broker and doc

boyang9527:
[#158573473]


boyang9527:
@rohitsharma04  done, pls check whether there is still anything missing.

PR #361, opened 2018-07-11
Added custom-metrics metrics-forwarder components

paltanmoy:
Custom Metrics feature related changes[MetricsForwarder component added].


paltanmoy:
Merge dependency on : https://github.com/cloudfoundry-incubator/app-autoscaler/pull/360

paltanmoy:
Tracker Story #157900428

src/autoscaler/cf/client.go
```diff
@@ -92,9 +95,24 @@ func NewCFClient(conf *CFConfig, logger lager.Logger, clk clock.Clock) CFClient
...
+	uaaClient, err := uaago.NewClient(conf.UAAEndpoint)
+	if err != nil {
+		c.logger.Error("Failed to create uaago client:", err)
+	}
+	c.uaaClient = uaaClient
+
 	return c
 }
 
+func (c *cfClient) GetAuthToken() (string, error) {
```
cdlliuy: It seems that the function is not called by anyone. Unnecessary?

src/autoscaler/cf/config.go
```diff
@@ -55,5 +56,26 @@ func (conf *CFConfig) Validate() error {
 			return fmt.Errorf("Configuration error: client_id is empty")
 		}
 	}
+	if conf.UAAEndpoint != "" {
```
cdlliuy: why put uaa endpoint in config when it should be get from CC API /info?

src/autoscaler/cf/client.go
paltanmoy: @cdlliuy Its being used during reading from firehose in metricscollector. 

https://github.com/cloudfoundry-incubator/app-autoscaler/pull/362/files#diff-82c5b923bf774ec1cce41431344fe4ceR103

src/autoscaler/routes/routes.go
```diff
@@ -28,12 +28,16 @@ const (
 
 	SyncActiveSchedulesPath      = "/v1/syncSchedules"
 	SyncActiveSchedulesRouteName = "SyncActiveSchedules"
+
+	CustomMetricsPath          = "/v1/{appid}/metrics"
```
cdlliuy: how about using /v1/apps/{appId}/metrics to align with the prefix of other APIs .

src/autoscaler/cf/client.go
paltanmoy: Removed the function from here and will put it on PR #362

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+	if err != nil {
+		mh.logger.Error("error-decoding-authorizaion-header", err, lager.Data{"authheader": auth})
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error decoding credentials"})
+		return
+	}
+
+	appID := vars["appid"]
```
cdlliuy: how about moving this line just before credentialCache.Get(appID)

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+		handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+			Code:    "Authorization-Failure-Error",
+			Message: "Basic authorization credential is not used properly"})
+		return
+	}
+
+	username := pair[0]
+	password := pair[1]
+	var isValid bool
+	var credentials models.CustomMetricCredentials
```
cdlliuy: if we put the hashed "username:password" directly in credential database, we won't need to handle username and password separately for each request.   In service instance 's credential information, the username is meaningless in fact.

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -145,3 +146,41 @@ func (pdb *PolicySQLDB) DeletePolicy(appId string) error {
...
+	err := pdb.sqldb.QueryRow(query, appId).Scan(&username, &password)
+
+	if err != nil {
+		pdb.logger.Error("get-custom-metrics-creds-from-credentials-table", err, lager.Data{"query": query})
+		return "", "", err
+	}
+	return username, password, nil
+}
+
+func (pdb *PolicySQLDB) ValidateCustomMetricTypes(appGUID string, metricsConsumer *models.MetricsConsumer) (bool, error) {
```
cdlliuy: why put ValidateCustomMetricTypes here?  No SQL statement in this implementation.

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -145,3 +146,41 @@ func (pdb *PolicySQLDB) DeletePolicy(appId string) error {
...
+	scalingPolicy, err := pdb.GetAppPolicy(appGUID)
+	if err != nil {
+		pdb.logger.Error("errorr-getting-policy", err, lager.Data{"appId": appGUID})
+		return false, errors.New("not able to get policy details")
+	}
+	if err == nil && scalingPolicy == nil {
+		pdb.logger.Debug("no-policy-found", lager.Data{"appId": appGUID})
+		return false, errors.New("no policy found")
+	}
+	allowedMetricTypeSet := make(map[string]struct{}, len(scalingPolicy.ScalingRules))
```
cdlliuy: If my memory is right, we discussed to save "memoryused" "memoryutil" "responsetime" and "throughput" as the reserved keywords. it is necessary  to reject these metrics type as well

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -145,3 +146,41 @@ func (pdb *PolicySQLDB) DeletePolicy(appId string) error {
...
+	for _, metrictype := range scalingPolicy.ScalingRules {
+		allowedMetricTypeSet[metrictype.MetricType] = struct{}{}
+	}
+	pdb.logger.Debug("allowed-metrics-types", lager.Data{"metrics": allowedMetricTypeSet})
+	for _, metric := range metricsConsumer.CustomMetrics {
```
cdlliuy: Also, I guess we need to cache the allowed metric type for an application to avoid query & parsing policy on each request

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		handlers.WriteJSONResponse(w, http.StatusBadRequest, models.ErrorResponse{
+			Code:    "Bad-Request",
+			Message: "Error validating custom metrics type"})
+		return
+	}
+
+	metrics := mh.parseMetrics(metricsConsumer)
+
+	if len(metrics) <= 0 {
```
cdlliuy: how about moving this len() check prior to ValidateCustomMetricTypes()?   I assume Len(metrics) ==0 only happens when len(metricsConsumer.CustomMetrics) ==0

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+	err = json.Unmarshal(body, &metricsConsumer)
+	if err != nil {
+		mh.logger.Error("error-unmarshaling-metrics", err, lager.Data{"body": r.Body})
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error unmarshaling custom metrics request body"})
+		return
+	}
+	isValidRequest, err := mh.policyDB.ValidateCustomMetricTypes(appID, metricsConsumer)
```
boyang9527: same as above, we don't need to validate whether there is rule defined for the metric, we only need to make sure it is not using predefined metric names.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+	isUsernameAuthenticated := bcrypt.CompareHashAndPassword([]byte(usernameHash), []byte(username))
+	isPasswordAuthenticated := bcrypt.CompareHashAndPassword([]byte(passwordHash), []byte(password))
+	if isPasswordAuthenticated == nil && isUsernameAuthenticated == nil { // password matching successfull
+		return true
+	}
+	mh.logger.Debug("failed-to-authorize-credentials")
+	return false
+}
+
+func (mh *CustomMetricsHandler) parseMetrics(metricsConsumer *models.MetricsConsumer) []*models.CustomMetric {
```
boyang9527: this is actually not "parse", you may use "getMetrics"

src/autoscaler/models/custom_metrics.go
```diff
@@ -0,0 +1,21 @@
+package models
+
+type CustomMetric struct {
+	Name          string  `json:"name"`
+	Type          string  `json:"type"`
```
boyang9527: what is type, looks like it is not used when emitting ...

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error reading custom metrics request body"})
+		return
+	}
+	var metricsConsumer *models.MetricsConsumer
+	err = json.Unmarshal(body, &metricsConsumer)
+	if err != nil {
+		mh.logger.Error("error-unmarshaling-metrics", err, lager.Data{"body": r.Body})
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
paltanmoy: Removed

src/autoscaler/db/db.go
```diff
@@ -43,6 +43,8 @@ type PolicyDB interface {
 	Close() error
 	EmitHealthMetrics(h healthendpoint.Health, cclock clock.Clock, interval time.Duration)
 	DeletePolicy(appId string) error
+	GetCustomMetricsCreds(appId string) (string, string, error)
```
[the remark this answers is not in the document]
paltanmoy: Yes, we will use a seperate CredentialDB. But will create a separate PR once this PR is merged

src/autoscaler/metricsforwarder/cmd/metricsforwarder/main.go
```diff
@@ -0,0 +1,116 @@
...
+
+	err = <-monitor.Wait()
+	if err != nil {
+		logger.Error("exited-with-failure", err)
+		os.Exit(1)
+	}
+	logger.Info("exited")
+}
+
+func initLoggerFromConfig(logLevelConfig string) lager.Logger {
```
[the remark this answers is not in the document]
paltanmoy: Updated using helper method

src/autoscaler/metricsforwarder/cmd/metricsforwarder/metricsforwarder_test.go
```diff
@@ -0,0 +1,135 @@
...
+		It("should stop", func() {
+			runner.Session.Interrupt()
+			Eventually(runner.Session, 5).Should(Exit(0))
+		})
+
+	})
+
+	Describe("MetricsForwarder REST API", func() {
+		Context("when a request with custom metrics comes", func() {
+			Context("when using polling for metrics collection", func() {
```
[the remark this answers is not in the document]
paltanmoy: Fixed typo

src/autoscaler/metricsforwarder/forwarder/metric_forwarder.go
```diff
@@ -0,0 +1,64 @@
...
+func (mf *metricForwarder) EmitMetric(metric *models.CustomMetric) {
+	mf.logger.Debug("custom-metric-emit-request-received:", lager.Data{"metric": metric})
+
+	gauge_tags := map[string]string{
+		"applicationGuid":     metric.AppGUID,
+		"applicationInstance": fmt.Sprint(metric.InstanceIndex),
+	}
+	options := []loggregator.EmitGaugeOption{
+		loggregator.WithGaugeAppInfo(metric.AppGUID, 0),
+		loggregator.WithEnvelopeTags(gauge_tags),
```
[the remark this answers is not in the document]
paltanmoy: Removed Envelope Tags.  Now metrics will have following format - 

```
origin:"autoscaler_metrics_forwarder" eventType:ValueMetric timestamp:1537793107625432469 deployment:"app-autoscaler" job:"metricsforwarder" index:"669ebefe-92e8-44ba-bf2e-f7ad698445e1" ip:"10.244.0.147" tags:<key:"instance_id" value:"0" > tags:<key:"source_id" value:"9027969b-1784-4fa5-a112-338f1eee1ca9" > valueMetric:<name:"jobqueue" value:200 unit:"unit" >

```

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Renamed to getMetrics

src/autoscaler/models/custom_metrics.go
paltanmoy: As we are only supporting "Gauge" metrics so there is no need to ask user to pass the Metrics Type. Removed the Type info

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+		policyDB:        policyDB,
+		credentialCache: credentialCache,
+		cacheTTL:        cacheTTL,
+		logger:          logger,
+	}
+}
+
+func (mh *CustomMetricsHandler) PublishMetrics(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	w.Header().Set("Content-Type", "application/json")
+	auth := strings.SplitN(r.Header.Get("Authorization"), " ", 2)
```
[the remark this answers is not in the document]
paltanmoy: used this nice handy API.

src/autoscaler/db/sqldb/policy_sqldb.go
paltanmoy: @boyang9527 , @cdlliuy  IMHO it's not a good idea to allow user to emit any custom metrics through autoscaler without defining the metrics in the policy.

Could you please explain the use case we are targeting here ?

@pradyutsarma  : Your thoughts ?

src/autoscaler/db/sqldb/policy_sqldb.go
paltanmoy: Removed to custom_metrics_handlers.go

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: As discussed during the Credential Schema design phase, we will have username and password separate. As this proposal will not have much impact in general and the schema is already finalised I would like to skip this.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Done

src/autoscaler/metricsforwarder/config/config_test.go
```diff
@@ -0,0 +1,236 @@
...
+			BeforeEach(func() {
+				conf.Db.PolicyDb.URL = ""
+			})
+
+			It("should error", func() {
+				Expect(err).To(MatchError(MatchRegexp("Configuration error: Policy DB url is empty")))
+			})
+		})
+
+		Context("when policy db url is not set", func() {
```
qibobo: duplicated with the previous test case.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,193 @@
...
+	if isPasswordAuthenticated == nil && isUsernameAuthenticated == nil { // password matching successfull
+		return true
+	}
+	mh.logger.Debug("failed-to-authorize-credentials")
+	return false
+}
+
+func (mh *CustomMetricsHandler) validateCustomMetricTypes(appGUID string, metricsConsumer *models.MetricsConsumer) (bool, error) {
+	standardMetricsTypes := make(map[string]struct{})
+	standardMetricsTypes["memoryused"] = struct{}{}
```
qibobo: These types should be constant?

paltanmoy:
@boyang9527 , @cdlliuy , @qibobo  any more comments ?

src/autoscaler/metricsforwarder/config/config.go
```diff
@@ -0,0 +1,81 @@
...
+
+	return conf, nil
+}
+
+func (c *Config) Validate() error {
+
+	if c.Db.PolicyDb.URL == "" {
+		return fmt.Errorf("Configuration error: Policy DB url is empty")
+	}
+
```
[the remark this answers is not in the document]
qibobo: @paltanmoy  I have the same concern as Bo @boyang9527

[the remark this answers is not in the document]
paltanmoy: yes LoggregatorConfig is mandatory to connect to Doppler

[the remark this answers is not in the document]
boyang9527: then you need validate, right?

src/autoscaler/cf/uaa_client.go
```diff
@@ -0,0 +1,43 @@
...
+		uc.logger.Error("failed-to-create-uaago-client", err)
+		os.Exit(1)
+	}
+	uc.uaaClient = uaaClient
+
+	return uc
+}
+
+func (uc *uaaClient) RefreshAuthToken() (string, error) {
+	token, err := uc.uaaClient.GetAuthToken(uc.conf.ClientID, uc.conf.Secret, uc.conf.SkipSSLValidation)
```
cdlliuy: autoscaler has an implementation of request auth token in 
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/master/src/autoscaler/cf/client.go#L124-L156
which will do post request to "/oauth/token" to get granted as well.  

Why we need to do a uaa_client.go here?

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,193 @@
...
+func (mh *CustomMetricsHandler) PublishMetrics(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	w.Header().Set("Content-Type", "application/json")
+
+	username, password, authOK := r.BasicAuth()
+
+	if authOK == false {
+		mh.logger.Info("error-processing-authorizaion-header")
+		handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+			Code:    "Authorization-Failure-Error",
+			Message: "Error varifying user credentials. Basic authorization is not used properly"})
```
cdlliuy: Typo for "varifying" ?

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,193 @@
...
+		// If Credentials in DB is not valid
+		if !isValid {
+			mh.logger.Error("error-validating-authorizaion-header", err)
+			handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+				Code:    "Authorization-Failure-Error",
+				Message: "Basic authorization credential does not match"})
+			return
+		}
+		// Valid credentials found in DB so update the cache
+		mh.credentialCache.Set(appID, credentials, mh.cacheTTL)
```
cdlliuy: I think we need to update credentialCache with "usernameHash, passwordHash"  regardingless whether the validation result of the auth  info in incoming request. 
"usernameHash, passwordHash" is retrieved from database, so these values should be correct,

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,193 @@
...
+	mh.logger.Debug("failed-to-authorize-credentials")
+	return false
+}
+
+func (mh *CustomMetricsHandler) validateCustomMetricTypes(appGUID string, metricsConsumer *models.MetricsConsumer) (bool, error) {
+	standardMetricsTypes := make(map[string]struct{})
+	standardMetricsTypes[models.MetricNameMemoryUsed] = struct{}{}
+	standardMetricsTypes[models.MetricNameMemoryUtil] = struct{}{}
+	standardMetricsTypes[models.MetricNameThroughput] = struct{}{}
+	standardMetricsTypes[models.MetricNameResponseTime] = struct{}{}
```
cdlliuy: how about claim the  standardMetricsTypes  as constant?

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,193 @@
...
+
+	scalingPolicy, err := mh.policyDB.GetAppPolicy(appGUID)
```
cdlliuy: we need to fetch AppPolicy for each request and parse the allowedMetricTypeSet here. 

How about caching the allowedMetricTypeSet as well, and set a reasonable expire TTL., i.e. 5 minutes. 
Just as the credential cache,   
if cache expired, fetch it from policy db
if not found,  fetch it from policy db
if the incoming metric type is invalid, fetch it from policy db to get the latest policy and revalidate
if the incoming metric type is valid previously, but not valid anymore due to the policy changes  and cache entry doesn't expire,  then just forward the metric to loggregator.   Then, in metric collector,  we can add an implementation to discard the staled metrics .

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Wanted to keep this as a set of StandardMetrics.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Updated after adding cache for metric types

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Done

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Fixed

src/autoscaler/cf/uaa_client.go
paltanmoy: Followed  noaa samples suggestion  https://github.com/cloudfoundry/noaa/blob/master/samples/token_refresher/main.go 

I will have a look if we can remove this through a separate PR

src/autoscaler/cf/config.go
paltanmoy: Removed

src/autoscaler/routes/routes.go
paltanmoy: Done

src/autoscaler/db/sqldb/policy_sqldb.go
paltanmoy: Done

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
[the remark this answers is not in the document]
paltanmoy: Used

src/autoscaler/metricsforwarder/config/config.go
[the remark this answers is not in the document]
paltanmoy: Validation Added

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Done

paltanmoy:
@cdlliuy @boyang9527 @qibobo Do you have any more comments here?  If not, can we merge this PR ?

qibobo:
LGTM

cdlliuy:
@paltanmoy , sorry for the delay. I am working on that now.

src/autoscaler/db/sqldb/policy_sqldb.go
boyang9527: the user may emit the custom metrics without a rule defined. This will give the user the possibility to run performance test, watch how the metrics go with the workload and identify the right upper and lower threshold.

src/autoscaler/cf/uaa_client.go
cdlliuy: It looks like there is no reference to this piece of code now.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,208 @@
...
+	// Credentials not found in cache or
+	// stale cache entry with invalid credential found in cache
+	// search in the database and update the cache
+	if !found || !isValid {
+		usernameHash, passwordHash, err := mh.policyDB.GetCustomMetricsCreds(appID)
+		if err != nil {
+			mh.logger.Error("error-during-getting-binding-credentials-from-policyDB", err)
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error reading custom metrics request body"})
```
cdlliuy: This error msg seems not to be precise enough.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,208 @@
...
+			Message: "Error reading custom metrics request body"})
+		return
+	}
+	var metricsConsumer *models.MetricsConsumer
+	err = json.Unmarshal(body, &metricsConsumer)
+	if err != nil {
+		mh.logger.Error("error-unmarshaling-metrics", err, lager.Data{"body": r.Body})
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error unmarshaling custom metrics request body"})
```
cdlliuy: If the submitted custom metric doesn't align with the expected format of MetricsConsumer,  the json.unmarshal here will fail.  In this case,  should we should Interal-Server-Error ?  I think it should be 400 Bad request.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, metric := range metrics {
+		mh.metricForwarder.EmitMetric(metric)
+	}
+	w.WriteHeader(http.StatusOK)
+}
+
+func (mh *CustomMetricsHandler) validateCredentials(username string, usernameHash string, password string, passwordHash string) bool {
+	isUsernameAuthenticated := bcrypt.CompareHashAndPassword([]byte(usernameHash), []byte(username))
+	isPasswordAuthenticated := bcrypt.CompareHashAndPassword([]byte(passwordHash), []byte(password))
+	if isPasswordAuthenticated == nil && isUsernameAuthenticated == nil { // password matching successfull
```
cdlliuy: The variable name looks like a "bool" type , but it is not .. 
how about using 
usernameAuthErr := ...
passwordAuthErr := ...
if usernameAuthErr == nil && passwordAuthErr == nil 

or let us omit the separated variable here

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,208 @@
...
+		}
+		if err == nil && scalingPolicy == nil {
+			mh.logger.Debug("no-policy-found", lager.Data{"appId": appGUID})
+			return false, errors.New("no policy found")
+		}
+		for _, metrictype := range scalingPolicy.ScalingRules {
+			allowedMetricTypeSet[metrictype.MetricType] = struct{}{}
+		}
+		//update the cache
+		mh.allowedMetricCache.Set(ALLOWED_METRIC_TYPES, allowedMetricTypeSet, mh.cacheTTL)
```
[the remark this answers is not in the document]
cdlliuy: Another 2 comments
1.  as the validation of Credentials, we need to consider the refresh of allowedMetricscache.  i.e. when policy is updated. 
2.  if we always return "false" && "error" together  for validation failure , and "true" && nil for the validation success,  then let us return "err" only.    The line 113 	"if !isValidRequest && err != nil " looks strange.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
cdlliuy: I still think we need to do  len(metrics) <= 0 check first .  moving this part inside ValidateCustomMetricTypes(), and make it happen before checking the allowed metric types will save some kinds of unnecessary computation.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,208 @@
...
+	for _, metric := range metricsConsumer.CustomMetrics {
+		metrics = append(metrics, &models.CustomMetric{
+			AppGUID:       appID,
+			InstanceIndex: metricsConsumer.InstanceIndex,
+			Name:          metric.Name,
+			Value:         metric.Value,
+			Unit:          metric.Unit,
+		})
+	}
+	return metrics
```
cdlliuy: why make a copy of CustomMetrics here?  could we just use  range metricsConsumer.CustomMetrics directly when EmitMetric?

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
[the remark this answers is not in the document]
paltanmoy: Changed the cache key to appID, validation only returns the error now.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Renamed the variables

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Making sense, changed to BadRequest

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Fixed

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
paltanmoy: Wanted to maintain the modularity here.

src/autoscaler/cf/uaa_client.go
paltanmoy: This is being used while creating the firehose consumer to read the metrics.

qibobo:
@paltanmoy  All the test certs have been renewed (https://github.com/cloudfoundry-incubator/app-autoscaler/commit/0251d4395b125d0ac0891f77b696ccab3be7c7f5). I think you need to update the test certs in your PR.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers_test.go
```diff
@@ -0,0 +1,306 @@
...
+				})
+
+				It("should search in the database and returns status code 200", func() {
+					Expect(policyDB.GetCustomMetricsCredsCallCount()).To(Equal(1))
+					Expect(resp.Code).To(Equal(http.StatusOK))
+				})
+			})
+		})
+
+		Context("when a request to publish custom metrics comes with malformed request body", func() {
```
cdlliuy: Can we add more test case for this part?   for example but not limited to:  
1, invalid metrics which conflict with standard metric, 
2 invalid metrics which is not an  allowed one,   
3. allowed metric type set already, but policy is updated , then what will happen ? 
.....
....

cdlliuy:
Furthermore,  we need to consider rate limit on metric forwarder,  either by setting limits  for  incoming request or by putting limits when emit to loggregator.   
Maybe it requires another PR to enhance this part.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers_test.go
paltanmoy: Added/Modified Unit test cases

src/autoscaler/metricsforwarder/cmd/metricsforwarder/main.go
```diff
@@ -0,0 +1,116 @@
...
+
+	var policyDB db.PolicyDB
+	policyDB, err = sqldb.NewPolicySQLDB(conf.Db.PolicyDb, logger.Session("policy-db"))
+	if err != nil {
+		logger.Error("failed-to-connect-policy-database", err, lager.Data{"dbConfig": conf.Db.PolicyDb})
+		os.Exit(1)
+	}
+	defer policyDB.Close()
+
+	credentialCache := cache.New(conf.CacheTTL, -1)
```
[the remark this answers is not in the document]
paltanmoy: added a configurable cache cleanup interval

src/autoscaler/metricsforwarder/server/custom_metrics_handlers_test.go
cdlliuy: will you continue to enhance to address  the last suggestion  "allowed metric type set already, but policy is updated , then what will happen? "

src/autoscaler/metricsforwarder/cmd/metricsforwarder/metricsforwarder_suite_test.go
```diff
@@ -109,6 +109,10 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	cfg.Logging.Level = "debug"
 	cfg.MetronAddress = grpcIngressTestServer.GetAddr()
 	cfg.Server.Port = 10000 + GinkgoParallelNode()
+	cfg.Server.NodeAddrs = []string{"localhost"}
```
cdlliuy: I think sharing is not necessary for metric forwarder

paltanmoy: Yes, sharding has been removed.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers_test.go
paltanmoy: Added policymanager to keep updating the policy modification.

src/autoscaler/metricsforwarder/manager/policyManager.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			continue
+		}
+		policies := pm.computePolicies(policyJsons)
+
+		pm.pLock.Lock()
+		pm.policyMap = policies
+		pm.pLock.Unlock()
+
+		isRefreshed := pm.RefreshAllowedMetricCache(policies)
```
boyang9527: looks like isRefreshed will be false if there is error when replacing cache.  I would suggest to 
1) return error instead of true/false, and log the error
2) when there is error,  still wait for next cycle. In the current impl, it keeps query policy db and replace cache, which will cause performance issue.

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,206 @@
...
+func (mh *CustomMetricsHandler) PublishMetrics(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	w.Header().Set("Content-Type", "application/json")
+
+	username, password, authOK := r.BasicAuth()
+
+	if authOK == false {
+		mh.logger.Info("error-processing-authorizaion-header")
+		handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+			Code:    "Authorization-Failure-Error",
+			Message: "Error verifying user credentials. Basic authorization is not used properly"})
```
boyang9527: I would suggest the message to be "Incorrect credentials"

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,206 @@
...
+		// update the cache
+		mh.credentialCache.Set(appID, credentials, mh.cacheTTL)
+
+		isValid = mh.validateCredentials(username, credentials.Username, password, credentials.Password)
+		// If Credentials in DB is not valid
+		if !isValid {
+			mh.logger.Error("error-validating-authorizaion-header", err)
+			handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+				Code:    "Authorization-Failure-Error",
+				Message: "Basic authorization credential does not match"})
```
boyang9527: again "incorrect credentials"

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,206 @@
...
+		// Credentials found in cache
+		credentials = res.(models.CustomMetricCredentials)
+		isValid = mh.validateCredentials(username, credentials.Username, password, credentials.Password)
+	}
+
+	// Credentials not found in cache or
+	// stale cache entry with invalid credential found in cache
+	// search in the database and update the cache
+	if !found || !isValid {
+		usernameHash, passwordHash, err := mh.policyDB.GetCustomMetricsCreds(appID)
```
boyang9527: if the app keeps using the wrong credential,  it will always query database and update cache. I think we can 
1) increase the interval to active cache update
2) update the cache passively when last update time is long enough

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,184 @@
...
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		mh.logger.Error("error-reading-request-body", err, lager.Data{"body": r.Body})
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error reading custom metrics request body"})
+		return
+	}
+	var metricsConsumer *models.MetricsConsumer
```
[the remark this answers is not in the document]
boyang9527: I would agree with Yang.  This should be 
```var metricsConsumer models.MetricsConsumer```
or 
```metricsConsumer := models.MetricsConsumer{}```

File: src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go [hunk 40]
```diff
@@ -0,0 +1,206 @@
...
+			Code:    "Bad-Request",
+			Message: "Error unmarshaling custom metrics request body"})
+		return
+	}
+	err = mh.validateCustomMetricTypes(appID, metricsConsumer)
+	if err != nil {
+		mh.logger.Error("failed-validating-metrictypes", err, lager.Data{"metrics": metricsConsumer})
+		handlers.WriteJSONResponse(w, http.StatusBadRequest, models.ErrorResponse{
+			Code:    "Bad-Request",
+			Message: "Error validating custom metrics type"})
```
boyang9527: We need to give a clear message "custom metric xxx is not defined in policy"

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,206 @@
...
+	standardMetricsTypes[models.MetricNameResponseTime] = struct{}{}
+	standardMetricsTypes[models.MetricNameCPUUtil] = struct{}{}
+
+	allowedMetricTypeSet := make(map[string]struct{})
+	res, found := mh.allowedMetricCache.Get(appGUID)
+	if found {
+		// AllowedMetrics found in cache
+		allowedMetricTypeSet = res.(map[string]struct{})
+	} else {
+		//  AllowedMetrics not found in cache, find AllowedMetrics from Database
```
boyang9527: again this will always querying database if the app keeps using wrong metric name

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
boyang9527: I would agree, while you can just check whether len(metricsConsumer.CustomMetrics) == 0 or not.

boyang9527:
I am going to close this PR given it has been merged into custom_metrics branch

PR #365, opened 2018-07-25
reduce the log of info level

qibobo:
[#159265038]

### DIFF
--- a/src/autoscaler/db/sqldb/lock_sqldb.go
+++ b/src/autoscaler/db/sqldb/lock_sqldb.go
@@ -190,7 +190,7 @@ func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
 		}
 
 		if newLock {
-			ldb.logger.Debug("acquired-lock-successfully")
+			ldb.logger.Info("acquired-lock-successfully")
 		} else {
 			ldb.logger.Debug("renewed-lock-successfully")
 		}
--- a/src/autoscaler/db/sqldb/lock_sqldb.go
+++ b/src/autoscaler/db/sqldb/lock_sqldb.go
@@ -158,7 +158,7 @@ func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
 				return err
 			}
 			if lastUpdatedTimestamp.Add(time.Second * time.Duration(fetchedLock.Ttl)).Before(currentTimestamp) {
-				ldb.logger.Debug("lock-expired", lager.Data{"Owner": fetchedLock.Owner})
+				ldb.logger.Info("lock-expired", lager.Data{"Owner": fetchedLock.Owner})
 				err = ldb.remove(fetchedLock.Owner, tx)
 				if err != nil {
 					ldb.logger.Error("failed-to-release-existing-lock", err)
--- a/src/autoscaler/db/sqldb/lock_sqldb.go
+++ b/src/autoscaler/db/sqldb/lock_sqldb.go
@@ -142,7 +142,7 @@ func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
 		newLock := false
 		fetchedLock, err := ldb.fetch(tx)
 		if err == nil && fetchedLock == nil {
-			ldb.logger.Info("no-one-holds-the-lock")
+			ldb.logger.Debug("no-one-holds-the-lock")
 			newLock = true
 		} else if err != nil {
 			ldb.logger.Error("failed-to-fetch-lock", err)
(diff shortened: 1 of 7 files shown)


src/autoscaler/db/sqldb/lock_sqldb.go
```diff
@@ -190,9 +190,9 @@ func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
 		}
 
 		if newLock {
-			ldb.logger.Info("acquired-lock-successfully")
+			ldb.logger.Debug("acquired-lock-successfully")
```
cdlliuy: Since it is "newLock", how about keeping this as "Info"?

src/autoscaler/scalingengine/schedule/sync.go
```diff
@@ -70,7 +70,7 @@ func (ss *ActiveScheduleSychronizer) synchronizeActiveSchedules() {
 	for appId, schedule := range asScheduler {
 		scheduleId := asEngine[appId]
 		if scheduleId == "" || scheduleId != schedule.ScheduleId {
-			ss.logger.Info("synchronize-active-schedules-find-missing-active-schedule-start", lager.Data{"appId": appId, "schedule": schedule})
+			ss.logger.Debug("synchronize-active-schedules-find-missing-active-schedule-start", lager.Data{"appId": appId, "schedule": schedule})
```
boyang9527: this probably needs to be Info too

src/autoscaler/scalingengine/schedule/sync.go
```diff
@@ -81,7 +81,7 @@ func (ss *ActiveScheduleSychronizer) synchronizeActiveSchedules() {
 
 	for appId, scheduleId := range asEngine {
 		if asScheduler[appId] == nil {
-			ss.logger.Info("synchronize-active-schedules-find-missing-active-schedule-end", lager.Data{"appId": appId, "scheduleId": scheduleId})
+			ss.logger.Debug("synchronize-active-schedules-find-missing-active-schedule-end", lager.Data{"appId": appId, "scheduleId": scheduleId})
```
boyang9527: same as above

src/autoscaler/db/sqldb/lock_sqldb.go
```diff
@@ -158,7 +158,7 @@ func (ldb *LockSQLDB) Lock(lock *models.Lock) (bool, error) {
 				return err
 			}
 			if lastUpdatedTimestamp.Add(time.Second * time.Duration(fetchedLock.Ttl)).Before(currentTimestamp) {
-				ldb.logger.Info("lock-expired", lager.Data{"Owner": fetchedLock.Owner})
+				ldb.logger.Debug("lock-expired", lager.Data{"Owner": fetchedLock.Owner})
```
paltanmoy: This is an important and non-regular condition to check if the lock is expired without renew. So lets keep this as info only.

src/autoscaler/sync/dblock.go
```diff
@@ -63,13 +63,13 @@ func (dblock *DatabaseLock) InitDBLockRunner(retryInterval time.Duration, ttl ti
...
 					} else {
-						dblock.logger.Info("successfully-released-lock", lager.Data{"owner": owner})
+						dblock.logger.Debug("successfully-released-lock", lager.Data{"owner": owner})
 					}
 					os.Exit(1)
 				}
 				if isLockAcquired && readyToAcquireLock {
 					readyToAcquireLock = false
-					dblock.logger.Info("successfully-acquired-lock", lager.Data{"owner": owner})
+					dblock.logger.Debug("successfully-acquired-lock", lager.Data{"owner": owner})
```
paltanmoy: @qibobo acquiring lock is not that frequent in nature and its important from debugging perspective. Can we keep it as Info only ?

src/autoscaler/sync/dblock.go
```diff
@@ -63,13 +63,13 @@ func (dblock *DatabaseLock) InitDBLockRunner(retryInterval time.Duration, ttl ti
 					if releaseErr != nil {
 						dblock.logger.Error("failed-to-release-lock ", releaseErr)
 					} else {
-						dblock.logger.Info("successfully-released-lock", lager.Data{"owner": owner})
+						dblock.logger.Debug("successfully-released-lock", lager.Data{"owner": owner})
```
paltanmoy: @qibobo releasing lock is not that frequent in nature and its important from debugging perspective. Can we keep it as Info only ?

qibobo:
@boyang9527 @paltanmoy updated.

PR #366, opened 2018-07-26
scheduler synchronize rest api issue: when policy is updated without …

qibobo:
…schedule, the old schedules do not deleted.

[#159327430]


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -818,6 +818,10 @@ public SynchronizeResult synchronizeSchedules(){
 					updateCount++;
 					continue;
 				}
+			}else{
+				// there is no schedule in the new policy, so the old schedules of this app should be deleted.
+				toDeletedAppIds.add(appIdInPolicy);
+				deleteCount++;
```
boyang9527: this is still update, not deleteCount

PR #371, opened 2018-07-31
adapt public rest api to cf style

qibobo:
[#159391488]


api/lib/routes/routeHelper.js
```diff
@@ -13,19 +13,32 @@ exports.isString = function(value) {
...
+exports.pagination = function(resultList, page, resultsPerPage, req) {
   var result = {};
   var totalResults = resultList.length;
   var totalPages = 0;
   var perPage = Number.parseInt(resultsPerPage);
   var pageNo = Number.parseInt(page);
-  totalPages = (totalResults % perPage) === 0 ? (totalResults / perPage) : (totalResults / perPage + 1);
+  var prevUrl = null;
+  var nextUrl = null;
+  totalPages = (totalResults % perPage) === 0 ? (totalResults / perPage) : (Number.parseInt(totalResults / perPage) + 1);
```
boyang9527: why Number.parseInt ?

qibobo: var totalResults = 5;
var resultsPerPage = 2;
var totalPages = 0;
var perPage = Number.parseInt(resultsPerPage);
totalPages = (totalResults % perPage) === 0 ? (totalResults / perPage) : (totalResults / perPage + 1);
//totalPages = (totalResults % perPage) === 0 ? (totalResults / perPage) : (Number.parseInt(totalResults / perPage) + 1);
console.log(totalPages)
Without Number.parseInt the result of the code above is 3.5 and after adding it the result is 3.

boyang9527: thanks for the explanation.

boyang9527:
LGTM

docs/Public_API.rst
```diff
@@ -30,7 +30,7 @@ Parameters
 +--------------------+-------------------------------------------------------------------------------+---------------------------------------------------------------------+-----------------------+----------------------------------+
 | end-time           | The end time                                                                  | int, the number of nanoseconds elapsed since January 1, 1970 UTC.   | false. default 'now'  | end-time=1494989549117047288     |
 +--------------------+-------------------------------------------------------------------------------+---------------------------------------------------------------------+-----------------------+----------------------------------+
-| order              | The order type. The scaling history will be order by timestamp asc or desc.   | string,”asc” or “desc”                                              | false. default desc   | order=desc                       |
+| order-direction    | The order type. The scaling history will be order by timestamp asc or desc.   | string,”asc” or “desc”                                              | false. default desc   | order=desc                       |
```
cdlliuy: missed "pre_url" and "next_url" in this table

cdlliuy: use " instead of ” ?

cdlliuy: order-direction=desc

api/lib/routes/aggregated_metrics.js
```diff
@@ -13,13 +13,13 @@ module.exports = function(settings) {
     var metricType = req.params.metric_type;
     var startTime = req.query["start-time"];
     var endTime = req.query["end-time"];
-    var order = req.query["order"];
+    var orderDirection = req.query["order-direction"];
```
cdlliuy: considering the backward compatibility  , could we support both "order" and "order-direction" as a valid parameter?  In the external document, we can claim the "order-direction" only.

cdlliuy:
LGTM

PR #381, opened 2018-09-04
prune non-existent application details

paltanmoy:
[#159282737]


src/autoscaler/operator/applicationsyncer.go
```diff
@@ -0,0 +1,46 @@
...
+	as.logger.Debug("deleting non-existent application details")
+	// Get all the application details from policyDB
+	appIds, err := as.policyDb.GetAppIds()
+	if err != nil {
+		as.logger.Error("failed-to-get-apps", err)
+		return
+	}
+	// For each app check if they really exist or not via CC api call
+	for appID := range appIds {
+		_, err = as.cfClient.GetApp(appID)
```
cdlliuy: There are multiple possible errors in GetApp() calls
For example,  if the status code = 200 OK,  an error returned: 
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/9f93c7c2b4ae3091b7b409b855755555c74a6fd6/src/autoscaler/cf/app.go#L41-L45
With the code above,  what will happen when the CC api endpoint is temporarily unavailable  ? All GetAPP() may fail with error, then all policy might be deleted.    We met such a situation in daily operations ..

Technically, we can only delete App with the following response:
RESPONSE: [2018-09-05T10:55:18+08:00]
HTTP/1.1 404 Not Found
{
  "description": "The app could not be found: xxxxx",
  "error_code": "CF-AppNotFound",
  "code": 100004
}
Otherwise, we may false delete a valid policy.

paltanmoy: @cdlliuy  PR updated after incorporating the review comments

src/autoscaler/cf/app.go
```diff
@@ -39,8 +41,28 @@ func (c *cfClient) GetApp(appId string) (*models.AppEntity, error) {
...
-		c.logger.Error("get-app-instances-response", err)
+		respBody, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			c.logger.Error("failed-to-read-response-body-while-getting-app-summary", err, lager.Data{"appid": appId})
+			return nil, err
+		}
+		var bodydata map[string]interface{}
+		err = json.Unmarshal([]byte(respBody), &bodydata)
+		if err != nil {
+			c.logger.Error("failed-to-unmarshal-response-body-while-getting-app-summary", err, lager.Data{"appid": appId})
```
cdlliuy: 1. Besides resp.StatusCode != http.StatusOK , could we check whether resp.StatusCode == 404 first before parsing the response body ?

2.  When CC API is available,  I think the response  error should be written in JSON format.  But if CC API endpoint is not accessible here ,  the response body maybe empty or plain text . In this case, most likely the unmarshal error message talks about the json parsing error , instead of the original http response.  How about adding respBody into the c.logger.Error ?

src/autoscaler/cf/app_test.go
```diff
@@ -83,11 +90,35 @@ var _ = Describe("App", func() {
...
+		Context("when get app summary return non-200 and non-404 status code", func() {
 			BeforeEach(func() {
 				fakeCC.AppendHandlers(
 					ghttp.CombineHandlers(
-						ghttp.RespondWithJSONEncoded(http.StatusNotFound, ""),
+						ghttp.RespondWithJSONEncoded(http.StatusInternalServerError, CFSummaryResponse{
+							Description: "Server error",
+							ErrorCode:   "ServerError",
+							Code:        100001,
+						}),
```
cdlliuy: as above, we can't determine there will be a JSON response returned in each failure case.

src/autoscaler/cf/app_test.go
```diff
@@ -17,6 +18,12 @@ import (
 	"net/url"
 )
 
+type CFSummaryResponse struct {
```
cdlliuy: it looks like the error response struct, then the name CFSummaryResponse is confusing

paltanmoy:
travis build is failing for https://github.com/go-yaml/yaml/issues/401 which is caused by https://github.com/niemeyer/gopkg/issues/63

We have to wait till this got resolved.

src/autoscaler/cf/app_test.go
```diff
@@ -17,6 +18,12 @@ import (
 	"net/url"
 )
 
+type ErrorResponse struct {
```
boyang9527: you may want to move this to https://github.com/cloudfoundry-incubator/app-autoscaler/blob/develop/src/autoscaler/models/errors.go

src/autoscaler/cf/app.go
```diff
@@ -39,6 +41,32 @@ func (c *cfClient) GetApp(appId string) (*models.AppEntity, error) {
 	defer resp.Body.Close()
 
 	if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode == 404 {
+			respBody, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
```
boyang9527: this piece of code can be simplified to unmarshal to CF error response struct

src/autoscaler/helpers/errors.go
```diff
@@ -0,0 +1,14 @@
+package helpers
+
```
boyang9527: move to models package?

src/autoscaler/operator/config/config.go
```diff
@@ -109,6 +120,9 @@ var defaultConfig = Config{
 		LockRetryInterval: DefaultRetryInterval,
 		LockTTL:           DefaultLockTTL,
 	},
+	AppSyncer: AppSyncerConfig{
+		SyncInterval: DefaultSyncInterval,
```
boyang9527: the DefaultSyncInterval (10 minutes) is too short. Imagine we may have 10K apps,  it will take a pretty long time.  And I think remove the orphan app policies is not that urgent so we probably can give it one day as default sync interval

PR #387, opened 2018-09-20
Add http connection timeout as a configuration to all autoscaler components

qibobo:
[#160241806]


boyang9527:
the setting should be http_client_timeout instead of http_request_timeout

qibobo:
@boyang9527  updated.

PR #389, opened 2018-10-18
override tcp dial timeout of cfhttp httpclient

qibobo:


qibobo:
@boyang9527 Do we need to add another configuration for dial timeout? Seems there are too many configurations.

PR #392, opened 2018-10-24
Added validation to restrict threshold value to be integer only

paltanmoy:


api/lib/validation/attributeValidator.js
```diff
@@ -33,6 +33,10 @@ var normalizeEndDate = function(object) {
   return object.end_date;
 }
 
+var isInt = function(n) {
+  return n % 1 === 0;
+}
+
```
cdlliuy: According to the sample in https://www.npmjs.com/package/jsonschema ,  the change in schemaValidator.js is enough.   Why we need to check attributeValidator again?

api/lib/validation/schemaValidator.js
```diff
@@ -94,7 +94,7 @@ var getScalingRuleSchema = function() {
     'properties' : {
       'metric_type':{ 'type':'string' },
       'breach_duration_secs':{ 'type':'number','minimum': 60,'maximum': 3600 },
-      'threshold':{ 'type':'number'},
+      'threshold':{ 'type':'integer'},
```
cdlliuy: how about making breach_duration_secs and cool_down_secs to "integer" as well?

paltanmoy:
@cdlliuy build is green after re-try.

qibobo:
LGTM

cdlliuy:
LGTM

PR #393, opened 2018-10-25
[auth] allow admin to access autoscaler

zyjiaobj:


api/lib/oauth/oauth.js
```diff
@@ -20,6 +21,11 @@ module.exports = function(settings) {
       callback(null, false);
       return;
     }
+    var decodeJson = jwt.decode(userToken.split(' ').length > 0 && userToken.split(' ')[1]);
+    if (decodeJson && decodeJson.scope && decodeJson.scope.indexOf("cloud_controller.admin") >= 0) {
+      callback(null, true);
```
cdlliuy: please revise the solution.  The token must be validated from UAA , instead of decoding locally.   Otherwise, we will have security risk here.

qibobo:
LGTM

PR #394, opened 2018-11-01
Add health endpoint for go component

qibobo:


src/autoscaler/metricscollector/config/config.go
```diff
@@ -68,10 +70,14 @@ var defaultCollectorConfig = CollectorConfig{
 	MetricCacheSizePerApp: DefaultMetricCacheSizePerApp,
 }
 
+type HealthConfig struct {
```
boyang9527: move out as a shared type definition

src/autoscaler/operator/config/config.go
```diff
@@ -78,9 +78,17 @@ type AppSyncerConfig struct {
 	DB           db.DatabaseConfig `yaml:"db"`
 	SyncInterval time.Duration     `yaml:"sync_interval"`
 }
+type HealthConfig struct {
```
boyang9527: same as above

qibobo:
@boyang9527 updated

PR #401, opened 2018-11-13
servicebroker rewritten in golang

rohitsharma04:
This PR only contains broker related code 

I'll do different PRs for the followings:
- Public API
- Policy JSON validator
- Credential Management


src/autoscaler/api/cmd/api/main.go
```diff
@@ -0,0 +1,82 @@
...
+	}
+	configFile.Close()
+
+	err = conf.Validate()
+	if err != nil {
+		fmt.Fprintf(os.Stdout, "failed to validate configuration : %s\n", err.Error())
+		os.Exit(1)
+	}
+
+	cfhttp.Initialize(5 * time.Second)
```
qibobo: cfhttp.Initialize(conf.HttpClientTimeout)

src/autoscaler/api/server/api_handler.go
```diff
@@ -0,0 +1,185 @@
...
+			Code:    "Bad Request",
+			Message: "Malformed or missing mandatory data",
+		})
+		return
+	}
+
+	err = h.bindingdb.CreateServiceInstance(instanceId, body.OrgGUID, body.SpaceGUID)
+	if err != nil {
+		if err == db.ErrAlreadyExists {
+			w.Write(nil)
```
qibobo: lack of http status code?

src/autoscaler/api/server/api_handler.go
```diff
@@ -0,0 +1,185 @@
...
+			Code:    "Bad Request",
+			Message: "Malformed or missing mandatory data",
+		})
+		return
+	}
+
+	err = h.bindingdb.CreateServiceBinding(bindingId, instanceId, body.AppID)
+	if err != nil {
+		if err == db.ErrAlreadyExists {
+			w.Write(nil)
```
qibobo: http status code?

src/autoscaler/api/server/api_handler_test.go
```diff
@@ -0,0 +1,370 @@
...
+					PlanID:    "a-plan-id",
+				}
+				body, err := json.Marshal(bindingRequestBody)
+				Expect(err).NotTo(HaveOccurred())
+
+				req, err = http.NewRequest(http.MethodPut, "", bytes.NewReader(body))
+
+				bindingdb.CreateServiceBindingReturns(db.ErrAlreadyExists)
+			})
+			It("succeeds with 200", func() {
```
qibobo: The status code should be 409.

src/autoscaler/api/server/api_handler_test.go
```diff
@@ -0,0 +1,370 @@
...
+					},
+				}
+				body, err := json.Marshal(instanceCreationReqBody)
+				Expect(err).NotTo(HaveOccurred())
+
+				req, err = http.NewRequest(http.MethodPut, "", bytes.NewReader(body))
+
+				bindingdb.CreateServiceInstanceReturns(db.ErrAlreadyExists)
+			})
+			It("succeeds with 200", func() {
```
qibobo: The status code should be 409.

src/autoscaler/api/server/api_handler_test.go
```diff
@@ -0,0 +1,370 @@
...
+				req, err = http.NewRequest(http.MethodPut, "", bytes.NewReader(body))
+
+				bindingdb.DeleteServiceBindingReturns(db.ErrDoesNotExist)
+			})
+			It("fails with 410", func() {
+				Expect(resp.Code).To(Equal(http.StatusGone))
+			})
+		})
+
+		Context("When database DeleteServiceBinding call returns error other than ErrAlreadyExists", func() {
```
qibobo: ErrAlreadyExists => ErrDoesnotExist?

src/autoscaler/db/sqldb/binding_sqldb.go
```diff
@@ -0,0 +1,145 @@
...
+		return err
+	}
+
+	if rows.Next() {
+		rows.Close()
+		return db.ErrAlreadyExists
+	}
+	rows.Close()
+
+	query = "INSERT INTO service_instance" +
```
qibobo: How about we directly insert into the table without a select? The serviceInstanceId is the primary key, so a duplicated insertion will cause a duplicated error. By this, only one SQL statement is executed.
And the same for deletion, just delete without select. We can check how many rows affected by the delete statement. If the result is 0, so we can know that the record does not exist.

src/autoscaler/api/server/api_handler_test.go
rohitsharma04: > 200 SHOULD be returned if the Service Instance already exists, is fully provisioned, and the requested parameters are identical to the existing Service Instance.
https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#provisioning

src/autoscaler/api/server/api_handler_test.go
rohitsharma04: > 200 SHOULD be returned if the binding already exists and the requested parameters are identical to the existing binding. The expected response body is below.
https://github.com/openservicebrokerapi/servicebroker/blob/v2.14/spec.md#binding

src/autoscaler/db/sqldb/binding_sqldb.go
rohitsharma04: For deletion this can work.
For Insertion, primary key duplication error is just text and comparing text is not a good idea as it might change with different versions.

src/autoscaler/api/cmd/api/api_test.go
```diff
@@ -0,0 +1,99 @@
...
+		})
+
+		It("should stop", func() {
+			runner.Session.Interrupt()
+			Eventually(runner.Session, 5).Should(Exit(0))
+		})
+
+	})
+
+})
```
boyang9527: need to add basic http server test

src/autoscaler/api/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	if c.DB.BindingDB.URL == "" {
+		return fmt.Errorf("Configuration error: BindingDB URL is empty")
+	}
+	if c.BrokerUsername == "" {
+		return fmt.Errorf("Configuration error: BrokerUsername is empty")
+	}
+	if c.BrokerPassword == "" {
+		return fmt.Errorf("Configuration error: BrokerPassword is empty")
+	}
+	if c.Catalog == "" {
```
boyang9527: need to verify catalog is a valid json with required fields

src/autoscaler/api/config/config.go
```diff
@@ -0,0 +1,80 @@
...
+	BindingDB db.DatabaseConfig `yaml:"binding_db"`
+}
+
+type Config struct {
+	Logging              helpers.LoggingConfig `yaml:"logging"`
+	Server               ServerConfig          `yaml:"server"`
+	DB                   DBConfig              `yaml:"db"`
+	BrokerUsername       string                `yaml:"broker_username"`
+	BrokerPassword       string                `yaml:"broker_password"`
+	Catalog              string                `yaml:"catalog"`
```
boyang9527: perhaps we can put the catalog to a file

src/autoscaler/api/server/api_handler.go
```diff
@@ -0,0 +1,185 @@
...
+			return
+		}
+
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error deleting service instance"})
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
boyang9527: should return 204

src/autoscaler/api/server/api_handler.go
```diff
@@ -0,0 +1,185 @@
...
+	w.Write([]byte("{}"))
+}
+
+func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request, vars map[string]string) {
```
boyang9527: we need to enable binding with policies

src/autoscaler/api/main.go
```diff
@@ -0,0 +1,26 @@
+package main
```
boyang9527: this file should be removed

boyang9527:
@rohitsharma04  please remove src/autoscaler/api/main.go

pradyutsarma:
LGTM

PR #404, opened 2018-11-20
Updated metric_type schema to support custom metrics

rohitsharma04:
- metric_type can be alphanumeric characters and it can also have
underscores

### DIFF
--- a/servicebroker/config/catalog.schema.json
+++ b/servicebroker/config/catalog.schema.json
@@ -201,13 +201,8 @@
                   "description": "The interval between two successive scaling activity"
                 },
                 "metric_type": {
-                  "enum": [
-                    "memoryused",
-                    "memoryutil",
-                    "responsetime",
-                    "throughput"
-                  ],
-                  "type": "string"
+                  "type": "string",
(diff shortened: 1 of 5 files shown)


servicebroker/config/catalog.schema.json
```diff
@@ -201,13 +201,8 @@
...
                 "metric_type": {
-                  "enum": [
-                    "memoryused",
-                    "memoryutil",
-                    "responsetime",
-                    "throughput"
-                  ],
-                  "type": "string"
+                  "type": "string",
+                  "pattern": "^[a-zA-Z0-9_]+$"
```
qibobo: the metric type can start with a number? such as "999metric".

rohitsharma04: @pradyutsarma @paltanmoy @boyang9527 @cdlliuy 
should custom metric name be allowed to start with number ?

boyang9527: I think that is fine

boyang9527:
LGTM

PR #407, opened 2018-11-30
send default breach duration to scaling engine

boyang9527:
to fix issue https://github.com/cloudfoundry-incubator/app-autoscaler/issues/406


cdlliuy:
LGTM

PR #408, opened 2018-12-01
add timestamp index for data prune

boyang9527:


cdlliuy:
LGTM. it is necessary to create a timestamp index on instancemetric db.  We found the delete took a long time in purge.

boyang9527:
closing

PR #413, opened 2018-12-20
enable setting not to store instance metrics to database

boyang9527:


cdlliuy:
@boyang9527 , conflicts now ..

PR #424, opened 2019-01-03
Custom metrics forwarder

rohitsharma04:


src/autoscaler/metricsforwarder/manager/policyManager.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (pm *PolicyManager) RefreshAllowedMetricCache(policies map[string]*models.AppPolicy) error {
+	pm.mLock.Lock()
+	defer pm.mLock.Unlock()
+	allowedMetricTypeSet := make(map[string]struct{})
+	allowedMetricMap := pm.allowedMetricCache.Items()
+	//Iterating over the cache and replace the allowed metrics for existing policy
+	for applicationId := range allowedMetricMap {
+		scalingPolicy := pm.policyMap[applicationId].ScalingPolicy
```
cdlliuy: we need to validate whether the application's policy exists in the map before using scalingPolicy

src/autoscaler/metricsforwarder/manager/policyManager.go
```diff
@@ -0,0 +1,124 @@
...
+		for _, metrictype := range scalingPolicy.ScalingRules {
+			allowedMetricTypeSet[metrictype.MetricType] = struct{}{}
+		}
+		err := pm.allowedMetricCache.Replace(applicationId, allowedMetricTypeSet, pm.cacheTTL)
+		if err != nil {
+			pm.logger.Error("Error updating allowedMetricCache", err)
+			return err
+		}
```
cdlliuy: So, we  only do "cache update" here , right?  Maybe we can consider to add new allowedMetric to cache and delete staled ones from cache  as well.

src/autoscaler/metricsforwarder/config/config.go
```diff
@@ -0,0 +1,98 @@
...
+
+type LoggingConfig struct {
+	Level string `yaml:"level"`
+}
+
+type LoggregatorConfig struct {
+	CACertFile     string `yaml:"ca_cert"`
+	ClientCertFile string `yaml:"client_cert"`
+	ClientKeyFile  string `yaml:"client_key"`
+}
```
cdlliuy: Why put MetronAddress as a separated entry?    I propose the following configuration yaml format which is more consistent with the other components. 
loggregator:
      metron_address:  xxxx
      tls:
         ca_cert:
        client_cert:
        client_key:

src/autoscaler/metricsforwarder/server/custom_metrics_handlers.go
```diff
@@ -0,0 +1,206 @@
...
+
+	metrics := mh.getMetrics(appID, metricsConsumer)
+
+	if len(metrics) <= 0 {
+		mh.logger.Debug("failed-parsing-custom-metrics-request-body", lager.Data{"metrics": metrics})
+		handlers.WriteJSONResponse(w, http.StatusBadRequest, models.ErrorResponse{
+			Code:    "Bad-Request",
+			Message: "Error parsing request body"})
+		return
+	}
```
cdlliuy: I suggest to move the check of `len(metrics) <=0 ` prior to invoking `validateCustomMetricTypes` to avoid unnecessary validation.

PR #431, opened 2019-01-07
add config not to store instance metrics to db

boyang9527:


qibobo:
The new configuration name "IsMetricsPersistencySupported" does not accurately describe how it works. For "xxxSupported", I think it means whether the component has the the ability to do something.
Now if IsMetricsPersistencySupported is false, so the instance_metric_db is unnecessary, but now it is mandatory without any one to use it.
Maybe it is better to change the name to "IsPersistMetrics" or we can check whether the instance_metric_db configuration is empty or not to determine whether to persist metrics.

boyang9527:
I would prefer not to detect whether there is instance metric db.  I will change the name of the configuration.

PR #445, opened 2019-01-21
remove consul lock relevant code

qibobo:
[#163317022]


src/autoscaler/operator/config/config.go
```diff
@@ -93,7 +87,6 @@ type Config struct {
 	ScalingEngine     ScalingEngineConfig           `yaml:"scaling_engine"`
 	Scheduler         SchedulerConfig               `yaml:"scheduler"`
 	AppSyncer         AppSyncerConfig               `yaml:"app_syncer"`
-	Lock              LockConfig                    `yaml:"lock"`
 	DBLock            DBLockConfig                  `yaml:"db_lock"`
 	EnableDBLock      bool                          `yaml:"enable_db_lock"`
```
boyang9527: if we only support dbLock, then we need to remove this config too.  There will be corresponding changes in main.go.

qibobo: @boyang9527  You mean while consul lock is removed, so dbLock is mandatory for operator?

boyang9527: yes

qibobo:
@boyang9527 updated.

boyang9527:
LGTM

PR #449, opened 2019-01-22
Policy Validation for new Servicebroker

rohitsharma04:


src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			errDetails := gojsonschema.ErrorDetails{
+				"instance_min_count": recSched.ScheduledInstanceMin,
+				"instance_max_count": recSched.ScheduledInstanceMin,
+			}
+			formatString := "instance_min_count {{.instance_min_count}} is higher or equal to instance_max_count {{.instance_max_count}}"
+			err := newPolicyValidationError(instanceMinContext, formatString, errDetails)
+			result.AddError(err, errDetails)
+		}
+
+		if recSched.ScheduledInstanceInit > recSched.ScheduledInstanceMax {
```
boyang9527: how about initial_min less than intance_min?

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			formatString := "initial_min_instance_count {{.initial_min_instance_count}} is greater than instance_max_count {{.instance_max_count}}"
+			err := newPolicyValidationError(initialInstanceMinContext, formatString, errDetails)
+			result.AddError(err, errDetails)
+		}
+
+		//start_time should be before end_time
+		if compareTimesGTEQ(recSched.StartTime, recSched.EndTime) {
+			currentRecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), recurringScheduleContext)
+			errDetails := gojsonschema.ErrorDetails{
+				"i": i,
```
boyang9527: use a better name ? instead of using "i"

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+		}
+		if recSched.StartDate != "" && recSched.EndDate != "" {
+			startDate, _ = time.ParseInLocation(DateLayout, recSched.StartDate, location)
+			endDate, _ = time.ParseInLocation(DateLayout, recSched.EndDate, location)
+			if endDate.Sub(startDate) < 0 {
+
+				currentRecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), recurringScheduleContext)
+				errDetails := gojsonschema.ErrorDetails{
+					"i": i,
```
boyang9527: same as above

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			err := newPolicyValidationError(initialInstanceMinContext, formatString, errDetails)
+			result.AddError(err, errDetails)
+		}
+
+		// start_date_time should be after current_date_time and before end_date_time
+		dateTime := newDateTimeRange(specSched.StartDateTime, specSched.EndDateTime, policy.Schedules.Timezone)
+		if dateTime.startDateTime.Sub(time.Now()) <= 0 {
+			currentSpecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), specficDateScheduleContext)
+			errDetails := gojsonschema.ErrorDetails{
+				"i": i,
```
boyang9527: same as above

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+			formatString := "start_date_time is before current_date_time"
+			err := newPolicyValidationError(currentSpecSchedContext, formatString, errDetails)
+			result.AddError(err, errDetails)
+		}
+		if dateTime.endDateTime.Sub(dateTime.startDateTime) <= 0 {
+			currentSpecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), specficDateScheduleContext)
+			errDetails := gojsonschema.ErrorDetails{
+				"i": i,
```
boyang9527: same as above

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			instanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.instance_min_count", i), specficDateScheduleContext)
+			errDetails := gojsonschema.ErrorDetails{
+				"instance_min_count": specSched.ScheduledInstanceMin,
+				"instance_max_count": specSched.ScheduledInstanceMax,
+			}
+			formatString := "instance_min_count {{.instance_min_count}} is higher or equal to instance_max_count {{.instance_max_count}}"
+			err := newPolicyValidationError(instanceMinContext, formatString, errDetails)
+			result.AddError(err, errDetails)
+		}
+		if specSched.ScheduledInstanceInit > specSched.ScheduledInstanceMax {
```
boyang9527: same as above

src/autoscaler/api/server/api_handler.go
```diff
@@ -139,6 +149,43 @@ func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request,
...
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error creating/updating schedules"})
+			return
+		}
+
+		h.logger.Info("saving policy json", lager.Data{"policy": body.Policy})
+		err = h.policydb.SaveAppPolicy(body.AppID, body.Policy, policyGuid.String())
+		if err != nil {
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
```
boyang9527: try to delete the previous schedule if saving policy to db fails

src/autoscaler/api/server/api_handler.go
```diff
@@ -139,6 +149,43 @@ func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request,
...
+		policyGuid, err := uuid.NewV4()
+		if err != nil {
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error generating policy guid"})
+			return
+		}
+
+		h.logger.Info("creating/updating schedules", lager.Data{"policy": body.Policy})
+		err = h.schedulerUtil.CreateOrUpdateSchedule(body.AppID, body.Policy, policyGuid.String())
```
cdlliuy: There is no validation for the existence of body.AppID in previous code

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+		formatString := "instance_min_count {{.instance_min_count}} is higher or equal to instance_max_count {{.instance_max_count}}"
+		err := newPolicyValidationError(instanceMinContext, formatString, errDetails)
+		result.AddError(err, errDetails)
+	}
+
+	// check InstanceMinCount, IntanceMaxCount and InitialInstanceMinCount for schedules
+	if policy.Schedules == nil {
+		return
+	}
+	schedulesContext := gojsonschema.NewJsonContext("schedules", rootContext)
```
cdlliuy: Missing the implementation for threshold validation as
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/master/api/lib/validation/attributeValidator.js#L238

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+			result.AddError(err, errDetails)
+		}
+
+		//start_time should be before end_time
+		if compareTimesGTEQ(recSched.StartTime, recSched.EndTime) {
+			currentRecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), recurringScheduleContext)
+			errDetails := gojsonschema.ErrorDetails{
+				"i": i,
+			}
+			formatString := "start_time is after end_time"
```
cdlliuy: considering the function compareTimes**GTEQ**,  the msg here should be "same or after"

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+
+		if recSched.StartDate != "" {
+			startDate, _ = time.ParseInLocation(DateLayout, recSched.StartDate, location)
+
+			if startDate.Sub(time.Now()) < 0 {
+				currentRecSchedContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d", i), recurringScheduleContext)
+				errDetails := gojsonschema.ErrorDetails{
+					"i": i,
+				}
+				formatString := "start_date is before current_date"
```
cdlliuy: There is no "hh:mm:ss" setting on start date, so time.ParseInLocation will use 00:00:00 when converts startDate.  With the implementation above, I can't define a schedule starting from today .. 
for example, assuming I want to create a recurring schedule for 10:00 ~ 18:00 with start date  `2019-03-01` at **2019-03-01 8:00AM  **,   the first trigger should be 2019-03-01 10:00AM which is later than the CURRENT time.  This is a correct schedule definition.  But in above implementation,  the start date is 2019-03-01 00:00AM which is before CURRENT ,  the policy is rejected.

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+				err := newPolicyValidationError(currentRecSchedContext, formatString, errDetails)
+				result.AddError(err, errDetails)
+			}
+		}
+		if recSched.StartDate != "" {
```
cdlliuy: use `if recSched.endDate != "" ` ?

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -0,0 +1,369 @@
...
+
+func compareDatesGTEQ(firstDate string, secondDate string) bool {
+	if firstDate == "" {
+		firstDate = "0001-01-01"
+	}
+	if secondDate == "" {
+		secondDate = "0001-01-01"
+	}
+	fd, _ := time.Parse(DateLayout, firstDate)
+	sd, _ := time.Parse(DateLayout, secondDate)
```
cdlliuy: when end-date is not define, we can't use "0001-01-01" as its default value.   The overlapping check will fail

src/autoscaler/api/policyvalidator/policy_validator.go
rohitsharma04: Added

src/autoscaler/api/policyvalidator/policy_validator.go
rohitsharma04: fixed.

src/autoscaler/api/policyvalidator/policy_validator.go
rohitsharma04: fixed.

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -113,67 +126,113 @@ func (pv *PolicyValidator) validateAttributes(policy *models.ScalingPolicy, resu
...
 	recurringScheduleContext := gojsonschema.NewJsonContext("recurring_schedule", schedulesContext)
-	for i, recSched := range policy.Schedules.RecurringSchedules {
+	for scheduleIndex, recSched := range policy.Schedules.RecurringSchedules {
 		if recSched.ScheduledInstanceMin >= recSched.ScheduledInstanceMax {
-			instanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.instance_min_count", i), recurringScheduleContext)
+			instanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.instance_min_count", scheduleIndex), recurringScheduleContext)
 			errDetails := gojsonschema.ErrorDetails{
+				"scheduleIndex":      scheduleIndex,
 				"instance_min_count": recSched.ScheduledInstanceMin,
 				"instance_max_count": recSched.ScheduledInstanceMin,
```
cdlliuy: "instance_max_count": recSched.ScheduledInstanceMax,

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -113,67 +126,113 @@ func (pv *PolicyValidator) validateAttributes(policy *models.ScalingPolicy, resu
...
+		if (recSched.ScheduledInstanceInit != 0) && (recSched.ScheduledInstanceInit < recSched.ScheduledInstanceMin) {
+			initialInstanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.initial_min_instance_count", scheduleIndex), recurringScheduleContext)
 			errDetails := gojsonschema.ErrorDetails{
+				"scheduleIndex":              scheduleIndex,
 				"initial_min_instance_count": recSched.ScheduledInstanceInit,
-				"instance_max_count":         recSched.ScheduledInstanceMin,
+				"instance_min_count":         recSched.ScheduledInstanceMin,
 			}
-			formatString := "initial_min_instance_count {{.initial_min_instance_count}} is greater than instance_max_count {{.instance_max_count}}"
+			formatString := "recurring_schedule[{{.scheduleIndex}}].initial_min_instance_count {{.initial_min_instance_count}} is smaller than recurring_schedule[{{.scheduleIndex}}].instance_min_count {{.instance_min_count}}"
```
cdlliuy: it is good to check initial_min >= min,  but we still need the original check of initial_min < max as well. please not delete the original one ..

src/autoscaler/api/policyvalidator/policy_validator.go
```diff
@@ -199,45 +258,47 @@ func (pv *PolicyValidator) validateRecurringSchedules(policy *models.ScalingPoli
...
 			}
-			formatString := "instance_min_count {{.instance_min_count}} is higher or equal to instance_max_count {{.instance_max_count}}"
+			formatString := "specific_date[{{.scheduleIndex}}].instance_min_count {{.instance_min_count}} is higher or equal to specific_date[{{.scheduleIndex}}].instance_max_count {{.instance_max_count}}"
 			err := newPolicyValidationError(instanceMinContext, formatString, errDetails)
 			result.AddError(err, errDetails)
 		}
-		if specSched.ScheduledInstanceInit > specSched.ScheduledInstanceMax {
-			initialInstanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.initial_min_instance_count", i), specficDateScheduleContext)
+		if (specSched.ScheduledInstanceInit != 0) && (specSched.ScheduledInstanceInit < specSched.ScheduledInstanceMin) {
+			initialInstanceMinContext := gojsonschema.NewJsonContext(fmt.Sprintf("%d.initial_min_instance_count", scheduleIndex), specficDateScheduleContext)
```
cdlliuy: same as above. we still need to ensure initial_min <= max

src/autoscaler/api/policyvalidator/policy_validator.go
rohitsharma04: done.

src/autoscaler/api/server/api_handler.go
```diff
@@ -131,14 +141,51 @@ func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request,
...
+		if err != nil {
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error saving policy"})
+			return
+		}
+
+		h.logger.Info("creating/updating schedules", lager.Data{"policy": body.Policy})
+		err = h.schedulerUtil.CreateOrUpdateSchedule(body.AppID, body.Policy, policyGuid.String())
+		if err != nil {
```
cdlliuy: No, we still need to CreateOrUpdateSchedule before save the policy, since scheduler may return 400 bad request to reject the policy, see 
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/dd581153997db77d3d82709e918a074f91aeea6f/src/autoscaler/api/schedulerutil/schedulerutil.go#L69-L72
Also,  in this situation,  we should not return InternalServerError to end-user. 
I didn't take time to check scheduler validation further. If the `400 bad request` response won't happen at all, let us delete the validation in scheduler to avoid confusion.

src/autoscaler/api/server/api_handler.go
```diff
@@ -131,14 +141,51 @@ func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request,
...
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error creating/updating schedules"})
+			return
+		}
+	}
+
 	err = h.bindingdb.CreateServiceBinding(bindingId, instanceId, body.AppID)
```
cdlliuy: Referring to the original service broker implementation, we need to create service binding prior to the policy handling part. 
https://github.com/cloudfoundry-incubator/app-autoscaler/blob/master/servicebroker/lib/routes/binding.js#L63-L130

src/autoscaler/api/server/api_handler.go
```diff
@@ -131,14 +141,51 @@ func (h *ApiHandler) BindServiceInstance(w http.ResponseWriter, r *http.Request,
...
+		h.logger.Info("creating/updating schedules", lager.Data{"policy": body.Policy})
+		err = h.schedulerUtil.CreateOrUpdateSchedule(body.AppID, body.Policy, policyGuid.String())
+		if err != nil {
+			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+				Code:    "Interal-Server-Error",
+				Message: "Error creating/updating schedules"})
+			return
+		}
+	}
+
```
cdlliuy: If saving policy is succeed, but creating binding fails ..  we need to rollback the save policy ... otherwise, autoscaler will start to work the app without the binding.  It violates the service model ..  In previous service broker implementation,  all the actions here are bundled in a transaction.

src/autoscaler/api/server/api_handler.go
```diff
@@ -167,14 +214,23 @@ func (h *ApiHandler) UnbindServiceInstance(w http.ResponseWriter, r *http.Reques
 		return
 	}
 
-	if instanceId == "" || bindingId == "" || body.ServiceID == "" || body.PlanID == "" {
+	if body.AppID == "" || instanceId == "" || bindingId == "" || body.ServiceID == "" || body.PlanID == "" {
```
cdlliuy: for unbinding, appid is not required.

src/autoscaler/api/server/api_handler.go
```diff
@@ -188,6 +244,16 @@ func (h *ApiHandler) UnbindServiceInstance(w http.ResponseWriter, r *http.Reques
...
+
+	h.logger.Info("deleting schedules", lager.Data{"appId": body.AppID})
+	err = h.schedulerUtil.DeleteSchedule(body.AppID)
+	if err != nil {
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Error deleting schedules"})
+		return
+	}
+
```
cdlliuy: The same transaction consideration as above.

src/autoscaler/api/policyvalidator/policy_validator_test.go
```diff
@@ -0,0 +1,2344 @@
...
+							Context:     "(root).scaling_rules.0.cool_down_secs",
+							Description: "Must be less than or equal to 3600",
+						},
+					}))
+				})
+			})
+		})
+		Context("Schedules", func() {
+
+			Context("when timezone is missing", func() {
```
cdlliuy: how about adding another test case to check whether the timezone value is valid?

src/autoscaler/api/server/api_handler.go
rohitsharma04: I've removed the schedules validation from scheduler.

cdlliuy:
LGTM

paltanmoy:
LGTM

PR #450, opened 2019-01-24
fix bug "Wrong index and length parameters are provided when creating EnvelopeProcessors in metricsserver"

qibobo:
[#163445959]


paltanmoy:
LGTM

PR #456, opened 2019-02-08
Public Api with Oauth Protection

rohitsharma04:
This PR doesn't have policy handling endpoints as those endpoints require policy validation code from #449
I'll do a different PR for that after these #449 and this PR both are merged.

### DIFF
--- a/src/autoscaler/api/cmd/api/main.go
+++ b/src/autoscaler/api/cmd/api/main.go
@@ -7,6 +7,7 @@ import (
 
 	"autoscaler/api/brokerserver"
 	"autoscaler/api/config"
+	"autoscaler/api/publicapiserver"
 	"autoscaler/db"
 	"autoscaler/db/sqldb"
 	"autoscaler/helpers"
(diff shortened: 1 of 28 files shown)


src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func (oam *OAuthMiddleware) getUserScope(userToken string, userId string) ([]string, error) {
+	tokenEndpoint, err := oam.getCFTokenEndpoint()
+	if err != nil {
+		return nil, err
+	}
+
+	parameters := url.Values{}
+	parameters.Add("token", strings.Split(userToken, " ")[1])
```
boyang9527: check the token to avoid out of range

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -0,0 +1,242 @@
+package publicapiserver
```
boyang9527: we can move all this to cf package.

src/autoscaler/api/publicapiserver/public_api_handler.go
```diff
@@ -0,0 +1,257 @@
...
+		h.logger.Error("Failed to info file", err, lager.Data{"info-file-path": h.conf.InfoFilePath})
+		handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
+			Code:    "Interal-Server-Error",
+			Message: "Failed to load info"})
+		return
+	}
+	w.Write([]byte(info))
+}
+
+func (h *PublicApiHandler) GetHealth(w http.ResponseWriter, r *http.Request, vars map[string]string) {
```
boyang9527: looks like we always return "alive" then why we need it?

boyang9527: I would assume this should be prometheus.

cdlliuy:
@rohitsharma04 ,  will you add CURD of policy to public API endpoint in another PR?

src/autoscaler/api/publicapiserver/public_api_handler.go
```diff
@@ -0,0 +1,257 @@
...
+			Message: "metrictype is required",
+		})
+		return
+	}
+
+	path, _ := routes.EventGeneratorRoutes().Get(routes.GetAggregatedMetricHistoriesRouteName).URLPath("appid", appId, "metrictype", metricType)
+
+	url := h.conf.EventGenerator.EventGeneratorUrl + path.RequestURI() + "?" + parameters.Encode()
+
+	resp, err := h.metricsCollectorClient.Get(url)
```
cdlliuy: use eventgenClient?

src/autoscaler/api/publicapiserver/public_api_handler.go
```diff
@@ -0,0 +1,257 @@
...
+	handlers.WriteJSONResponse(w, resp.StatusCode, paginatedResponse)
+}
+
+func (h *PublicApiHandler) GetInstanceMetricsHistories(w http.ResponseWriter, r *http.Request, vars map[string]string) {
+	appId := vars["appId"]
+
+	metricType := vars["metricType"]
+	instanceIndex := r.URL.Query().Get("instance-index")
+
+	h.logger.Info("Get ScalingHistories", lager.Data{"appId": appId, "metricType": metricType, "instanceIndex": instanceIndex})
```
cdlliuy: Get InstanceMetricsHistories

src/autoscaler/api/publicapiserver/public_api_handler.go
cdlliuy: it is a legacy design issue.  Before moving to Prometheus,   there is a /health endpoint on api server to show uptime.  Also  autoscaler CLI is using the /health endpoint to determine whether the api server is accessible.  i.e.  CLI expects to get 200 OK with /health
https://github.com/cloudfoundry-incubator/app-autoscaler-cli-plugin/blob/master/src/cli/api/apihelper.go#L140

src/autoscaler/api/publicapiserver/oauth.go
rohitsharma04: done

cdlliuy:
@rohitsharma04 ,  could you please resolve the conflict?

rohitsharma04:
> @rohitsharma04 , could you please resolve the conflict?

@cdlliuy  done.

src/autoscaler/api/cmd/api/main.go
```diff
@@ -48,31 +51,49 @@ func main() {
...
-		logger.Error("failed to connect bindingdb database", err, lager.Data{"dbConfig": conf.DB.BindingDB})
+		logger.Error("failed to connect to policydb database", err, lager.Data{"dbConfig": conf.DB.PolicyDB})
 		os.Exit(1)
 	}
-	defer bindingDB.Close()
+	defer policyDb.Close()
+
+	if !conf.UseBuildInMode {
+		var bindingDB db.BindingDB
+		bindingDB, err = sqldb.NewBidingSQLDB(conf.DB.BindingDB, logger.Session("bindingdb-db"))
```
cdlliuy: BidingDB?  typo?

cdlliuy:
@rohitsharma04 ,  I guess you will have another PR to add policy CRUD in public API server, right?   If so,  this PR is LGTM

rohitsharma04:
> @rohitsharma04 , I guess you will have another PR to add policy CRUD in public API server, right? If so, this PR is LGTM

yes.

paltanmoy:
@rohitsharma04  Please create the separate PR for policy CRUD. 

LGTM

PR #459, opened 2019-02-14
fix doc's markdown format issue

cdlliuy:
To fix the incorrect display of markdown in some tables or table entries


qibobo:
LGTM

PR #460, opened 2019-02-15
Modified custom metrics tags

paltanmoy:


src/autoscaler/metricsserver/cmd/metricsserver/main.go
```diff
@@ -130,15 +130,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	serverNodeAddrs := make([]string, len(conf.NodeAddrs))
```
boyang9527: the NodeAddrs does not include port, that is the reason why the serverNodeAddrs is composed with NodeAddrs and port

paltanmoy: in metricscollector we construct the nodeAddress combining the address and port like below - 
`addrs=sorted_instances.map{|i| "'#{i.address}:#{p('autoscaler.metricscollector.server.port')}'"}
  nodeAddrs="[" + addrs.join(",") +"]"  
`
Do you suggest any other way of constructing the nodeAddress for metricsserver ?

boyang9527: for metrics server, we have both ports for websocket server and API server.  The nodeAddr is defined as the metric server config so it is not good to have port in it. 

I would suggest release code to have only address, not to add port.  This may have issue for testing multiple instances on same VM, but should be fine given we are not deploying multiple instances on one host.

paltanmoy: @boyang9527  I have added the serverNodeAddrs construction part back. 
@qibobo : Please take care of removing the port from job template as discussed.

qibobo:
LGTM

PR #468, opened 2019-03-18
metricsgateway should fail to start when emitter fails to connect metricsserver

qibobo:
[#164590471]

### DIFF
--- a/src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_test.go
+++ b/src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_test.go
@@ -68,17 +68,14 @@ var _ = Describe("Metricsgateway", func() {
 			})
 		})
 		Context("with missing configuration", func() {
-			var backupPolicyDBURL string = ""
 			BeforeEach(func() {
 				runner.startCheck = ""
-				backupPolicyDBURL = conf.AppManager.PolicyDB.URL
 				missingConfig := conf
 				missingConfig.AppManager.PolicyDB.URL = ""
-				runner.configPath = writeConfig(missingConfig).Name()
+				runner.configPath = writeConfig(&missingConfig).Name()
 			})
 
 			AfterEach(func() {
-				conf.AppManager.PolicyDB.URL = backupPolicyDBURL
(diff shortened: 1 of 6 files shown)


src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_test.go
```diff
@@ -70,15 +68,17 @@ var _ = Describe("Metricsgateway", func() {
 			})
 		})
 		Context("with missing configuration", func() {
+			var backupPolicyDBURL string = ""
```
boyang9527: why you do backup

src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_test.go
```diff
@@ -88,22 +88,32 @@ var _ = Describe("Metricsgateway", func() {
 			})
 		})
 	})
-
-	Describe("when an interrupt is sent", func() {
+	Describe("when it fails to connect to metricsserver when starting", func() {
+		var backupMetricsServerAddrs []string = []string{}
 		BeforeEach(func() {
-			runner.Start()
+			backupMetricsServerAddrs = conf.MetricServerAddrs
```
boyang9527: again why backup? the conf is not rewrite.

src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_test.go
qibobo: @boyang9527 conf is initialized in https://github.com/cloudfoundry/app-autoscaler/blob/ca95b4ac4e5eed8adc29e63f81cec444ae839b47/src/autoscaler/metricsgateway/cmd/metricsgateway/metricsgateway_suite_test.go#L91
so it will be overwritten in this case.

boyang9527: you have "missingConfig := conf" , this will not rewrite.

qibobo:
@boyang9527 updated.

PR #486, opened 2019-05-16
skip other triggers when one is fired.

boyang9527:


src/autoscaler/eventgenerator/generator/appEvaluationManager.go
```diff
@@ -42,44 +42,40 @@ func NewAppEvaluationManager(logger lager.Logger, evaluateInterval time.Duration
...
 	}, nil
 }
 
 func (a *AppEvaluationManager) getTriggers(policyMap map[string]*models.AppPolicy) map[string][]*models.Trigger {
 	if policyMap == nil {
 		return nil
 	}
-	triggersByType := make(map[string][]*models.Trigger)
+	triggersByApp := make(map[string][]*models.Trigger)
 	now := a.emClock.Now().UnixNano()
```
qibobo: It is better to put now into the for-loop as for-loop may take a little longer time to finish.

PR #487, opened 2019-05-20
make golang apiserver's public api the same as the nodejs apiserver(instance_metrics)

qibobo:


qibobo:
some code change to apply to the golang apiserver for it should behave the same as the nodejs one
1 When uaa returns 401, then apiserver should return 401.
2 When apiserver returns 401, the body should be an empty json object "{}".
3 While "order-direction" is just for public apiserver, it should be "order" when apiserver requests metricscollector.
4 The same as no.3, "instance-index" => "instanceindex".

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -64,18 +60,24 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
...
-			oam.logger.Error("failed to check spacedeveloper permissions", err, nil)
-			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
-				Code:    "Interal-Server-Error",
-				Message: "Failed to check space developer permission"})
-			return
+			if err == cf.ErrUnauthrorized {
+				http.Error(w, "{}", http.StatusUnauthorized)
+				return
+			} else {
+				oam.logger.Error("failed to check spacedeveloper permissions", err, nil)
```
cdlliuy: spacedeveloper ==> space developer

cdlliuy: I suggest to validate the `usertoken` first prior to `IsUserAdmin` check to return 401 error as early as possible if the user token is wrong

src/autoscaler/api/publicapiserver/route_helper.go
```diff
@@ -144,11 +136,11 @@ func paginateResource(resourceList []byte, parameters *url.Values, r *http.Reque
 	return result, nil
 }
 
-func getPageUrl(r *http.Request, parameters *url.Values, targetPageNo int) string {
+func getPageUrl(r *http.Request, queries *url.Values, targetPageNo int) string {
 	pageUrl, _ := url.Parse(r.URL.String())
 
 	pageParams := url.Values{}
-	for key, value := range *parameters {
+	for key, value := range *queries {
```
cdlliuy: The "queries" seems unnecessary, since it derives from "r".  Also, the queries & parameters here looks confusion.  The "pageNo" is still extracted from "parameter", then why we need to enforce the usage of "query" instead of "parameter" ?

src/autoscaler/cf/oauth.go
```diff
@@ -20,6 +20,9 @@ var (
 )
 
 func (c *cfClient) IsUserSpaceDeveloper(userToken string, appId string) (bool, error) {
+	if !c.isValidUserToken(userToken) {
```
cdlliuy: How about using "isValidUserToken()" directly in upper level to skip` IsUserSpaceDeveloper() ` and `IsUserAdmin()` directly with an invalid token?

cdlliuy:
LGTM

qibobo:
@boyang9527 could you take a look at this PR?

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -35,10 +32,13 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
 		userToken := r.Header.Get("Authorization")
 		if userToken == "" {
 			oam.logger.Error("userToken is not present", nil, lager.Data{"url": r.URL.String()})
-			http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+			http.Error(w, "{}", http.StatusUnauthorized)
```
boyang9527: you may add more description in the error json

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -48,34 +48,47 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
...
-		isUserAdmin, err := oam.cfClient.IsUserAdmin(userToken)
+		isUserSpaceDeveloper, err := oam.cfClient.IsUserSpaceDeveloper(userToken, appId)
 		if err != nil {
-			oam.logger.Error("failed to check if user is admin", err, nil)
-			handlers.WriteJSONResponse(w, http.StatusInternalServerError, models.ErrorResponse{
-				Code:    "Interal-Server-Error",
-				Message: "Failed to check if user is admin"})
-			return
+			if err == cf.ErrUnauthrorized {
+				http.Error(w, "{}", http.StatusUnauthorized)
```
boyang9527: same as above

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -48,34 +48,47 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
...
 			return
 		}
-		if isUserSpaceDeveloper {
+		if isUserAdmin {
 			next.ServeHTTP(w, r)
 			return
 		}
 
-		http.Error(w, http.StatusText(http.StatusUnauthorized), http.StatusUnauthorized)
+		http.Error(w, "{}", http.StatusUnauthorized)
```
boyang9527: same as above

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -48,34 +48,47 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
...
 		return
 	})
 }
+
+func (oam *OAuthMiddleware) isValidUserToken(userToken string) bool {
+	tokenSplitted := strings.Split(userToken, " ")
+	if len(tokenSplitted) != 2 {
```
boyang9527: don't you want to  check whether it is a bearer token?

src/autoscaler/models/api.go
```diff
@@ -21,3 +21,24 @@ type UnbindingRequestBody struct {
 	BrokerCommonRequestBody
 	AppID string `json:"app_guid"`
 }
+
+type PublicApiResultBase struct {
```
boyang9527: rename from "xxxResultxxx" to "xxxResponsexxx"

qibobo:
@boyang9527 updated.

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -32,11 +32,15 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
 		userToken := r.Header.Get("Authorization")
 		if userToken == "" {
 			oam.logger.Error("userToken is not present", nil, lager.Data{"url": r.URL.String()})
-			http.Error(w, "{}", http.StatusUnauthorized)
+			handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+				Code:    "Unauthorized",
+				Message: "Authorization is required"})
```
boyang9527: User token is not present in Authorization header

src/autoscaler/api/publicapiserver/oauth.go
```diff
@@ -32,11 +32,15 @@ func (oam *OAuthMiddleware) Middleware(next http.Handler) http.Handler {
...
 			return
 		}
 		if !oam.isValidUserToken(userToken) {
-			http.Error(w, "{}", http.StatusUnauthorized)
+			handlers.WriteJSONResponse(w, http.StatusUnauthorized, models.ErrorResponse{
+				Code:    "Unauthorized",
+				Message: "Authorization is invalid formated"})
```
boyang9527: "Invalid bearer token"

qibobo:
@boyang9527 fixed.

PR #491, opened 2019-06-12
Fixed too many log lines when db not available

MarcelMerkle:
This change fixes the problem of writing too many error messages when the policy database is not available. In our landscape this issue lead to problems with the elk stack because up to 10000 log lines / s were written. 
Now, when a database retrieve fails, the retry happens after the specified poll interval.


cdlliuy:
LGTM

PR #492, opened 2019-06-17
monitor metricsgateway nozzle status with prometheus

qibobo:


qibobo:
Below is a sample of the new metric:
```
# HELP autoscaler_metricsgateway_envelope_number_from_rlp Number of envelopes from rlp
# TYPE autoscaler_metricsgateway_envelope_number_from_rlp gauge
autoscaler_metricsgateway_envelope_number_from_rlp 1555
```

cdlliuy:
We need to clarify the purpose of this PR.  I think it is used to detect the possible nozzle error if the incoming envelop number over a period dropped significantly, or even become 0; 
To measure a "total number" over a period ,  we can refer to prometheus's common practice with increase(), 
i.e. https://prometheus.io/docs/prometheus/latest/querying/functions/#increase

In short, we need to just count the number of envelop by count() ;  then use prometheus's query to calculate avg envelop number over time. With this approach, we don't need to consider the observation interval at all.

PR #496, opened 2019-06-18
metricsgateway could output envelope streamer log

qibobo:


qibobo:
Currently the go-loggregator log is disabled. This PR will make the log output by lager:
`{"timestamp":"1560996102.127928972","source":"AppManager-test","message":"AppManager-test.Nozzle.envelope_streamer.Error connecting to Logs Provider: %s","log_level":1,"data":{"data":[{"code":14,"message":"all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: authentication handshake failed: x509: certificate is valid for reverselogproxy, not wrong-server-name\""}],"session":"1.1"}}`

cdlliuy:
Instead of creating a new config item for this , could we output the streamer log when log_level is set to "debug"?

cdlliuy:
LGTM

PR #497, opened 2019-06-19
when operator lost the dblock, it should exit

qibobo:
Currently when a lock owner lost the lock due to expired, it does not exit, so all its jobs keep running while all the competitor's jobs are running as the competitor has acquired the lock.
The right behavior is that when a lock owner lost the lock, all its jobs should exit and later it can be restarted to trying to acquire the lock and before it gets the lock, all its other jobs will not run.

### DIFF
--- a/src/autoscaler/sync/dblock.go
+++ b/src/autoscaler/sync/dblock.go
@@ -3,7 +3,9 @@ package sync
 import (
 	"autoscaler/db"
 	"autoscaler/models"
+
 	"os"
+	"sync/atomic"
 	"time"
 
 	"code.cloudfoundry.org/lager"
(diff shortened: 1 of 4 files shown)


cdlliuy:
LGTM

src/autoscaler/sync/dblock.go
```diff
@@ -3,31 +3,41 @@ package sync
...
+	"sync/atomic"
 	"time"
 
 	"code.cloudfoundry.org/lager"
 
 	"github.com/tedsuo/ifrit"
 )
 
+const (
+	OwnLock = iota
```
boyang9527: LockStatusHeld
LockStausReleased

src/autoscaler/sync/dblock.go
```diff
@@ -3,31 +3,41 @@ package sync
...
+	LostLock
+)
+
 type DatabaseLock struct {
 	logger lager.Logger
+	// isHoldLock represents the lock status of the last acquiring.
+	isHoldLock int32
```
boyang9527: status?

src/autoscaler/sync/dblock.go
```diff
@@ -36,7 +46,9 @@ func (dblock *DatabaseLock) InitDBLockRunner(retryInterval time.Duration, ttl ti
 		}
 		if isLockAcquired {
 			readyToAcquireLock = false
+			atomic.StoreInt32(&dblock.isHoldLock, OwnLock)
```
boyang9527: why atomic?

qibobo:
fixed.

PR #510, opened 2019-08-05
loose the breach/cooldown duration in policy validator

cdlliuy:
I just loose the min_value of breach and cooldown duration to 1 directly in this PR

### DIFF
--- a/api/config/settings.json
+++ b/api/config/settings.json
@@ -67,5 +67,7 @@
         "certFile": "../test-certs/api_public.crt",
         "caCertFile": "../test-certs/autoscaler-ca.crt"
     },
+    "minBreachDurationSecs": 30,
+    "minCoolDownSecs": 30,
     "httpClientTimeout": 5000
 }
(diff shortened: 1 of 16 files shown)


silvestre:
Would it maybe make sense to couple the minimum cooldown to `evaluation_manager_execute_interval` in the eventgenerator config? Because if not, I think it will be possible to set cooldowns lower than the interval in which rules will be evaluated, which could be confusing for users, and look like a bug to them.

boyang9527:
I am with Silvestre.   We can make them lower but they should not be lower than the aggregate interval.  If we allow cooldown to be very low, it will also cause issue when you are scale out autoscaler components.

cdlliuy:
@boyang9527  @silvestre  ,then we need to align the min value with `collect_interval`, `aggregator_execute_interval` and `evaluation_manager_execute_interval` ...  
If these 3 parameters  are different, then we need to choose the max of these three as the minimum breach_duration /cooldown setting ..    
And also, this limitation would change per deployment ...  The policy which works for deployment A will be invalid for deployment B ...    I think this is not good to setup API / policy specification ..

cdlliuy:
@boyang9527  @silvestre  @qibobo , please review the PR again.  
I changed the min breach/cooldown setting to be configurable in Node.js API server now.  
For Go API server,   the schema.json would be configurable in its bosh release job, so no need for further source code update.

silvestre:
LGTM

PR #514, opened 2019-08-15
Api/brokerserver: Allow hashed OSBAPI credentials

silvestre:
### Description:
Allow to specify the OSB API credentials for the service broker to be
specified as `bcrypt` hashes instead of needing to provide the clear
text.

#### Problems fixed:
* The credentials check was not done in constant time.
* The credentials needed to be provided as clear text.


cdlliuy:
LGTM

PR #516, opened 2019-08-22
api/brokerserver: Handle conflicting service instances

silvestre:
### Description:
Already existing service instances with conflicting parameters now result in a `409 Conflict`. Also the `dashboard_url` is now returned for non-conflicting service instances.

#### References:
* Spec: https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#response-3


cdlliuy:
LGTM

PR #529, opened 2019-09-06
enable db transaction timeout & rollback db transaction if there is an error

qibobo:


src/autoscaler/metricscollector/collector/collector.go
```diff
@@ -196,12 +195,15 @@ func (c *Collector) SaveMetricsInDB() {
 		case metric := <-c.dataChan:
 			metrics = append(metrics, metric)
 		case <-ticker.C():
-			go func(instancemetricsDb db.InstanceMetricsDB, metrics []*models.AppInstanceMetric) {
-				instancemetricsDb.SaveMetricsInBulk(metrics)
+			if c.persistMetrics && len(metrics) > 0 {
```
boyang9527: will empty metrics slice cause issue?

qibobo: @boyang9527 
Do you mean the following condition:

```
package main
import (
	"fmt"
)
func main() {
	var emptySlice []int = []int{}
	var nilSlice []int = nil
	fmt.Printf("the length of empty slice is %d\n", len(emptySlice))
	fmt.Printf("the leng of nil slice is %d\n", len(nilSlice))

}
```
The result is :
```
the length of empty slice is 0
the leng of nil slice is 0
```

cdlliuy: I guess @boyang9527  is just asking why the `len()>0` check is added here.  
In fact, just to avoid the necessary db connection/transition when there is nothing to save.

qibobo: @cdlliuy yes. I have checked with him.

PR #531, opened 2019-09-10
fix parameter parse bug

zyjiaobj:


src/autoscaler/api/policyvalidator/policy_validator_test.go
```diff
@@ -2343,7 +2343,7 @@ var _ = Describe("PolicyValidator", func() {
 					"threshold":90,
 					"operator":">=",
 					"cool_down_secs":300,
-					"adjustment":"+1"
+					"adjustment":"+10%"
```
cdlliuy: please add a new test case to validate "%"  as an adjust parameter , instead of modifying current "good" path.  We may different good path here.

src/autoscaler/api/publicapiserver/route_helper.go
```diff
@@ -76,8 +76,8 @@ func parseParameter(r *http.Request, vars map[string]string) (*url.Values, error
 		return nil, fmt.Errorf("results-per-page must be greater than 0")
 	}
 	parameters := &url.Values{}
-	parameters.Add("start-time", startTime)
-	parameters.Add("end-time", endTime)
+	parameters.Add("start", startTime)
```
cdlliuy: Any test case for this?   When the `parameters`  is wrong previously,  I believe the related integration test case should fail ..  Or maybe we miss the corresponding integration test for this part.

zyjiaobj: add test cases

src/integration/integration_golangapi_eventgenerator_test.go
```diff
@@ -259,7 +259,99 @@ var _ = Describe("Integration_GolangApi_EventGenerator", func() {
...
+						models.AppMetric{
+							AppId:      appId,
+							MetricType: models.MetricNameMemoryUsed,
+							Unit:       models.UnitMegaBytes,
+							Value:      "123456",
+							Timestamp:  666666,
+						},
+					},
+				}
+				By("check public api")
```
qibobo: please remove all `By("check public api")` as in golangapiserver integration tests there is no internal api.

cdlliuy:
LGTM

PR #535, opened 2019-10-17
ServiceBroker: Allow to specify a default policy on instance level

silvestre:
It is now possible to specify a default policy on service instance level
by passing it as a service creation parameter called `default_policy`.

This default policy applies to all service bindings that are created
without an explicit policy.

The default policy may be changed with `cf update-service` and the
changes will apply to all apps which do not have a any policy set, and, if
there was a previous default policy, also to those apps which still have the
default policy set.

If you explicitly set a policy which has the same _content_ as the
default policy on an app, this app will _not_ be regarded as having the
default policy set when the default policy is updated. If you want to
actually "set the default policy", you could set it by removing the
explicit policy and updating the default policy.

To remove the default policy, set it to an empty JSON object when
calling `cf update-service`.


silvestre:
Note: I changed the format of the service broker's liquibase changelog to YAML, and added it to golang api/db directory. The new file is used in the travis.yml but would need to also be reflected in the package spec in the release.

src/autoscaler/db/sqldb/binding_sqldb.go
```diff
@@ -50,43 +51,89 @@ func (bdb *BindingSQLDB) Close() error {
...
+			OrgId:             existingOrgId,
+			SpaceId:           existingSpaceId,
+			DefaultPolicy:     existingDefaultPolicy.String,
+			DefaultPolicyGuid: existingDefaultPolicyGuid.String,
+		}, nil
+	}
+	return nil, nil
+}
+
+func (bdb *BindingSQLDB) UpdateServiceInstance(serviceInstance models.ServiceInstance) error {
```
qibobo: missing UT for UpdateServiceInstance and GetAppIdsByInstanceId?

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -72,6 +72,29 @@ func (pdb *PolicySQLDB) GetAppIds() (map[string]bool, error) {
 	return appIds, nil
 }
 
+func (pdb *PolicySQLDB) GetAppIdsWithPolicy(policyGuid string) ([]string, error) {
```
qibobo: Rename the function to GetAppIdsWithPolicyGuid?

.travis.yml
```diff
@@ -29,7 +29,7 @@ before_script:
   - nvm use $NODE_VERSION
   - psql -c 'CREATE DATABASE autoscaler' -U postgres
   - java -cp 'db/target/lib/*' liquibase.integration.commandline.Main --url jdbc:postgresql://127.0.0.1/autoscaler --driver=org.postgresql.Driver --changeLogFile=api/db/api.db.changelog.yml update
-  - java -cp 'db/target/lib/*' liquibase.integration.commandline.Main --url jdbc:postgresql://127.0.0.1/autoscaler --driver=org.postgresql.Driver --changeLogFile=servicebroker/db/servicebroker.db.changelog.json update
+  - java -cp 'db/target/lib/*' liquibase.integration.commandline.Main --url jdbc:postgresql://127.0.0.1/autoscaler --driver=org.postgresql.Driver --changeLogFile=src/autoscaler/api/db/servicebroker.db.changelog.yaml update
```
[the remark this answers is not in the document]
silvestre: Done

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,171 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+	}
+
+	if serviceInstance == nil {
+		h.logger.Error("failed to find service instance to update", err, lager.Data{"instanceId": instanceId})
+		writeErrorResponse(w, http.StatusNotFound, "Failed to find service instance to update")
+		return
+	}
+
+	updatedDefaultPolicyGuid := ""
+	if updatedDefaultPolicy != "" {
```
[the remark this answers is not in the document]
silvestre: Good idea, thanks, done.

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,171 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+	}
+	updatedDefaultPolicy := string(*body.Parameters.DefaultPolicy)
+
+	if emptyJSONObject.MatchString(updatedDefaultPolicy) {
+		// accept an empty json object "{}" as a default policy update to specify the removal of the default policy
+		updatedDefaultPolicy = ""
+	}
+
+	serviceInstance, err := h.bindingdb.GetServiceInstance(instanceId)
+	if err != nil {
```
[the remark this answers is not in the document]
silvestre: Added some assertions for this usage of GetServiceInstance

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,171 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+			}
+			updatedAppIds, err := h.policydb.SetDefaultAppPolicy(appIds, updatedDefaultPolicy, updatedDefaultPolicyGuid)
+			if err != nil {
+				h.logger.Error("failed to set default policies", err, lager.Data{"instanceId": instanceId})
+				writeErrorResponse(w, http.StatusInternalServerError, "Error updating service instance")
+				return
+			}
+			// there is synchronization between policy and schedule, so errors creating schedules should not break
+			// the whole update process
+			for _, appId := range updatedAppIds {
```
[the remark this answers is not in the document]
silvestre: Added some assertions for that.

src/autoscaler/db/sqldb/binding_sqldb.go
silvestre: Oops, sorry. Added.

src/autoscaler/db/sqldb/policy_sqldb.go
silvestre: Removed it, by returning the affected rows in `ReplaceAppPolicies` and `DeletePoliciesByPolicyGuid`, which also reduces the number of queries.

silvestre:
The travis build failed due to 
```
The job exceeded the maximum log length, and has been terminated.
```

I'll try to split out the tests for the old components (node apiserver, metricscollector) and see if that helps.

qibobo:
@silvestre 
Maybe we can remove the nodejs apiserver&servicebroker and metricscollector relevant tests from travis build as they won't be updated.

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,166 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+		h.logger.Error("failed to update service instance when trying to get mandatory data", nil, lager.Data{"instanceId": instanceId, "serviceId": body.ServiceID, "planId": body.PlanID})
+		writeErrorResponse(w, http.StatusBadRequest, "Malformed or missing mandatory data")
+		return
+	}
+
+	if body.Parameters == nil || body.Parameters.DefaultPolicy == nil {
+		h.logger.Error("failed to update instance, only default policy updates allowed", nil, lager.Data{"instanceId": instanceId, "serviceId": body.ServiceID, "planId": body.PlanID})
+		writeErrorResponse(w, http.StatusUnprocessableEntity, "Failed to update service instance: Only default_policy updates allowed")
+		return
+	}
```
cdlliuy: Would we use the "update" service instance to delete `default policy` by passing through `nil` parameters?  
i.e.  at first , I may "create-service -c xx" at first, and do some binding,  then I want to delete the default policy but don't want to unbind & delete the service instance.  Could I achieve it through "update" service instance?

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,166 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+	updatedDefaultPolicy := string(*body.Parameters.DefaultPolicy)
+
+	if emptyJSONObject.MatchString(updatedDefaultPolicy) {
+		// accept an empty json object "{}" as a default policy update to specify the removal of the default policy
+		updatedDefaultPolicy = ""
+	}
```
cdlliuy: So we must remove the default policy by updating it to  "{}", right?   If no parameter is passing, could we treat it as the removal as well?   I am not sure whether we need to document this constraint in document.

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -80,6 +83,27 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
 		return
 	}
 
+	policyStr := ""
+	if body.Parameters.DefaultPolicy != nil {
+		policyStr = string(*body.Parameters.DefaultPolicy)
+	}
```
cdlliuy: I can't agree to Parameters.DefaultPolicy.  Referring to the bindrequestbody struct:
```
type BindingRequestBody struct {
	BrokerCommonRequestBody
	AppID  string          `json:"app_guid"`
	Policy json.RawMessage `json:"parameters,omitempty"`
}
```
we set the "parameters" directly as the policy.   But with `Parameters.DefaultPolicy` definition,  we are requesting to  the following JSON format:
```
{
 "default_policy":"xxx"
}
````
which is different from the binding experience.

File: src/autoscaler/db/sqldb/policy_sqldb.go [hunk 10]
```diff
@@ -134,6 +134,72 @@ func (pdb *PolicySQLDB) SaveAppPolicy(appId string, policyJSON string, policyGui
...
+	var appIds []string
+
+	query := "UPDATE policy_json SET guid = $2, policy_json = $3 WHERE guid = $1 RETURNING app_id"
+
+	rows, err := pdb.sqldb.Query(query, oldPolicyGuid, newPolicyGuid, newPolicy)
+	if err != nil {
+		pdb.logger.Error("replace-app-policy", err, lager.Data{"query": query, "oldPolicyGuid": oldPolicyGuid, "newPolicyGuid": newPolicyGuid, "newPolicy": newPolicy})
+		return nil, err
+	}
+
```
cdlliuy: rows.close()

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -104,6 +128,166 @@ func (h *BrokerHandler) CreateServiceInstance(w http.ResponseWriter, r *http.Req
...
+
+			// also set the new default policy on bound apps without a policy
+			allBoundApps, err := h.bindingdb.GetAppIdsByInstanceId(serviceInstance.ServiceInstanceId)
+			if err != nil {
+				h.logger.Error("failed to retrieve bound apps", err, lager.Data{"instanceId": instanceId})
+				writeErrorResponse(w, http.StatusInternalServerError, "Error updating service instance")
+				return
+			}
+			newAppsWithDefaultPolicy, err := h.policydb.SetDefaultAppPolicy(allBoundApps, updatedDefaultPolicy, updatedDefaultPolicyGuid)
+			if err != nil {
```
cdlliuy: If there was a "default policy" defined previously, and `bind-service` will enforce the "default policy" to bound apps ,  I think all the bound apps should have a policy already.  This part is not necessary. 

Furthermore,  if anything error happens here, whether we need to roll back previous update on default policy ?  Maybe we need a cron job to sync the service instance default policy record && policy db  as we did for scheduler ..

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -143,6 +209,27 @@ func (pdb *PolicySQLDB) DeletePolicy(appId string) error {
...
 }
 
+func (pdb *PolicySQLDB) DeletePoliciesByPolicyGuid(policyGuid string) ([]string, error) {
+	var appIds []string
+
+	query := "DELETE FROM policy_json WHERE guid = $1 RETURNING app_id"
+	rows, err := pdb.sqldb.Query(query, policyGuid)
+	if err != nil {
+		pdb.logger.Error("failed-to-delete-policies-by-policy-guid", err, lager.Data{"query": query, "policyGuid": policyGuid})
+		return nil, err
```
cdlliuy: rows.close()

src/autoscaler/api/brokerserver/broker_handler.go
silvestre: We could do this also, but we would need to distinguish between "no parameters passed" and no "default_policy passed".

> Note: The parameters specified are expected to be the values specified by an end-user. Whether the user chooses to include the complete set of configuration options or just a subset (or even none) is their choice. How a Service Broker interprets these parameters (including the absence of any particular parameter) is out of scope of this specification - with the exception that if this field is not present in the request then the Service Broker MUST NOT change the parameters of the instance as a result of this request.

https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-6

src/autoscaler/api/brokerserver/broker_handler.go
silvestre: We can, I just thought that it might be useful to leave us an option to introduce more parameters if needed. But we can also remove `default_policy`.

src/autoscaler/api/brokerserver/broker_handler.go
silvestre: The policy could have been removed from apps in the mean time. This gives the use to "set the default policy" by calling a service update.

I agree we need a rollback, but I don't think we need a sync. I will add the rollback

src/autoscaler/api/brokerserver/broker_handler.go
cdlliuy: As we discussed to remove the "default_policy" layer to make the policy definition format consistent,  here we only need to discuss the exact cmd to trigger the policy removal. 
To make it more clearly,   I listed the possible cmds as below: 
`cf update-service <instance-name>`   --> Expected result:  Remove?  or treat it as "no changes" ? 
`cf update-service <instance-name>  -c ""`  --> Expected result:  Invalid parameter? 
`cf update-service <instance-name>  -c "{}" ` --> Expected result:  remove policy

cdlliuy:
@silvestre ,  for the update-service-instance logic, how about using the below flow: 
```
if updatedDefaultPolicyGuid != "" {
//however , we need to update or insert policy here. 
allBoundApps, err := h.bindingdb.GetAppIdsByInstanceId(serviceInstance.ServiceInstanceId)
//get all bound app list first

newAppsWithDefaultPolicy, err := h.policydb.UpdateDefaultAppPolicy(allBoundApps, updatedDefaultPolicy, updatedDefaultPolicyGuid)
//for each bound app, we either insert the default policy , update default policy or skip if the policy guid is not the same
.... 
}   else {
     if serviceInstance.DefaultPolicyGuid != "" {
          //remove the default policy
     }
}
```

With above flow, you don't need to repeat the `SetDefaultAppPolicy` in different switches,  also combined all changes in policyDB into a single function, so that you don't need to pass through transactions cross functions.    I hope the transaction related stuffs would be hidden in db layer, instead of bringing it to broker_handler.go  
Just a suggestion .

silvestre:
@cdlliuy Thanks for the suggestion! I have added a commit with a change.

However removing `default_policy` is a no-go at the moment: Already the `cf` cli does not pass an empty JSON object:

So:
> `cf update-service <instance-name>` --> Expected result: Remove? or treat it as "no changes" ?
> `cf update-service <instance-name> -c ""` --> Expected result: Invalid parameter?
> `cf update-service <instance-name> -c "{}" ` --> Expected result: remove policy

Are all No-Ops as enforced by the `cf` cli.

I would vote to stay with the `default_policy`

cdlliuy:
@silvestre , let us use "default_policy" then.

cdlliuy:
@silvestre , what will happen with the following sequence:
```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice  
cf detach-autoscaling-policy app 
```
In current implementation, the default policy will be removed. 


Or using another example
```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice   -c <policy>
cf detach-autoscaling-policy app 
```
in current implementation,   the <default-policy> won't be apply to <app> after the detach.  

This behaviour seems to conflict with the "default policy" implementation in bind-service.

silvestre:
I think it depends on how you interpret "default policy".

In my mind it means "the policy to apply if no other policy was explicitly set"
In this case the above results are consistent as the user explicitly removed the policy.

I guess you could take it to mean "the policy that applies if there is not a more specific policy".
In that case I suppose the following results would be expected:

```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice  
cf detach-autoscaling-policy app 
=> ERROR: Cannot detach default policy
```

```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice   -c <policy>
cf detach-autoscaling-policy app 
=> The default policy applies
```

I'm not sure if one is better than the other, other than the latter implementation does not allow some states (apps without policies when a default policy exists) ...

cdlliuy:
@silvestre ,  when I read your response for the first time, I tended to claim in document that "the default policy is just the policy to apply if no other policy was explicitly set, but not a constraint" , that means we won't apply default policy after the policy is detached.   
But then,  I thought of another sequence:
```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice  
cf detach-autoscaling-policy app 
=> ERROR: Cannot detach default policy
cf update-service -c <another default policy>
```
In this case,  if the default policy doesn't apply to the app,  then with the `update-service`, the new default policy will be applied.  It looks a little bit conflict again. 

So,   I tend to use 
```
cf create-service xxx xxx asservice -c <default-policy>
cf bind-service  app asservice   -c <policy>
cf detach-autoscaling-policy app 
=>  The default policy applies
cf update-service -c <another default policy> 
```
since it looks consistent.    Your idea?

cdlliuy:
@silvestre , any update for the latest comment? Also, the branch is diverted..  You need to rebase.

src/autoscaler/api/brokerserver/broker_handler.go
```diff
@@ -117,7 +265,7 @@ func (h *BrokerHandler) DeleteServiceInstance(w http.ResponseWriter, r *http.Req
 	if err != nil {
 		if err == db.ErrDoesNotExist {
 			h.logger.Error("failed to delete service instance: service instance does not exist", err,
-				lager.Data{"instanaceId": instanceId})
+				lager.Data{"instanceId": instanceId})
```
cdlliuy, line 268: when re-reading the `deleteserviceinstance` implementation, I just realized that when a default policy exists,  if we just remove the service instance, then an un-used policy will be left behind in the policy db forever.  Is my understanding correct?  @silvestre ?

silvestre:
@aqan213 Thank you very much for the comments, the tests now also run on MySQL

src/autoscaler/api/brokerserver/broker_handler.go
silvestre: No, you cannot remove a service instance without deleting all bindings first.

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -155,6 +264,69 @@ func (pdb *PolicySQLDB) DeletePolicy(appId string) error {
...
+
+	// then actually delete them
+	query = tx.Rebind("DELETE FROM policy_json WHERE guid = ?")
+	_, err = tx.Exec(query, policyGuid)
+	if err != nil {
+		pdb.logger.Error("failed-to-delete-policies-by-policy-guid", err, lager.Data{"query": query, "policyGuid": policyGuid})
+		return nil, err
+	}
+	defer func() {
+		if err := rows.Close(); err != nil {
```
cdlliuy, line 317: @silvestre , why have two defer func for `rows.Close()` ?

src/autoscaler/db/sqldb/policy_sqldb.go
```diff
@@ -146,6 +146,115 @@ func (pdb *PolicySQLDB) SaveAppPolicy(appId string, policyJSON string, policyGui
...
+		_, err = tx.Exec(query, newPolicyGuid, newPolicy, oldPolicyGuid)
+		if err != nil {
+			pdb.logger.Error("rollback-set-or-update-app-policies", err, lager.Data{"query": query, "oldPolicyGuid": oldPolicyGuid, "newPolicyGuid": newPolicyGuid, "newPolicy": newPolicy})
+			return nil, err
+		}
+		defer func() {
+			if err := rows.Close(); err != nil {
+				pdb.logger.Error("rollback-set-or-update-app-policies-close-rows", err, lager.Data{"query": query, "oldPolicyGuid": oldPolicyGuid, "newPolicyGuid": newPolicyGuid, "newPolicy": newPolicy})
+			}
+		}()
```
cdlliuy, line 221: @silvestre , besides this, we have another defer in line 184?

KevinJCross (approved):

silvestre:
LGTM, thanks @KevinJCross !

PR #541, opened 2019-11-15
MetricsCollector: Only process HTTPStartStopEvents with peerType client

silvestre:
### Description:
Only process `HTTPStartStop` events with `peerType` `Client`, as they
contain the app instance index in the `instanceIndex` field.

#### Problems fixed:
`HTTPStartStop` events with `peerType` `Server` would be counted against
instance `0` and thus duplicate the number of requests counted by the
App Autoscaler.

### DIFF
--- a/src/autoscaler/metricscollector/collector/app_streamer.go
+++ b/src/autoscaler/metricscollector/collector/app_streamer.go
@@ -107,7 +107,7 @@ func (as *appStreamer) processEvent(event *events.Envelope) {
 	} else if event.GetEventType() == events.Envelope_HttpStartStop {
 		as.logger.Debug("process-event-get-httpstartstop-event", lager.Data{"event": event})
 		ss := event.GetHttpStartStop()
-		if ss != nil {
+		if ss != nil && ss.GetPeerType() == events.PeerType_Client {
 			as.numRequests[ss.GetInstanceIndex()]++
 			as.sumReponseTimes[ss.GetInstanceIndex()] += (ss.GetStopTimestamp() - ss.GetStartTimestamp())
 		}
--- a/src/autoscaler/metricsgateway/nozzle.go
+++ b/src/autoscaler/metricsgateway/nozzle.go
@@ -121,7 +121,7 @@ func (n *Nozzle) filterEnvelopes(envelops []*loggregator_v2.Envelope) {
 					n.envelopChan <- e
 				}
 			case *loggregator_v2.Envelope_Timer:
-				if e.GetTimer().GetName() == "http" {
+				if e.GetTimer().GetName() == "http" && e.GetDeprecatedTags()["peer_type"].GetText() == "Client" {
 					n.logger.Debug("filter-envelopes-get-httpstartstop", lager.Data{"index": n.index, "appID": e.SourceId, "message": e.Message})
 					n.envelopChan <- e
 				}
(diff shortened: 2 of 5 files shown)


src/autoscaler/metricsgateway/nozzle.go
```diff
@@ -121,7 +121,7 @@ func (n *Nozzle) filterEnvelopes(envelops []*loggregator_v2.Envelope) {
 					n.envelopChan <- e
 				}
 			case *loggregator_v2.Envelope_Timer:
-				if e.GetTimer().GetName() == "http" {
+				if e.GetTimer().GetName() == "http" && e.GetDeprecatedTags()["peer_type"].GetText() == "Client" {
```
qibobo: @silvestre 
**1** According to the document [peertype](https://github.com/cloudfoundry/dropsonde-protocol/blob/ce0e609a7a57f041b74e661f13d32e5f7a673a5d/events/README.md#peertype).

Name | Number | Description
-- | -- | --
Client | 1 | Request is made by this process.
Server | 2 | Request is received by this process.

`e.GetDeprecatedTags()["peer_type"].GetText() == "Client" ` means we only care the requests made by the application. I think we should count the requests received by the application, right?
**2** And  method GetDeprecatedTags() contains "Deprecated", so maybe one day there is no "peer_type" tag so the nozzle will skip all http event?

silvestre: @qibobo : Client/server is from the viewpoint of the application emitting the event, in this case the `gorouter`. We should only count requests made by the gorouter to the application, when it has decided which app instance to target. The other event is emitted when the request is received and has no information on which app instance will process the request.

With regards to the deprecation, I had assumed you had already discussed this with the routing team, as just two lines above we already check for the origin which is also in the deprecated tags: https://github.com/cloudfoundry/app-autoscaler/blob/a911a395ac4a2caede4b63adccc8eb6dc0b32fa5/src/autoscaler/metricsgateway/nozzle.go#L119

I will ping the routing team to get their thoughts on the deprecation strategy.

silvestre: Raised the deprecation question in https://cloudfoundry.slack.com/archives/CFX13JK7B/p1574065292072500

qibobo: @silvestre 
How about to check wether the "peer_type" tag exists first and then check the value? So that when it is removed, it will not break the feature.

silvestre: @qibobo Good idea, implemented in 4f5e3aa24e7db8af1aae9bb96e73c6642b7ba70b

src/autoscaler/metricscollector/collector/app_streamer.go
```diff
@@ -107,7 +107,7 @@ func (as *appStreamer) processEvent(event *events.Envelope) {
 	} else if event.GetEventType() == events.Envelope_HttpStartStop {
 		as.logger.Debug("process-event-get-httpstartstop-event", lager.Data{"event": event})
 		ss := event.GetHttpStartStop()
-		if ss != nil {
+		if ss != nil && ss.GetPeerType() == events.PeerType_Client {
```
qibobo: Do we need to check whether peertype exists as the nozzle does?

cdlliuy:
@silvestre , rebase please :-)

src/autoscaler/metricscollector/collector/app_streamer.go
silvestre: No, that is not necessary as it already the default: https://github.com/cloudfoundry/sonde-go/blob/04fbe57ef6878d0050682b5b0697091003ebf960/events/http.pb.go#L262-L267

PR #543, opened 2019-11-26
Add mysql support for scheduler component

aqan213:
To support Mysql, 

- [x] Change all varchar to varchar(length) in all liquibase changelog file

- [x] Set the sequence column type to bigserial in Postgres and set it to bigint(auto_increment) in MySQL.

- [x] Add quartz mysql file.

- [x] Add mysql test in  `.travis.yml` file


.travis.yml
```diff
@@ -14,6 +14,7 @@ jdk:
 - openjdk8
 services:
   - postgresql
+  - mysql
```
qibobo: mysql is installed in "before_install", is the mysql service here still necessary?

qibobo:
Please only update the changelog files of scheduler, because for other ones, the changes can not be verified now.

.travis.yml
aqan213: I have removed `mysql` from service.

aqan213:
I have removed other changelog files and just updated the file of scheduler and api . Since the scheduler also needs to access `policy_json` table.

scheduler/db/scheduler.changelog-master.yaml
```diff
@@ -3,13 +3,14 @@ databaseChangeLog:
...
       author: Fujitsu
+      dbms: postgresql
       changes:
         - createTable:
             tableName: app_scaling_specific_date_schedule
             columns:
               - column:
                   name: schedule_id
-                  type: bigint
+                  type: bigserial
```
cdlliuy: is there any `upgrade` test executed to validate whether the `type` changes take effect on an existing db?

scheduler/db/scheduler.changelog-master.yaml
```diff
@@ -228,4 +233,160 @@ databaseChangeLog:
...
+      id: 7
+      author: aqan213
+      dbms: mysql
+      changes:
+        - createTable:
+            tableName: app_scaling_specific_date_schedule
+            columns:
+              - column:
+                  name: schedule_id
+                  type: bigint
```
cdlliuy: Myabe my knowledge on db site is not enough, but why change `bigint` to `bigserial` in postgres, and use `bigint` again in mysql ?

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestDataDbUtil.java
```diff
@@ -190,4 +199,25 @@ private int convertArrayToBits(int[] values) {
...
+	public enum DatabaseType {
+		POSTGRESQL, MYSQL,
+	}
+
+	public DatabaseType getDatabaseTypeFromDataSource() throws Exception {
+		if (this.databaseType != null) {
+			return this.databaseType;
+		}
+		String driverName = this.dataSource.getConnection().getMetaData().getDriverName().toLowerCase();
+		if (driverName.indexOf("postgresql") > -1) {
```
cdlliuy: check  driverName not null and not empty before doing any further operation.

scheduler/db/scheduler.changelog-master.yaml
aqan213: Made changes directly in the original changelog will cause error when upgrading db. So we add the changes in new changesets.

And the `bigserial` is not the real data type,  it can not use `ALTER ..`  to change the type. 

According to the document https://www.postgresql.org/docs/9.6/datatype-numeric.html#DATATYPE-SERIAL, we add the default value for the column`schedule_id` to `nextval('app_scaling_recurring_schedule_seq')`.

I did the `upgrade` test for the `type` changes:
steps:
1) Run the original liquibase changelog file and create the `autoscaler` db and also the related tables.
2) Run the modified changelog file and check the data type.
```
autoscaler=# \d app_scaling_specific_date_schedule;
                                      Table "public.app_scaling_specific_date_schedule"
           Column           |            Type             | Collation | Nullable |                  Default
----------------------------+-----------------------------+-----------+----------+-------------------------------------------
 schedule_id                | bigint                      |           | not null | nextval('schedule_id_sequence'::regclass)
```
3. Run unit test and all test case passed.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestDataDbUtil.java
aqan213: Added the check for dirverName.

scheduler/db/scheduler.changelog-master.yaml
aqan213: Now all the `bigint` will not be changed in postgres instead of setting the default value of `schedule_id` to `nextval('schedule_id_sequence')` .

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/util/TestDataDbUtil.java
qibobo: @cdlliuy @aqan213 
If we need to do empty/null check, so we need to check dataSource, getConnection, getMetaData, getDriverName to ensure the return value is not null.

scheduler/db/scheduler.changelog-master.yaml
cdlliuy: we may still need to create `app_scaling_active_schedule`, see changeset 5.   Also, I wonder why missing `app_scaling_active_schedule` won't result in UT test failure.  Can you check the related UT as well?

qibobo: @cdlliuy There is no dbms condition for app_scaling_active_schedule, so it will be created for all dbms.

cdlliuy:
LGTM

Issue #553, opened 2019-12-17
Schedules can not be removed in synchronization when an application is deleted directly


qibobo:
**Description:**
When an application is deleted without detaching its policy:
1 operator will delete the policy from policydb
2 scheduler will try to delete the schedules and notify scalingengine. Now scalingengine returns an error when the app is not found, then scheduler fails to delete the schedules.

qibobo:
**Description:**
When an application is deleted without detaching its policy:
1 operator will delete the policy from policydb
2 scheduler will try to delete the schedules and notify scalingengine. Now scalingengine returns an error when the app is not found, then scheduler fails to delete the schedules.

aqan213:
The fix has already been in release 3.0.1.

PR #554, opened 2019-12-20
bug fix: Schedules can not be removed in synchronization when an application is deleted directly #553

qibobo:


src/autoscaler/scalingengine/scalingengine.go
```diff
@@ -330,6 +331,11 @@ func (s *scalingEngine) RemoveActiveSchedule(appId string, scheduleId string) er
 
 	appEntity, err := s.cfClient.GetApp(appId)
 	if err != nil {
+		if _, ok := err.(*models.AppNotFoundErr); ok {
+			logger.Error("app-not-found", err)
+			history.Status = models.ScalingStatusIgnored
+			return nil
```
cdlliuy: Why still log error here ... maybe just log info here since it is a type of expected error.

cdlliuy:
LGTM

PR #561, opened 2020-04-14
basic authentication added to healthcheck

aadeshmisra:


cdlliuy:
@aadeshmisra  thanks for the PR .   Just wonder, what is the motivation of this PR?  
Ideally,  the health endpoint didn't have any sensitive information exposed .  Is that security enforcement ?

zyjiaobj:
@aadeshmisra Can we make the auth check to be configurable, if the username and password are provided in the config file, then do the auth check; if the username or password are not defined, we can skip the check. Or we can introduce another config item named `enable_basic_check` to be clearer.
In this way, this change can smoothly deployed with no breaking to our current monitoring mechanism.

aadeshmisra:
@cdlliuy Information exposed for version, debugging & profiling are also treated as sensitive information at SAP. Therefore we introduced this in the open source version too.

@zyjiaobj In case the username and password options are not defined, then the basic authentication is not used. It is only used once both the values are provided.

zyjiaobj:
@aadeshmisra Thanks! Could you please add some test cases for the scenario? When username and password are not provided, the requests to `/health` should always success.

aadeshmisra:
@zyjiaobj The tests without the authentication already existed, I just added the basic authentication tests.

src/autoscaler/models/health.go
```diff
@@ -1,5 +1,48 @@
...
 type HealthConfig struct {
-	Port int `yaml:"port"`
+	Port                    int    `yaml:"port"`
+	HealthCheckUsername     string `yaml:"username"`
+	HealthCheckUsernameHash string `yaml:"username_hash"`
+	HealthCheckPassword     string `yaml:"password"`
+	HealthCheckPasswordHash string `yaml:"password_hash"`
+}
+
+func (c *HealthConfig) Validate(component string) error {
```
qibobo, line 17: I wonder if we need a component name here. The log of each component writes to its own file, so user won't get confused which component has error.

aqan213:
LGTM. @aadeshmisra It will need to update the app-autoscaler-release and add the imported lib to the packages.
`- app-autoscaler/src/golang.org/x/crypto/bcrypt/* # gosub
- app-autoscaler/src/golang.org/x/crypto/blowfish/* # gosub`

src/autoscaler/models/health.go
aadeshmisra: @qibobo  I think that is a reasonable argument. I will do the necessary changes.

qibobo:
@aadeshmisra travis failed.

PR #574, opened 2020-07-20
Update Documentation for Emit metric API Spec

rahuldeepattri:
Hi Maintainers, I was working on custom metrics API and found an issue in the documentation [here](https://github.com/cloudfoundry/app-autoscaler/blob/develop/docs/Public_API.rst#put-v1appsguidmetrics)

image

As we can see above in the header section the request is documented as a **PUT** request while in route section it is documented as **POST** request. The HTTP request should be same in both the cases as shown above. I tried for my cf provider and only POST request works.

Fixed the misleading documentation for Submit custom metric to Autoscaler metric server API

### DIFF
--- a/docs/Public_API.rst
+++ b/docs/Public_API.rst
@@ -1100,7 +1100,7 @@ Status
 **Submit custom metric to Autoscaler metric server**
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-**PUT /v1/apps/:guid/metrics**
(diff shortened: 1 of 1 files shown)


cdlliuy:
@rahuldeepattri  thank you for the PR!  but you don't have easyCLA signed .. 

Can you fix it? or we can go ahead to fix the wrong docs directly.

rahuldeepattri:
Hi @cdlliuy  Sure will do! I had some problems in EasyCLA UI. Please let me try it again.

rahuldeepattri:
@cdlliuy Done!

PR #575, opened 2020-07-29
Allow uaa clients interact with autoscaler api

daviddob:
This changes the auth flow to check if a token has the admin scope before checking space developer access. This cuts down on the number of requests to uaa as well as allows clients with cloud_controller.admin to use the autoscaler api where currently clients (even those with admin scope) will fail to auth based on the fact they cant be space developers as well as do not have openid scope for UAA.

If a client connects without admin scope the behavior remains the same as currently implemented with the response being "Failed to check space developer permission". This may not be ideal, however it quickly enables the functionality required while not introducing breaking changes.


aqan213:
@daviddob the travis test was failed. can you please fix it? thanks,

silvestre (changes requested):
I don't think it is a good idea to first check for admin permissions - generally I would expect there to be way more space developers than admins so the check for space developer permission should be the first with the admin check as a fallback.

To be honest I could not reproduce the error with an admin user - I could access the App Autoscaler API without issues.

What was the exact error? Maybe we can start from there.

daviddob:
The issue with checking space developer first is that client_ids for service accounts do not have the openID scopes and the current check results in a 500 currently as they are not full users. A client with cloud_controller.admin should be able to use the autoscaler-api and not need to be a full user with space developer permissions. This is currently not possible.

lnguyen:
If the tests were fixed, can this be merged in? are there app autoscaler maintainers? @silvestre ?

silvestre:
Yes, sure! We are currently working on switching the tests over to Github Actions (#584), as travis-ci.org is sunsetting, but when the tests are green we can get to merging this.

lnguyen:
@silvestre can we get approved to run tests?

lnguyen (approved):
Can this be approved? Per Dave comments?

silvestre:
@lnguyen @daviddob Thank your for your contribution! 

It's very much appreciated. 

We should be quicker from now on to get such requests merged. :smiley:

silvestre (approved):

PR #577, opened 2020-08-21
Fix typo in custom metric docs

michalc:
There are potentially other related changes to be done: it feels to me that "custom metric" is a bit clunky, and suspect should be "custom metrics" or "a custom metric" in most cases, but deliberately not addressing that, and just fixing what seems to clearly be a typo,


cdlliuy (approved):
LGTM

PR #580, opened 2020-11-11
update golang version from 1.13 to 1.15 in the doc

aqan213:
The golang used in autoscaler was upgraded to 1.15.1.  If using old version,  will got error when running the test case.


silvestre (approved):
LGTM

PR #584, opened 2021-06-15
Added a simple GitHub action flow for pull requests

garethjevans:
* Only running against MySQL at the moment, Postgress to follow.
* TravisCI builds are still running
* Only running unit tests at the moment


silvestre:
Failing travis tests are related to outdated test certificates.
We would need to run [scheduler/scripts/generate_unit_test_certs.sh](https://github.com/cloudfoundry/app-autoscaler/blob/main/scheduler/scripts/generate_unit_test_certs.sh) to get them green and it should be needed anyways later on.

silvestre:
There's a separate script for the test certs used by the Java part, which I linked in the previous comment. I guess it should be included in the top level script for better usability.

garethjevans:
> There's a separate script for the test certs used by the Java part, which I linked in the previous comment. I guess it should be included in the top level script for better usability.

Ah, I think I've worked out what you mean - I was looking at the script in the wrong directory.  Should those certs be added to source control or generated for each build?

silvestre:
I think it would be cleaner to generate them when the test is run, and not check them in. Up to you if you want to tackle this in this PR, or we should leave it as a follow-up item.

garethjevans:
> I think it would be cleaner to generate them when the test is run, and not check them in. Up to you if you want to tackle this in this PR, or we should leave it as a follow-up item.

I can tackle it in this PR, there will be a few more commits as I'm adding the postgres work.

silvestre (approved):
LGTM

silvestre (approved):

garethjevans:
Closing this to recreate - hoping it will allow the triggering of PRs

PR #588, opened 2021-06-16
Convert autoscaler to use go modules

garethjevans:
Some notes for the reviewer:
* To ensure everything is included in the go.mod files, build tools have been added to `tools.go`, they can be installed with the make target `make buildtools`
* Integrations tests have been moved into the src/autoscaler directory
* Dependencies are imported as the `Incompatible` versions for v2+ modules - this will be rectified in a later PR.
* Integration tests use the pre-built go binaries to avoid building them more than once.
* Config unmarshalling uses the `UnmarshalStrict` to generate the expected errors for the tests


src/autoscaler/metricsgateway/nozzle_test.go
```diff
@@ -197,7 +197,8 @@ var _ = Describe("Nozzle", func() {
 				LogServerName = "wrong-server-name"
 			})
 			It("should output the err", func() {
-				Eventually(logger.Buffer).Should(Say("certificate is valid for reverselogproxy, not wrong-server-name"))
+				// Eventually(logger.Buffer).Should(Say("certificate is valid for reverselogproxy, not wrong-server-name"))
+				Eventually(logger.Buffer).Should(Say("Error connecting to Logs Provider"))
```
silvestre, line 201: Just out of curiosity: Why did this change? I suppose "... is valid for ...,not ..." comes from the standard library and we now check for the error that is generated by `go-loggregator`. Did they modify the error message?

silvestre:
Maybe we should document that `make build` is required to run the integration tests in https://github.com/cloudfoundry/app-autoscaler/blob/main/README.md#integration-tests

silvestre (approved):
LGTM, maybe you can address my comments when you rebase this PR

src/autoscaler/metricsgateway/nozzle_test.go
garethjevans: Yeah, it looks like the error coming from `go-loggregator` is coming from https://github.com/cloudfoundry/go-loggregator/blob/5adba004bd0139991d9e89899c0ceeb92db75054/envelope_stream_connector.go#L197, the older version which we using before (although I not 100% which version is was) seemed to use a `v2.NewClient` to connect which is why we may have got the error message directly.

garethjevans: I do think there is a bug in the `go-loggregator` code though, as the cause of the error is not passed in the message (although it looks like it should be), I'm going to try and update to the /v8 version in a later PR that fixes the incompatible imports

silvestre: Sounds like a plan! :+1:

src/autoscaler/integration/integration_suite_test.go
```diff
@@ -168,27 +167,30 @@ var _ = BeforeEach(func() {
...
 
-	builtExecutables[GolangAPIServer], err = gexec.BuildIn(rootDir, "autoscaler/api/cmd/api", "-race")
-	Expect(err).NotTo(HaveOccurred())
+	// builtExecutables[MetricsGateway], err = gexec.BuildIn(rootDir, "autoscaler/metricsgateway/cmd/metricsgateway", "-race")
+	builtExecutables[MetricsGateway] = path.Join(rootDir, "src", "autoscaler" , "build", "metricsgateway")
+
+	// builtExecutables[MetricsServerHTTP], err = gexec.BuildIn(rootDir, "autoscaler/metricsserver/cmd/metricsserver", "-race")
+	builtExecutables[MetricsServerHTTP] = path.Join(rootDir, "src", "autoscaler" , "build", "metricsserver")
+
+	// builtExecutables[GolangAPIServer], err = gexec.BuildIn(rootDir, "autoscaler/api/cmd/api", "-race")
```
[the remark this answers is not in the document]
garethjevans: done

PR #634, opened 2021-07-06
Simplify application context

asalan316:
In preparation for allowing unit tests to be run with h2, we need to simplify how the application context is loaded.


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/HealthConfig.java
```diff
@@ -0,0 +1,37 @@
+package org.cloudfoundry.autoscaler.scheduler.conf;
+
+import org.apache.commons.dbcp2.BasicDataSource;
+import org.cloudfoundry.autoscaler.scheduler.health.DBStatusCollector;
+import org.cloudfoundry.autoscaler.scheduler.health.HealthExporter;
+import org.graalvm.compiler.lir.BailoutAndRestartBackendException;
```
garethjevans, line 6: this think this import needs to be removed

garethjevans (changes requested):

garethjevans (approved):

garethjevans (approved):

garethjevans:
Quartz in Postgres seems to be failing with:
```
Caused by: org.springframework.scheduling.SchedulingException: Could not start Quartz Scheduler; nested exception is org.quartz.SchedulerConfigException: Failure occured during job recovery. [See nested exception: org.quartz.JobPersistenceException: Couldn't retrieve trigger: Bad value for type long : \x [See nested exception: org.postgresql.util.PSQLException: Bad value for type long : \x]]
```

garethjevans:
@asalan316 the integration tests will need to be updated with the new quartz properties in here https://github.com/cloudfoundry/app-autoscaler/blob/5da771ce496e03628fa6b0e50884ebc4378b1073/src/autoscaler/integration/components.go#L344-L346

PR #646, opened 2021-07-16
fix: enable errcheck linter

garethjevans:


src/autoscaler/db/sqldb/appmetric_sqldb.go
```diff
@@ -83,35 +83,38 @@ func (adb *AppMetricSQLDB) SaveAppMetricsInBulk(appMetrics []*models.AppMetric)
...
+			_ = txn.Rollback()
 			return err
 		}
 
 	case "mysql":
 		sqlStr := "INSERT INTO app_metric(app_id,metric_type,unit,timestamp,value)VALUES"
 		vals := []interface{}{}
 		if len(appMetrics) == 0 {
-			txn.Rollback()
+			err = txn.Rollback()
```
KevinJCross, line 114: not sure what this is about ?
I think you can just put _ = ?

src/autoscaler/eventgenerator/aggregator/aggregator.go
```diff
@@ -106,7 +106,7 @@ func (a *Aggregator) startSavingAppMetric() {
 		case <-ticker.C():
 			if len(appMetricArray) > 0 {
 				go func(appMetricDB db.AppMetricDB, metrics []*models.AppMetric) {
-					appMetricDB.SaveAppMetricsInBulk(metrics)
+					_ = appMetricDB.SaveAppMetricsInBulk(metrics)
```
KevinJCross, line 109: maybe we should log this error ?

src/autoscaler/metricsgateway/helpers/ws_helper_test.go
```diff
@@ -108,10 +108,9 @@ var _ = Describe("WsHelper", func() {
 			err = wsHelper.Ping()
 			Expect(err).ShouldNot(HaveOccurred())
 			Eventually(pingPongChan, 5*time.Second, 1*time.Second).Should(Receive(Equal(1)))
-
 		})
 		It("close the websocket connection", func() {
-			wsHelper.CloseConn()
+			err = wsHelper.CloseConn()
 			Expect(err).NotTo(HaveOccurred())
```
KevinJCross, line 114: this was checking the error that was set in errr before the action ... maybe we should do 2 checks ?

src/autoscaler/db/sqldb/appmetric_sqldb.go
garethjevans: I've updated it to `return err`

src/autoscaler/metricsserver/collector/collector.go
```diff
@@ -214,7 +214,7 @@ func (c *metricCollector) saveMetrics() {
 		case <-ticker.C():
 			if c.PersistMetrics && len(metrics) > 0 {
 				go func(instancemetricsDb db.InstanceMetricsDB, metrics []*models.AppInstanceMetric) {
-					instancemetricsDb.SaveMetricsInBulk(metrics)
+					_ = instancemetricsDb.SaveMetricsInBulk(metrics)
```
KevinJCross, line 217: same as the other comment

src/autoscaler/scalingengine/scalingengine.go
```diff
@@ -68,7 +68,9 @@ func (s *scalingEngine) Scale(appId string, trigger *models.Trigger) (*models.Ap
 		Reason:       getDynamicScalingReason(trigger),
 	}
 
-	defer s.scalingEngineDB.SaveScalingHistory(history)
+	defer func() {
+		_ = s.scalingEngineDB.SaveScalingHistory(history)
```
KevinJCross, line 72: hmmm should we log the error here ?

KevinJCross: or maybe SaveScalingHistory should log it if it does not already

src/autoscaler/metricsgateway/helpers/ws_helper_test.go
garethjevans: the previous error is checked in the `BeforeEach` func

src/autoscaler/scalingengine/scalingengine.go
KevinJCross: it logs it anyway

garethjevans: done

KevinJCross (approved):

PR #683, opened 2021-08-17
AAS-75: Slight improvement github actions and test-cert scripts.

KevinJCross:
- removed the test certs and added it the gitignore so that we dont keep getting modifications to commit.


bonzofenix (approved):
lgtm

PR #684, opened 2021-08-18
Fix: enabled the prometheus metrics endpoint and fixed tests to properly assert

KevinJCross:
To make this PR work in the release it requires this PR to be merged before releaseing
https://github.com/cloudfoundry/app-autoscaler-release/pull/314


style-guide/inspect-java-format-0.1.sh
```diff
@@ -4,21 +4,20 @@
...
 ###################################################################################################
 set -e -o pipefail
-set +x
 
 REPO_PATH=$(git rev-parse --show-toplevel)
 DOWNLOAD_PATH="$REPO_PATH"/.cache
 CHECKSTYLE_JAR_NAME="checkstyle-8.44-all.jar"
-GOOGLE_JAR_NAME="google-java-format-1.11.0-all-deps.jar"
 GOOGLE_JAR_VERSION="1.11.0"
+GOOGLE_JAR_NAME="google-java-format-${GOOGLE_JAR_VERSION}-all-deps.jar"
```
KevinJCross, line 12: so the version is not in 2 places

style-guide/google-format-ci-v0.1.sh
```diff
@@ -1,17 +1,20 @@
 #!/usr/bin/env bash
-
+set -eu
 ###################################################################################################
  # This script downloads google formatter and displays formatting issues on Github actions
 ###################################################################################################
-
-curl -LJO "https://github.com/google/google-java-format/releases/download/v$GOOGLE_JAR_VERSION/$GOOGLE_JAR_NAME"
+GOOGLE_JAR_VERSION=${GOOGLE_JAR_VERSION:-"1.11.0"}
```
KevinJCross, line 6: So I can run it over all the code locally intead of using commit hook ... which is not working for me for some reason

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/health/SchedulerHealthEndpointTest.java
```diff
@@ -20,26 +21,29 @@
...
-    assertThat(result.contains("jvm_memory_pool_bytes"));
-    assertThat(result.contains("autoscaler_scheduler_data_source"));
-    assertThat(result.contains("autoscaler_scheduler_policy_db_data_source"));
+  public void greetingShouldReturnDefaultMessage() {
+    ResponseEntity<String> response =
+        this.restTemplate.getForEntity(
+            "http://localhost:" + metricsConfig.getPort() + "/metrics", String.class);
+    assertThat(response.getStatusCode().value()).isEqualTo(200);
+    String result = response.toString();
+    assertThat(result)
```
KevinJCross, line 33: this test ever actually asserted anything .... 😆 ... the assertThat is the new version import which required assertThat(xx).contains(yy)

style-guide/google_checks.xml
```diff
@@ -41,7 +41,8 @@
 
   <module name="LineLength">
     <property name="fileExtensions" value="java"/>
-    <property name="max" value="100"/>
+    <!-- intellij standard page line -->
+    <property name="max" value="150"/>
```
asalan316, line 45: I assume, that 120 is the hard wrap as default line Length in IntelliJ (atleast in Ultimate version )

style-guide/google-format-ci-v0.1.sh
asalan316: You might need to install Precommit on your local system .e.g, `brew install pre-commit`

KevinJCross: yeah I have installed pre-commit as in the docs but it still does not do  the pre-commit hooks :/

style-guide/google_checks.xml
KevinJCross: sure ... I just think something above 100 is good for me my standard page is 150.
I feel its better to be able to not force line breaks but leave it to the developer to decide what is more readable but have a limit as not to go crazy.

scheduler/pom.xml
```diff
@@ -164,7 +185,8 @@
                 <configuration>
                     <disableXmlReport>true</disableXmlReport>
                     <redirectTestOutputToFile>true</redirectTestOutputToFile>
-                    <runOrder>random</runOrder>
+<!--                    <runOrder>random</runOrder>-->
```
garethjevans, line 188: I think the random order was intentional to try to determine race conditions?

garethjevans: @KevinJCross not sure if you saw my comment here, but I think this line was deliberately set, presumably so it mimics the ginkgo configuration and runs tests in a random order each time.

KevinJCross: hi yeah ... I think this change snuck in as I was trying to work out why we fork the jvm for each test class which means spring has to build a new context for each test. This is very slow. 
I will remove this change :)

aadeshmisra (approved):
LGTM

scheduler/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "properties": [
+    {
+      "name": "scheduler.metrics.port",
```
garethjevans, line 4: should this be `healthserver`?

KevinJCross: yeah we missed committing that by mistake.
This particular file is more for editors and verifiers on the properties files thats why no tests picked it up.

PR #692, opened 2021-08-25
Added basic auth options to the prometheus metrics endpoint.

KevinJCross:


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/MetricsConfig.java
```diff
@@ -1,15 +1,42 @@
...
+  private AuthProvider getAuthProvider(MetricsConfiguration config) {
+    return (JsonObject authInfo, Handler<AsyncResult<User>> resultHandler) -> {
+      String password = authInfo.getString("password");
+      String username = authInfo.getString("username");
+      if (config.getUsername().equals(username) && config.getPassword().equals(password)) {
+        resultHandler.handle(Future.succeededFuture(null));
+      } else {
+        resultHandler.handle(Future.failedFuture("Not authorised"));
+      }
+    };
```
silvestre, line 40: Please check if this change can be replaced by using the new `Authenticator` support in [simpleclient_httpserver](https://github.com/prometheus/client_java/pull/682/files#diff-05c7094a6207ad292780607cc229e38691d89448909d71ab461b379ec0b1eb9dR246-R257)

The we would not need to add the VertX as dependency.

silvestre (changes requested):

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/MetricsConfig.java
bonzofenix: We will take a look at it, thank you

KevinJCross: @silvestre good spot we have now reverted back to using the built in httpserver.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/MetricsConfig.java
```diff
@@ -1,15 +1,27 @@
...
-    return new HTTPServer(config.getPort());
+  @Bean(destroyMethod = "close")
+  HTTPServer metricsServer(MetricsConfiguration config) throws IOException {
+    Builder builder = new Builder().withPort(config.getPort());
+    if (config.isBasicAuthEnabled()) {
+      builder.withAuthenticator(
+          new BasicAuthenticator("/") {
+            @Override
+            public boolean checkCredentials(String username, String password) {
+              return config.getUsername().equals(username) && config.getPassword().equals(password);
```
KevinJCross, line 21: use constant time string comparison.

PR #711, opened 2021-09-07
basic authentication added to healthcheck

garethjevans:


.github/workflows/generate.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
           pushd src/autoscaler
             make buildtools
             make generate
+            git diff
```
KevinJCross, line 24: would prefer this to only come out if there is an error

KevinJCross (approved):

PR #713, opened 2021-09-09
Re-implement quota API support in golang service broker.

bonzofenix:
Signed-off-by: Cross, Kevin <kevin.cross@sap.com>


.github/workflows/postgres.yaml
```diff
@@ -9,9 +9,9 @@ jobs:
   postgres:
     strategy:
       matrix:
-        go: [1.16]
+        go: [1.15]
         java: [11]
-        postgres: [9.6, 12]
+        postgres: [12]
```
garethjevans, line 14: we still need to support 9.6

.github/workflows/postgres.yaml
```diff
@@ -9,9 +9,9 @@ jobs:
   postgres:
     strategy:
       matrix:
-        go: [1.16]
+        go: [1.15]
```
garethjevans, line 12: we run with golang 1.16.7 in production

garethjevans (dismissed):

.github/workflows/postgres.yaml
asalan316: Yes, I think, It was mistakenly changed becuase of act

.github/workflows/postgres.yaml
asalan316: Yes, because of debugging locally

.github/workflows/postgres.yaml
KevinJCross: Im trying 1.15 to see if the problem happens on 1.15.

.tool-versions
```diff
@@ -0,0 +1,2 @@
+golang 1.16.8
+java adoptopenjdk-jre-11.0.12+7
```
asalan316, line 2: I think, we need JDK (instead of JRE) to compile Java cod.

garethjevans: @asalan316 correct - its waiting in here https://github.com/cloudfoundry/app-autoscaler/pull/720 for review, but I've got no idea why these commits are in here?

garethjevans:
There are a pile of commits on this PR that aren't related to this? I'm not sure how they got there.  They should be removed and the PR squashed

src/autoscaler/go.mod
```diff
@@ -1,6 +1,6 @@
 module autoscaler
 
-go 1.15
+go 1.16
```
garethjevans, line 3: This is likely to cause issues until https://github.com/cloudfoundry/app-autoscaler-ci/blob/126386f314db9fe93827f429cbfb3667ac96640a/dockerfiles/autoscaler-test/Dockerfile#L43-L51 is updated (assume the deploy image needs the same)

aadeshmisra (approved):
LGTM

src/autoscaler/api/config/config.go
```diff
@@ -68,15 +68,26 @@ type MetricsForwarderConfig struct {
...
+	Logging            helpers.LoggingConfig `yaml:"logging"`
+	BrokerServer       ServerConfig          `yaml:"broker_server"`
+	PublicApiServer    ServerConfig          `yaml:"public_api_server"`
+	DB                 DBConfig              `yaml:"db"`
+	BrokerUsername     string                `yaml:"broker_username"`
+	BrokerUsernameHash string                `yaml:"broker_username_hash"`
+	BrokerPassword     string                `yaml:"broker_password"`
+	BrokerPasswordHash string                `yaml:"broker_password_hash"`
+	APIClientId        string                `yaml:"api_client_id"`
+	//TODO: add yaml tag
```
garethjevans, line 89: we should fix this

PR #783, opened 2021-11-08
refactor credential by interface

asalan316:


src/autoscaler/api/custom_metrics_cred_helper/custom_metric_cred_helper.go
```diff
@@ -12,6 +12,28 @@ const (
 	MaxRetry = 5
 )
 
+type Credentials interface {
+	CreateCredential(appId string, userProvidedCredential *models.Credential, policyDB db.PolicyDB, maxRetry int) (*models.Credential, error)
```
garethjevans, line 16: We should remove `policyDB db.PolicyDB` from the interface methods and add these to the constructor

KevinJCross (approved):
I would prefer a better name than cred_helper as it lends itself to the helpers antipatern

PR #787, opened 2021-11-11
Loads plugin from config file for metrics forwarder

bonzofenix:
Signed-off-by: Gareth Evans <g.evans@sap.com>


src/autoscaler/api/config/config.go
```diff
@@ -224,6 +225,9 @@ func (c *Config) Validate() error {
 		if c.CatalogPath == "" {
 			return fmt.Errorf("Configuration error: CatalogPath is empty")
 		}
+		if c.CredHelperPluginPath == "" {
```
KevinJCross, line 228: does this break backwards compat ?

garethjevans: yep

garethjevans (approved):

PR #788, opened 2021-11-12
Attempt to shutdown plugin correctly

garethjevans:
To validate:

```
TEST=api make test-autoscaler-suite
```

and monitor with:

```
ps -ef | grep cred-helper | grep -v ps | grep -v grep
```

### DIFF
--- a/src/autoscaler/api/cmd/api/api_test.go
+++ b/src/autoscaler/api/cmd/api/api_test.go
@@ -28,6 +28,7 @@ var _ = Describe("Api", func() {
 	})
 
 	AfterEach(func() {
+		// then destroy
 		runner.KillWithFire()
 	})
 
@@ -110,6 +111,10 @@ var _ = Describe("Api", func() {
 	})
 
 	Describe("BuildIn Mode", func() {
+		AfterEach(func() {
+			runner.Interrupt()
+			Eventually(runner.Session, 5).Should(Exit(0))
+		})
 		Context("BuildIn Mode is false", func() {
 			BeforeEach(func() {
 				basicAuthConfig := cfg
@@ -120,6 +125,7 @@ var _ = Describe("Api", func() {
 			It("should start both broker and public-api", func() {
 				Eventually(runner.Session.Buffer, 2*time.Second).Should(Say("api.broker_http_server.broker-http-server-created"))
 				Eventually(runner.Session.Buffer, 2*time.Second).Should(Say("api.public_api_http_server.public-api-http-server-created"))
+				Eventually(runner.Session.Buffer, 2*time.Second).Should(Say("api.started"))
 			})
 		})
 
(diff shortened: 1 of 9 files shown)


src/autoscaler/cred_helper/loader.go
```diff
@@ -15,9 +16,23 @@ import (
 
 type PluginManager struct {
 	client *plugin.Client
+
+	// ctx is used to shutdown the device manager
+	ctx    context.Context
```
KevinJCross, line 21: do we use this context anywhere ?
we should probably  be using it in the RPC client requests ?

KevinJCross: OK ... so https://www.quora.com/What-is-the-difference-between-kill-command-and-interrupt-a-process
suggests that TERM should be caught and kill(9) should not be called.
So we should be calling sig term or `.Terminate()` and killwithFire should not be used (SEG_F/SIGKILL)

src/autoscaler/metricsforwarder/cmd/metricsforwarder/main.go
```diff
@@ -64,13 +64,14 @@ func main() {
 	}
 	defer policyDB.Close()
 
-	pm := cred_helper.PluginManager{}
+	pm := cred_helper.New()
+	defer pm.Kill()
```
KevinJCross, line 68: hmmm ... with the terminate does the plugin close the db connection it creates ?
Actually we should be doing this in the defer too.

src/autoscaler/api/cmd/api/api_test.go
```diff
@@ -28,6 +28,7 @@ var _ = Describe("Api", func() {
 	})
 
 	AfterEach(func() {
+		// then destroy
 		runner.KillWithFire()
```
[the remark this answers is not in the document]
KevinJCross: in sigmon.New it adds TERM and Interupt
```

func New(runner ifrit.Runner, signals ...os.Signal) ifrit.Runner {
	signals = append(signals, syscall.SIGINT, syscall.SIGTERM)
	return &sigmon{
		Signals: signals,
		Runner:  runner,
	}
}
```

So we can use either interrupt or term intead of `KillWithFire` ... typically Ive used the kill -9 when the process gets stuck terminating ... so we should attempt to do 
`Terminate()` Poll for termination for x seconds and if this does not terminate we should then call `kill -9` or `Kill()`

KevinJCross:
(x) Change needed. We need to replace `KillWithFIre()` withn preferably Terminate() or Interrupt().
Because `KillWithFIre()` terminates immediately and does not allow gracefull shutdown (closing connections) and should only be used in frozen or broken processes

garethjevans:
@KevinJCross the terminate change looks good, but I can't approve my own PR

src/autoscaler/api/cmd/api/main.go
```diff
@@ -80,13 +80,14 @@ func main() {
 		os.Exit(1)
 	}
 
-	pm := cred_helper.PluginManager{}
+	pm := cred_helper.New()
```
asalan316, line 83: Is the plugin manager needed although we have only one plugin here?

KevinJCross (approved):

asalan316 (approved):

PR #791, opened 2021-11-15
WIP: consider different options

asalan316:


docs/cred_helper_plugin.md
```diff
@@ -0,0 +1,89 @@
+# Discussion around choosing a Pluggable Architecture
+
+## Goal:
+
+Separate out some business functionality.
```
garethjevans, line 5:
```suggestion
The goal is to provide a method/pattern by which we can replace some of the OSS functionality with proprietary functionality.  The pattern should be as lightweight as possible and although initially will only be used for the credentials/authentication system, it may be used elsewhere in the future.
```

garethjevans (changes requested):
I'd add to the goals a bit

PR #794, opened 2021-11-17
Initial impl of stored procedure based binding creds

garethjevans:


src/autoscaler/api/cmd/api/main.go
```diff
@@ -80,10 +81,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	pm := cred_helper.PluginManager{}
+	pm := plugin.PluginManager{}
```
garethjevans, line 84: rename to `plugin.Manager{}`

src/autoscaler/db/db.go
```diff
@@ -14,6 +14,7 @@ const (
 	MysqlDriverName    = "mysql"
 	PolicyDb           = "policy_db"
 	BindingDb          = "binding_db"
+	StoredProcedureDb  = "storedprocedure_db"
```
garethjevans, line 17: should add a struct for this?

src/autoscaler/db/sqldb/storedprocedure_sqldb.go
```diff
@@ -0,0 +1,120 @@
...
+	"fmt"
+
+	"code.cloudfoundry.org/app-autoscaler/src/autoscaler/db"
+	"code.cloudfoundry.org/app-autoscaler/src/autoscaler/models"
+
+	"code.cloudfoundry.org/lager"
+	_ "github.com/lib/pq"
+)
+
+// FIXME should be moved to a common config
```
garethjevans, line 14: move

src/autoscaler/api/cmd/api/main.go
```diff
@@ -80,10 +81,10 @@ func main() {
 		os.Exit(1)
 	}
 
-	pm := cred_helper.PluginManager{}
+	pm := plugin.Manager{}
 	defer pm.Kill()
 
-	credentials, err := pm.LoadCredentialPlugin(conf.DB, conf.Logging, conf.CredHelperPluginPath)
+	credentials, err := pm.LoadCredentialsImplementation(conf.DB, conf.Logging, conf.CredHelperPlugin, conf.StoredProcedureConfig)
```
KevinJCross, line 87: hmmm it would seem to be much  simpler if we had a simple if ...
```
if conf.StoredProcedureConfig {
 credentials, err = NewStoredProc()
} else if conf.CredHelperPlugin {
 credentials, err = NewCredHelper()
} else {
  err = ERROR 
}
```
where `NewCredProcedure` is a the constructor for stored procs  
    and `NewCredPlugin` is the plugin constructor (which also does the `LoadCredentialsImplementation` internally)

There is no reall need for the plugin manager in the case of stored procedures.

src/autoscaler/cred_helper/plugin/loader.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (p *Manager) LoadCredentialsImplementation(dbConfigs map[db.Name]db.DatabaseConfig, loggingConfig helpers.LoggingConfig, credHelperPlugin string, storedProcedureConfig *models.StoredProcedureConfig) (cred_helper.Credentials, error) {
+	if credHelperPlugin == "" {
+		return nil, errors.New("credHelperPlugin is not configured")
+	}
+
+	switch credHelperPlugin {
+	case credHelperDefault:
+		logger := helpers.InitLoggerFromConfig(&loggingConfig, "default_cred_helper")
```
KevinJCross, line 46: ah ok so your side stepping in the default case.

src/autoscaler/cred_helper/plugin/loader.go
```diff
@@ -0,0 +1,100 @@
...
+		p.client.Kill()
+	}
+}
+
+func (p *Manager) LoadCredentialsImplementation(dbConfigs map[db.Name]db.DatabaseConfig, loggingConfig helpers.LoggingConfig, credHelperPlugin string, storedProcedureConfig *models.StoredProcedureConfig) (cred_helper.Credentials, error) {
+	if credHelperPlugin == "" {
+		return nil, errors.New("credHelperPlugin is not configured")
+	}
+
+	switch credHelperPlugin {
```
KevinJCross, line 44: Id consider putting this outside of the load since it is supposed to be loading the plugin not newing different instances

asalan316:
Improvement suggestions are good. but as a first step, we want to check whether it is working (cut a release).

asalan316 (approved):

PR #797, opened 2021-11-18
Remove cred-helper plugin impl

garethjevans:


src/autoscaler/api/cmd/api/api_suite_test.go
```diff
@@ -206,7 +202,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	cfg.RateLimit.MaxAmount = 10
 	cfg.RateLimit.ValidDuration = 1 * time.Second
 
-	cfg.CredHelperPlugin = pluginPath
+	cfg.CredHelperPlugin = "default"
```
asalan316, line 205: may be rename to credHelperImpl as we are not using the plugin anymore

asalan316 (approved):

PR #801, opened 2021-11-21
Reworked Credentials interface to add Validate func

garethjevans:


src/autoscaler/cred_helper/storedprocedure_cred_helper.go
```diff
@@ -64,6 +64,7 @@ func (c *storedProcedureCredentials) Delete(appId string) error {
 	}
 }
 
-func (c *storedProcedureCredentials) Get(appId string) (*models.Credential, error) {
-	return nil, nil
+func (c *storedProcedureCredentials) Validate(appId string, credential models.Credential) error {
```
KevinJCross, line 67: hmmm ... is the error used a  control for valid or not.
Im a little conscerned about mixing control flow and errors here.
What about the case when the process fails but the credentials are valid i.e. a database error ?
Can we return a 500 error when this happens ?

garethjevans: done in latest

joergdw (approved):

PR #842, opened 2025-11-01
Creation of service-keys

joergdw:
This contains the "app-autoscaler"-changes for <https://github.com/cloudfoundry/app-autoscaler-release/pull/3652>.

Current top-commit-id of that branch: "9ca68503e2f08bd8123df39b5722decb0fcde1bb"

This PR comes with the following changes:
 + Service-key-support
    - Introduce the necessary handling into the `Bind`-function of the broker.
    - Unit-tests
    - Divide the schema into three logical units for three different scenarios, where the "meta.schema.json" brings all of them together:
       * Creation of a service-key using the default-policy while enforcing the usage of the field `schema-version` (only allowed value: "0.9").
       * Creation of a service-key with explicitly set policy while enforcing the usage of the field `schema-version`.
       * All the other use-cases that have been supported until now allowing to skip the `schema-version`. This is for backwards-compatibility.
 + Multi-schema support for binding-requests:
    - Separate logic between request-handling and request-parsing. Introduce the interface `BindRequestParser` for request-parsers. This is an architectural improvement to simplify maintenance even for a growing variance and seconds the seperation of the internal representation of scaling-policies from their json-representation.
    - Adapt tests to the slighly changed error-messages. This is necessary to support the old schema as well. For acceptance-tests these changes are minimal.
 + Makefiles:
    - Fix usage of shell-styles
    - (Re-)Generate fakes and gorouterproxy only if relevant input-files have changed.
    - Fix dependencies (e.g. for `autoscaler.test`) and recipes.
    - Use programs provided by devbox instead of `go run …`.
    - Fix target `clean`.
 + Efficiency-improvement: Avoid parsing of the default credential-type for the communication with the metricsforwarder on each request.

### DIFF
--- a/acceptance/app/dynamic_policy_test.go
+++ b/acceptance/app/dynamic_policy_test.go
@@ -414,7 +414,7 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 
 				// Part-validation setup
 				By("Starting disk usage to trigger scale out")
-				helpers.StartDiskUsage(cfg, appToScaleName, 800, 5)
+				helpers.StartDiskUsage(cfg, appToScaleName, 800, 6)
 
 				// Validation
 				helpers.WaitForNInstancesRunning(appToScaleGUID, 2, 6*time.Minute)
@@ -425,7 +425,7 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 				helpers.StopDiskUsage(cfg, appToScaleName, 0)
 
 				// Validation
-				waitingTime := 8 * time.Minute // This validation can take a bit longer for unkown reasons.
+				waitingTime := 8 * time.Minute // This validation can take a bit longer for unknown reasons.
 				helpers.WaitForNInstancesRunning(appToScaleGUID, 1, waitingTime)
 			})
 		})


Makefile
```diff
@@ -1,8 +1,8 @@
 SHELL := /bin/bash
 .SHELLFLAGS := -eu -o pipefail -c
 MAKEFLAGS := -s
-aes_terminal_font_yellow := \e[38;2;255;255;0m
-aes_terminal_reset := \e[0m
+aes_terminal_font_yellow := \033[38;2;255;255;0m
+aes_terminal_reset := \033[0m
```
joergdw: Unrelated! For unknown reasons `\e` has stopped working on the shell.

api/policyvalidator/meta.schema.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "http://json-schema.org/draft-05/schema#",
+  "title": "Autoscaler Policy JSON Schema",
+  "description": "Unification of different schemas for scaling-policies for Autoscaler",
+
+  "anyOf": [
+    {"$ref": "./scaling-policy.schema.json"},
+    {"$ref": "./legacy.schema.json"},
+    {"$ref": "./service-key_only.schema.json"}
+  ]
```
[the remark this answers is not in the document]
joergdw: I thought a bit about this: We require backwards-compatibility. When you define `anyOf` in combination with the legacy-schema, then it is (trivially) guaranteed that you have backwards compatibility because the legacy-schema is one of the entries in the `anyOf`-construction. When you use `oneOf` it is not so trivial. And it has the potential to be wrong: Our legacy-schema is quite open to extension (due to missing `  "additionalProperties": false` all over the file) – and therefore may overlap with newer schemas.

This is indeed a problem. As this test-case shows we currently can not distinguish in some cases between legacy-schema and the newer schema. I think I have found a different solution: I restrict the usage of `schema-version` in the legacy-schema to the single value `null`. By that it will be always distinguishable on this single field from other schema-versions.

api/broker/broker_test.go
```diff
@@ -458,10 +460,344 @@ var _ = Describe("Broker", func() {
...
+					Expect(customMetricsStrategy.String()).To(Equal(models.DefaultCustomMetricsStrategy.String()))
+
+					// Verify policy was saved with the correct app GUID
+					Expect(fakePolicyDB.SaveAppPolicyCallCount()).To(Equal(1))
+					_, savedAppId, savedPolicy, savedPolicyGuid := fakePolicyDB.SaveAppPolicyArgsForCall(0)
+					Expect(savedAppId).To(Equal("12345678-abcd-1234-5678-123456789abc"))
+					Expect(savedPolicy).NotTo(BeNil())
+					Expect(savedPolicyGuid).NotTo(BeEmpty())
+				})
+			})
```
joergdw: 🤖⚠️ Attention: This part has been generated by using claude-sonnet-4 as a templating-engine.

api/policyvalidator/meta.schema.json
```diff
@@ -0,0 +1,11 @@
+{
```
geigerj0: is this intended that we are able to create empty bindings against apps without any configuration? 

Image

api/policyvalidator/service-key_only.schema.json
```diff
@@ -0,0 +1,23 @@
+{
```
geigerj0: shall we rename this file to `service-key_only.v0_9.schema.json` as well?

api/broker/broker.go
```diff
@@ -651,54 +658,40 @@ func (b *Broker) Bind(
...
-
-	logger.Debug("getOrDefaultCredentialType", lager.Data{"credential-type": credentialType})
-	return credentialType, nil
+	appData, err := b.cfClient.GetApp(ctx, appGUID)
+	if err != nil {
+		logger.Error("cf-client-get-app", err, lager.Data{"appGUID": appGUID})
+		return apiresponses.NewFailureResponse(
+			fmt.Errorf("internal error"), http.StatusInternalServerError, "create-service-key")
+	}
+	appInSpace := models.GUID(appData.Relationships.Space.Data.Guid) == instanceSpaceGuid
```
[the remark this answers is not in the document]
joergdw: Yes, welcome to Golang, where two principles (control of stack/heap-usage vs presence of a value) get mixed together.

We will fail with a 500 as I require this to be in the answer of the cloudcontroller.

api/policyvalidator/service-key_only.v0_9.schema.json
```diff
@@ -0,0 +1,23 @@
+{
```
joergdw: The idea is that the service-key-only-schema is just a variant that is part of the v0.9-schema. And `anyOf` is the operator that clues together both v0.9-variants: The one that is only used for simple service-key-creation (without policies) and the other one that is used for everything else in that v0.9-scheme.

Providing a metric-submission-strategy just makes sense if you provide a policy that makes use of those metrics you want to submit. We currently don't allow setting the policy-fields interactively one by one.

api/policyvalidator/meta.schema.json
```diff
@@ -0,0 +1,11 @@
...
+  "$schema": "http://json-schema.org/draft-05/schema#",
```
geigerj0: I forgot it: Why are we using such a 10 yearsl old schema version (`draft-05`) and not a more recent one?

api/policyvalidator/meta.schema.json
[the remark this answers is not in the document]
geigerj0: I would go with `oneOf` just to avoid any indirections and to make it more expressive that only one schema is allowed to match. 

We already did a little change in the legacy schema to clearly identify one:

```
    "schema-version": {
      "type": "null"
    },
```

I think this allows us to go with `oneOf`.

If there will be newer schema version will either remove the old version, in case it is a breaking change, or simply add more to ensure backwards compatibility.

api/policyvalidator/meta.schema.json
geigerj0: cc does not support it

api/broker/binding_request_parser/parser.go
```diff
@@ -0,0 +1,183 @@
...
+
+func (brp *bindRequestParser) Parse(details domain.BindDetails) (models.AppScalingConfig, error) {
+	var scalingPolicyRaw json.RawMessage
+	if details.RawParameters != nil {
+		scalingPolicyRaw = details.RawParameters
+	}
+
+	var err error
+
+	// This just gets used for legacy-reasons. The actually parsing happens in the step
```
geigerj0: > `This just gets used for legacy-reasons`

What are these "legacy-reasons"? Shall we improve the comment here a bit?

models/policy.go
```diff
@@ -170,7 +171,10 @@ func (sp ScalingPolicy) ToRawJSON() (json.RawMessage, error) {
 
 func ScalingPolicyFromRawJSON(data json.RawMessage) (*ScalingPolicy, error) {
 	// If the data is nil, we return a default ScalingPolicy with default configuration.
-	if len(data) <= 0 {
+	if noPolicyData := len(data) <= 0 ||
```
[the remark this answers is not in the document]
joergdw: This should be part of a multi-parser-refactoring. That "*FromJson*" should move into parsers.

🚧

api/policyvalidator/shared_definitions.json
```diff
@@ -5,6 +5,12 @@
       "description": "Unique identificator for a CF-resource, e.g. app, space, service-binding, …",
       "pattern": "(\\d|[a-f]){8}-(\\d|[a-f]){4}-(\\d|[a-f]){4}-(\\d|[a-f]){4}-(\\d|[a-f]){12}",
       "example": "8d0cee08-23ad-4813-a779-ad8118ea0b91"
+    },
+    "app-scaling-config-schema": {
+      "description": "Version-identifier for the used schema of this parameters-object.",
+      "type": "string",
+      "pattern": "^[0-9]+(\\.[0-9]+)?$",
+      "enum": ["0.9"]
```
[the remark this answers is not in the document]
joergdw: We agreed on `0.1`.

acceptance/helpers/helpers.go
```diff
@@ -540,6 +541,32 @@
...
 }
 
+func CreateServiceKeyWithParams(serviceInstanceName, serviceKeyName string, params string, timeout time.Duration) *Session {
+	session := cf.Cf("create-service-key", serviceInstanceName, serviceKeyName, "-c", params).
+		Wait(timeout)
+
+	return session
+}
+
+func CreateServiceInOtherSpace(
```
geigerj0: do we still need `CreateServiceInOtherSpace`? It seems to be unused

acceptance/app/dynamic_policy_test.go
```diff
@@ -25,316 +27,387 @@ var _ = Describe("AutoScaler dynamic policy", func() {
...
+				Expect(session).To(Exit(0))
 
-	Context("when there is a scaling policy for disk", func() {
-		BeforeEach(func() {
-			policy = GenerateDynamicScaleOutAndInPolicy(1, 2, "disk", 300, 600)
-			initialInstanceCount = 1
-		})
+				// Part-validation setup
+				By("Starting disk usage to trigger scale out")
+				helpers.StartDiskUsage(cfg, appToScaleName, 800, 5)
```
geigerj0: Maybe I am missing something but I don't know how the scaling has happened when the scaling policy is tailored for `memoryused` https://github.com/cloudfoundry/app-autoscaler/blob/55fb6617b9ea0cdb2f501dcc05025857d7e5c39e/acceptance/app/dynamic_policy_test.go#L357

acceptance/app/dynamic_policy_test.go
```diff
@@ -25,316 +27,387 @@ var _ = Describe("AutoScaler dynamic policy", func() {
...
+			AfterEach(func() {
+				helpers.DeleteServiceInstance(cfg, serviceInstanceName)
+			})
 
-			StartDiskUsage(cfg, appToScaleName, 800, 5)
-			WaitForNInstancesRunning(appToScaleGUID, 2, 5*time.Minute)
+			It("succeeds and scales both, up and down", func() {
+				// Execution
+				serviceKeyName := fmt.Sprintf("aas-key_for%s", appToScaleName)
+				session = helpers.CreateServiceKeyWithParams(
```
geigerj0: (unrelated to this code line)

I should have asked this earlier but the question just popped up in my head.

What happens inside the autoscaler service, whenever someone did not clean up the service key after the application has been deleted. in other words: what happens inside the autoscaler service if the application guid does no longer exist?

If we don't clean up these "orphaned" scaling policies, can we run into issues? Do we need to bring orphan deletion in a follow-up PR or so?

PR #845, opened 2025-11-07
 Allow customization of log level in apply-changelog.sh 

bonzofenix:
Set default LOG_LEVEL to INFO in mta.tpl.yaml for apply-changelog.sh

### DIFF
--- a/dbtasks/pom.xml
+++ b/dbtasks/pom.xml
@@ -30,6 +30,11 @@
       <version>4.33.0</version>
     </dependency>
     <dependency>
+      <groupId>info.picocli</groupId>
+      <artifactId>picocli</artifactId>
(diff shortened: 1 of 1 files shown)


silvestre (approved):

dbtasks/pom.xml
```diff
@@ -29,6 +29,11 @@
       <artifactId>liquibase-core</artifactId>
       <version>4.33.0</version>
     </dependency>
+    <dependency>
```
asalan316: Do we need this  info.picocli cli ? I think, the dependency is need to run GraalVM-enabled command line apps.

asalan316: Seems like that there was classnotfound issue.

PR #861, opened 2025-12-08
JUnit Jupiter migration from JUnit 4.x

silvestre:
# Issue

To be able to update to Spring 4.x we should be on Junit 5.

# Fix

1. Used Moderne[1] to create an initial update PR.
2. Fixed remaining issues with CoPilot

# AI Disclosure

Prompted GitHub CoPilot to replace the Junit4 TestRule with a Junit5 parametrized test.

[1]: Use this link to re-run the recipe: https://app.moderne.io/recipes/org.openrewrite.java.testing.junit5.JUnit4to5Migration?organizationId=MGM2YTRlNjktMmQ4YS00ZTAwLTk2YTUtMDdlZmFhYThkYTIx

Co-authored-by: Moderne <team@moderne.io>


scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/misc/ZonedDateTimeTest.java
joergdw: This probably needs to be in the same order as the content of `DEFAULT_ZONE_IDS`. Question: How does this work together with the `TimeZoneExtension` Looks like it needs to get somehow interleaved with these lines here? Or is it rather that we don't need the `@ExtendWith`-Annotation above?

joergdw (commented):
It looks mostly fine. However I have one open question mainly due to the nature of how this test-framework now works.

joergdw (approved):

PR #863, opened 2025-12-08
↑ Update golang to version `v1.24.5`;

joergdw:


silvestre:
Done in https://github.com/cloudfoundry/app-autoscaler/pull/1051

PR #867, opened 2025-12-10
Publish json-schema

joergdw:
This change fixes the missing binding-request-schema. It gets exported in (only) its most recent version.

This PR is complemented by <https://github.com/cloudfoundry/app-autoscaler-release/pull/3995>. Both need each other.


bonzofenix (approved):
lgtm , we might want to add a python lint later considering we introduce some python

PR #875, opened 2025-12-23
Add cleanup workflow and script for Autoscaler deployments

bonzofenix:
• Create a new GitHub Actions workflow to run cleanup tasks for Autoscaler deployments every Monday at 05:30 UTC.
 • Add a new bash script cleanup-autoscaler-deployments.sh to handle the cleanup process, including logging in to CF,
   identifying deployments, and purging resources.
 • Update common.sh to include purging orphaned service instances across all spaces and unset additional environment
   variables.

### DIFF
--- a/scripts/common.sh
+++ b/scripts/common.sh
@@ -128,15 +128,15 @@ function cleanup_apps(){
 	# Purge orphaned service instances from all spaces in the org
 	echo "- Purging orphaned service instances from all spaces"
 	set +e
-	cf spaces 2>/dev/null | tail -n +4 | awk '{print $1}' | while read -r space_name; do
+	cf spaces 2>/dev/null | tail --lines +4 | awk '{print $1}' | while read -r space_name; do
 		if [ -n "${space_name}" ] && [ "${space_name}" != "name" ]; then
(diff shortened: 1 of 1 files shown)


scripts/cleanup-autoscaler-deployments.sh
```diff
@@ -0,0 +1,32 @@
...
+# shellcheck source=scripts/common.sh
+source "${script_dir}/common.sh"
+
+function get_autoscaler_deployments(){
+	cf curl /v3/organizations | jq -r '.resources[] | select(.name | contains("autoscaler")) | .name'
+}
+
+function main(){
+	cf_login
+	step "Deployments to cleanup: $(get_autoscaler_deployments)"
```
geigerj0: shall we store the result of `get_autoscaler_deployments` somewhere? currently, the thing is being called twice. here, just for logging and then again for the loop https://github.com/cloudfoundry/app-autoscaler/blob/740120a8c2f060ec0d2185d33086507661d45247/scripts/cleanup-autoscaler-deployments.sh#L28

bonzofenix: fixed in 0b70afe3

PR #877, opened 2025-12-23
chore: fix java linting

silvestre:
# Issue

- There was a Java file not linted according to Google Java Format
- For some reason the GitHub Actions workflow for linting the Java code was downloading the Google Java Formatter, even though it is already part of the devbox

# Fix

- Introduce a target `check-style` to run `checkstyle`
- Call the scheduler Makefile from GHA workflow
- Reformat the offending file


bonzofenix (approved):
lgtm

PR #964, opened 2026-02-05
refactor: replace custom CF client with go-cfclient/v3

bonzofenix:
## Summary
- Replace custom CF API client implementation with the official `go-cfclient/v3` library
- Maintain existing interfaces for backward compatibility using wrapper/adapter pattern

## Changes
- Add `cf/cfclient_wrapper.go` implementing CFClient interfaces using go-cfclient/v3
- Add `cf/types.go` with type mappings between go-cfclient resources and existing types
- Add `cf/errors.go` with error handling and mapping for go-cfclient errors
- Remove 16 files (~3,200 lines) of custom implementation:
  - `app.go`, `app_processes.go`, `roles.go`, `oauth.go`
  - `service_instances.go`, `service_plans.go`, `endpoints.go`
  - `retriever.go`, `lazy.go`, `drain_transport.go`
  - Associated test files
- Update `cf/mocks/mocks.go` for go-cfclient v3's process GUID-based scaling API

### Key Implementation Details
- UAA `/introspect` uses direct HTTP with Basic Auth (client credentials)
- UAA `/userinfo` uses direct HTTP with Bearer token
- go-cfclient v3 scales by process GUID, not app GUID (requires fetching process first)
- Thread-safe lazy initialization for endpoints caching

## Testing
- [x] Unit tests pass (27 specs)
- [x] Integration tests pass
- [x] Acceptance tests pass (API suite)
- [x] Linters pass


cf/cfclient_wrapper.go
```diff
@@ -0,0 +1,523 @@
...
+	"sync"
+	"time"
+
+	"code.cloudfoundry.org/lager/v3"
+	"github.com/cloudfoundry/go-cfclient/v3/client"
+	"github.com/cloudfoundry/go-cfclient/v3/config"
+	"github.com/cloudfoundry/go-cfclient/v3/resource"
+)
+
+const uaaRequestTimeout = 30 * time.Second
```
bonzofenix: was this value set the same in the original implementation?

bonzofenix:
## Added: HTTP Client Configuration for go-cfclient

After reviewing the `cf.ClientConfig` usage, I noticed that the go-cfclient migration wasn't applying the retry and connection pool settings that were previously used. 

### What was missing
The previous implementation used `ClientConfig` to configure:
- `MaxRetries` / `MaxRetryWaitMs` - for automatic retry on transient failures
- `IdleConnectionTimeoutMs` / `MaxIdleConnsPerHost` - for connection pooling

The initial go-cfclient implementation wasn't using these settings, which could have resulted in:
1. No automatic retries on transient network failures
2. Default Go HTTP transport connection pooling (instead of the configured values)

### Fix (commit e0a370ce)
Now we create a properly configured HTTP client and pass it to go-cfclient via `config.HttpClient()`. This ensures:
- The same retry logic (via `retryablehttp`) is applied to CF API calls
- Connection pool settings match what was previously configured
- TLS validation settings are properly applied

The custom HTTP client option (`WithHTTPClient`) is still available for testing scenarios.

bonzofenix:
### Additional fix: User-Agent header on UAA requests

The previous implementation called `setUserAgent(req)` on every request including UAA requests (introspect, userinfo). After the migration, only go-cfclient requests had the User-Agent header set via `config.UserAgent()`.

**Fix:** Added `User-Agent` header to all UAA requests in `doUaaRequest()`.

Commit: 251dc0ee

cf/cfclient_wrapper.go
```diff
@@ -0,0 +1,479 @@
...
+	body := "token=" + url.QueryEscape(token)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, uaaURL+"/introspect", strings.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("failed to create introspect request: %w", err)
+	}
+
+	credentials := base64.StdEncoding.EncodeToString([]byte(w.conf.ClientID + ":" + w.conf.Secret))
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Authorization", "Basic "+credentials)
+
```
silvestre: This is the deprecated way of authenticating according to https://docs.cloudfoundry.org/api/uaa/version/78.10.0/index.html#introspect-token and the code currently on `main` is already using the recommended way. We should not get worse.

Can't we just use `ExecuteAuthRequest()`

cf/client.go
```diff
@@ -98,266 +35,20 @@ type (
...
-	retryClient.RetryMax = 0
-	if config.MaxRetries != 0 {
-		retryClient.RetryMax = config.MaxRetries
-	}
-	if config.MaxRetryWaitMs != 0 {
-		retryClient.RetryWaitMax = time.Duration(config.MaxRetryWaitMs) * time.Millisecond
-	}
+	retryClient.RetryMax = config.MaxRetries
+	retryClient.RetryWaitMax = time.Duration(config.MaxRetryWaitMs) * time.Millisecond
 	retryClient.Logger = LeveledLoggerAdapter{logger.Session("retryablehttp")}
```
silvestre, line 50: This overwrites the defaults of the `retryClient` if we haven't explicitly configured these values:The previous implementation was careful about this. 

This code would by default not retry any request while the previous implementation had sensible defaults set.

I think we should add a test case for that.

silvestre (changes requested):

cf/mocks/mocks.go
```diff
@@ -121,7 +121,8 @@ func (a AddMock) Info(url string) AddMock {
 }
 
 func (a AddMock) ScaleAppWebProcess() AddMock {
-	a.server.RouteToHandler("POST", regexp.MustCompile(`^/v3/apps/[^/]+/processes/web/actions/scale$`), ghttp.RespondWith(http.StatusAccepted, "{}"))
+	// go-cfclient v3 scales by process GUID, not app GUID
+	a.server.RouteToHandler("POST", regexp.MustCompile(`^/v3/processes/[^/]+/actions/scale$`), ghttp.RespondWith(http.StatusAccepted, "{}"))
```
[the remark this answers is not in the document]
bonzofenix: The previous implementation made direct HTTP calls to `/v3/apps/{appGUID}/processes/web/actions/scale`, bypassing go-cfclient for scaling. The new code uses `cfClient.Processes.Scale(processGUID, ...)` which internally calls `/v3/processes/{processGUID}/actions/scale`. Both are valid CF v3 endpoints — the mock changed to match what go-cfclient actually sends. The old mock "worked" because the old code was not routing scaling through the go-cfclient library.

cf/cfclient_wrapper.go
bonzofenix: No — the original used `cfhttp.NewClient()` which does not set `http.Client.Timeout` (defaults to 0, meaning no overall request timeout). This 30s timeout is new and stricter — it prevents UAA requests from hanging indefinitely. 30s is generous for a token introspection or userinfo call. Happy to adjust if needed.

PR #1168, opened 2026-04-23
Option to switch off basic-authentication

joergdw:
This PR allows operators of app-autoscaler to control how "Basic Authentication" for sending custom metrics to the `metricsforwarder` is handled.


api/cmd/api/api_suite_test.go
joergdw, line 1: In some tests within this file and within other files of this PR we need to circumvent that the parsing-internals of the configuration are not published anymore. Hence we use `map[string]any`-Values to build up a configuration as it can easily be serialised into a yaml-string.

api/cmd/api/api_suite_test.go
```diff
@@ -114,102 +110,19 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
-	}
-	conf.RateLimit.MaxAmount = 10
-	conf.RateLimit.ValidDuration = 1 * time.Second
-
-	conf.CredHelperImpl = "default"
-
-	configFile = writeConfig(&conf)
 
+	// Load base config from YAML file and override dynamic values
+	conf = loadBaseConfig()
```
joergdw, line 116: This base-config is in the file `base_config.yml` and contains many values that are common between different tests.

api/publicapiserver/public_api_handler.go
joergdw, line 1: Unrelated: This attribute was unused.

joergdw (commented):
Here some hints to the reviewer:
 + This is a PR which includes the use of claude-opus-4.6. All revisions with changes to the code authored by this model are marked with a `🤖` and the author-name is set accordingly to `anthropic/claude-opus-4-6`. Everything else has been written by myself.
 + The llm has focused on the tests (i.e. all files with name `*_test.go`). I have reviewed myself the changes it did. So there _should_ not be the need to review them again. (However please feel free to do so anyway, if you want.)
 + Regarding the lament of Sonarcloud due to "duplicated code" this mostly affects the tests. And it is debatable if it really improves the readability if you need to sync there even more the ginkgo-specs. (How can we switch that off?)

asalan316:
Adding criteria for better understanding the PR :)

New basic auth config is added as follows

```yaml
basic_auth_for_custom_metrics: "on" | "off" | "only_existing_bindings"
```

Acceptance Criteria:

- on → current behavior unchanged
- off → reject binding requests with credential-type: binding-secret
- only_existing_bindings → reject new binding-secret, existing continue working

api/broker/broker.go
```diff
@@ -89,15 +89,8 @@ func New(
...
-	if len(conf.DefaultCustomMetricsCredentialType) > 0 {
-		var err error
-		defaultCustomMetricsCredentialType, err = models.ParseCustomMetricsBindingAuthScheme(
-			conf.DefaultCustomMetricsCredentialType)
-		if err != nil {
-			logger.Fatal("parse-default-credential-type", err, lager.Data{
-				"default-credential-type": conf.DefaultCustomMetricsCredentialType,
-			})
-		}
+	if basicAuthAvailable := conf.CustomMetricsAuthConfig != nil; basicAuthAvailable {
```
asalan316: may be `if basicAuthAvailable {` improves readability (instead of inline assignment)

api/broker/broker.go
```diff
@@ -580,6 +576,15 @@ func (b *Broker) Bind(
 		return result, handleParsingError(err, logger)
 	}
 
+	allowBasicAuthForNewBindings := b.conf.CustomMetricsAuthConfig != nil &&
+		b.conf.CustomMetricsAuthConfig.BasicAuthHandling == config.BasicAuthHandlingOn
+	hasRequestedBasicAuth := *appScalingConfig.GetConfiguration().GetCustomMetricsBindingAuth() == models.BindingSecret
+	if !allowBasicAuthForNewBindings && hasRequestedBasicAuth {
+		err := errors.New("Requested binding with basic-authentication for custom metrics, but the server does not allow this.")
```
asalan316: 'this' in the error message is not clear

How about:

```
... custom metrics, but the server does not allow basic_auth_for_custom_metrics='only_existing_bindings'" or similar.
``

File: api/config/parser.go [hunk 4]
```diff
@@ -0,0 +1,413 @@
...
+				// service-degradation.
+				err := models.InvalidArgumentError{
+					Param: "DefaultCustomMetricsCredentialType",
+					Value: rawConfig.DefaultCustomMetricsCredentialType,
+					Msg: fmt.Errorf(
+						"An error occurred during api-service-configuration: %w\nThis is a programming error.",
+						err).Error(),
+				}
+				panic(err)
+			}
```
asalan316, line 126: Instead of throwing a panic, I would propose to return error and let the caller handle the error

```
if err != nil {
				return nil, fmt.Errorf("invalid DefaultCustomMetricsCredentialType %q: %w",
					rawConfig.DefaultCustomMetricsCredentialType, err)
			}

....
//caller

cmBasicAuthCfg, err := parseCMBasicAuthCfg(rawConfig)
	if err != nil {
		return Config{}, fmt.Errorf("failed to parse custom metrics auth config: %w", err)
	}


```

Something along the lines

api/broker/broker.go
```diff
@@ -1072,11 +1077,19 @@ func (b *Broker) deleteBinding(ctx context.Context, bindingId string, serviceIns
...
-	if err != nil {
-		logger.Error("failed to delete custom metrics credential for unbinding", err)
-		return ErrCredentialNotDeleted
+	// Actually you probably want to test this: `isBasicAuthAvailable := b.credentials != nil`
+	// However this may not work under some circumstances, where b.credentials may have a
+	// runtime-type and is therefore a typed nil, see: <https://go.dev/doc/faq#nil_error>
+	// Therefore we check it differently:
+	isBasicAuthAvailable := b.conf.CustomMetricsAuthConfig != nil
+	if isBasicAuthAvailable {
+		err = b.credentials.Delete(ctx, appId)
```
asalan316, line 1091: May be following case can occur:

If a switch from basic_auth_for_custom_metrics="on" to "off" is made ,
any previously created credentials might remain in the database as orphaned row.

If above case is legitimate, we can probaly inform the operator by logging ...e.g manual deletion of orphaned bindings are required

api/config/config.go
```diff
@@ -95,251 +88,65 @@ type LowerUpperThresholdConfig struct {
...
+type BasicAuthHandling int
 
-	conf.CatalogPath = catalogPath
+const (
+	BasicAuthHandlingOn BasicAuthHandling = iota
+	BasicAuthHandlingOnlyExistingBindings
 
-	return err
-}
+	// // The follwing gets encoded by having or not having a `CustomMetricsAuthConfig` at all.
```
asalan316, line 137: Should this comment be removed ?

api/config/parser.go
```diff
@@ -0,0 +1,413 @@
...
+		} else {
+			impl = models.BasicAuthHandlingNative{}
+		}
+		basicAuthImplCfg = &impl
+	}
+	if err := vcapReader.ConfigureDatabases(&conf.Db, basicAuthImplCfg); err != nil {
+		return err
+	}
+
+	// var storedProcConfig *models.StoredProcedureConfig
```
asalan316: comment ?

api/broker/broker.go
joergdw: Done

api/broker/broker.go
joergdw: Good point! I have proposed a change.

api/config/parser.go
joergdw: An error here is a programming error. If we just log it out, we probably will miss that our config-validation is broken. I rather would prefer to fail here as loud and noisy as possible.

I assume that this is not dangerous. The panic would occur only on startup – when you usually still have in place other instances of the apiserver.

api/broker/broker.go
joergdw: This is a legitimate case. Let's clarify how we want to handle those cases. The options are:

- Ignoring
- Notification via a warning-log that basic-auth has been switched of but there are credentials in the credential-helper. Please note that we can not do this in each case because the configuration only gets loaded once at (re)start-time. So we can not observe a change at runtime but instead need to rely on the state of the `credentials`-table and provide the hint if it exists and is not empty. In case stored-procedures have been used, there is nothing we can do. The [interface](<https://github.com/cloudfoundry/app-autoscaler/blob/basic-auth-switch/cred_helper/interface.go#L15-L21>) does not allow to check if credentials are there or not.
- Notification via release-message. This is simple but may be missed. Operators would need to implement their own clean-up as everybody needs to do that who has used stored-procedures.

api/config/config.go
joergdw: I have kept it there as an explanation why `BasicAuthHandlingOff` does not exist although you may expect that here, when your read the code. So this comment is there intentionally. However we can easily remove it. Has it helped you?

api/config/parser.go
joergdw: I have removed that.

api/broker/broker.go
joergdw: We just determined together that we will write into the release-notes that basic-authentication can now be switched off (and the other config-options). We place a hint that in case the operator has used the internal credentials-table how she/he can make it empty.

.golangci.yaml
```diff
@@ -80,6 +80,8 @@ linters:
         - -ST1001
         - -ST1003
         - -ST1005
+        - -ST1023
+        - -QF1011
```
joergdw, line 84: These changes are unrelated. This two lints do the same and forbid explicit type declarations for variables that directly get a value that is suitable for inference.

However there is no error in writing types explicitly.

api/config/parser_internal_test.go
```diff
@@ -0,0 +1,513 @@
...
+			conf.CredHelperImpl = ""
+			conf.DefaultCustomMetricsCredentialType = "binding-secret"
+		})
+
+		It("should return an error (contradictory config)", func() {
+			Expect(err).To(HaveOccurred())
+		})
+	})
+
+	When("basic_auth_for_custom_metrics is 'on'", func() {
```
asalan316, line 464: Not sure, if the following is somewhere already covered:

for backward compatibility reasons, shouldn't be another test

" if the `basic_auth_for_custom_metrics` is empty/missing, then default `on` is considered"
```
...
conf.BasicAuthForCustomMetrics = ""
conf.CredHelperImpl = "default"
conf.DefaultCustomMetricsCredentialType = "binding-secret"
...

```

asalan316 (approved):

joergdw:
This has been approved. I resolve all threads.

PR #1178, opened 2026-05-04
Avoid direnv on non-linux systems;

joergdw:
Direnv via Nix currently has issues, see <https://github.com/NixOS/nixpkgs/issues/507531>.

Because it can not be determined anymore why it was added, this PR just removes it.


devbox.json
```diff
@@ -23,7 +23,10 @@
     "credhub-cli":          "2.9.54",
     "curl":                 "8.17.0",
     "delve":                "1.26.1",
-    "direnv":               "2.37.1",
+    "direnv": {
+      "version": "2.37.1",
+      "platforms": ["x86_64-linux", "aarch64-linux"]
+    },
```
geigerj0: why not downgrading to 2.37.0?

joergdw: That's a solution as well. I don't care. The choice here was to distinguish between our development-hardware and the CI/CD-hardware (because there are no better means to differentiate).

devbox.json
```diff
@@ -23,7 +23,7 @@
     "credhub-cli":          "2.9.54",
     "curl":                 "8.17.0",
     "delve":                "1.26.1",
-    "direnv":               "2.37.1",
+    "direnv":               "2.37.0",
```
geigerj0: do we need to disable the renovate bumps for this as long as this is buggy?

example:
https://github.com/cloudfoundry/app-autoscaler/blob/3c063dc3366bd63ea38480640c0b2bda8455a22d/renovate.json#L37-L46

geigerj0 (approved):
looks good! 

maybe you can fix the PR title and description since the whole thing turned completely :)

PR #1179, opened 2026-05-05
feat: support Java Buildpack 5.x alongside 4.x

bonzofenix:
## Summary
- Pins the MTA deployment template to Java Buildpack v5.x and configures both the `dbtasks` and `scheduler` modules to use JBP v5.0.2 with a `21.+` JRE version wildcard, maintaining backward compatibility with JBP 4.x environments via the `java-buildpack` MTA parameter.
- Adds `ServletInitializer` extending `SpringBootServletInitializer` so the scheduler WAR is bootstrapped correctly under JBP 5.x, which uses the Servlet 3.0 SCI mechanism rather than the JBP 4.x spring-auto-reconfiguration approach.
- Fixes a previously latent test isolation bug exposed by `ServletInitializer`: `EmbeddedTomcatUtil.addWebapp()` was triggering Servlet 3.0 scanning that bootstrapped a rogue Spring application context with a clustered Quartz scheduler competing for JDBC triggers, causing `ScheduleRestControllerCreateScheduleAndNofifyScalingEngineTest` to time out on every CI run.

## Changes
- **`mta.tpl.yaml`**: Add `java-buildpack: java_buildpack_v5` MTA parameter; set `buildpack: ${java-buildpack}` on `dbtasks` and `scheduler` modules; simplify `JBP_CONFIG_OPEN_JDK_JRE` to `{ jre: { version: 21.+ } }` for both JBP generations.
- **`scripts/build-extension-file.sh`**: Propagate `JBP_CONFIG_OPEN_JDK_JRE` to the extension file for the `scheduler` module.
- **`dbtasks/src/main/resources/bin/apply-changelog.sh`**: Resolve `JAVA_BIN` from `JAVA_HOME` first so it works under JBP 5.x's changed JRE layout; rename shadowed `USER` variable to `DB_USER`.
- **`scheduler/src/main/java/.../ServletInitializer.java`**: New class — registers the scheduler as a Servlet 3.0 web application for JBP 5.x WAR deployment.
- **`scheduler/src/main/java/.../health/HealthExporter.java`**: Catch duplicate `CollectorRegistry` registration on Spring context refresh (triggered by `@DirtiesContext`) and re-register cleanly.
- **`scheduler/src/test/resources/application.yml`**: New file — overrides `isClustered: false` and `scheduler.healthserver.port: 0` for all test JVMs, preventing Quartz cluster-lock delays and Prometheus port conflicts.
- **`scheduler/src/test/java/.../util/EmbeddedTomcatUtil.java`**: Switch from `addWebapp()` to `addContext()` to prevent Servlet 3.0 scanning from bootstrapping a second Spring application on port 8091; use static `Tomcat.addServlet(Context, ...)` API.
- **`scheduler/src/test/java/.../util/TestJobListener.java`**: Replace `synchronized/wait/notify` with `CountDownLatch` for reliable job-completion signalling.
- **`scheduler/src/test/java/.../health/Scheduler*EndpointTest.java`**: Inject `HTTPServer` bean and call `getPort()` for the actual bound port (required when `port: 0`).

## Testing
- [x] Full scheduler test suite passes locally: 172 tests, 0 failures (PostgreSQL 16 via Docker)
- [x] `ScheduleRestControllerCreateScheduleAndNofifyScalingEngineTest` — the previously always-failing test — passes reliably
- [ ] CI validation on PostgreSQL 15 and 16 pending


bonzofenix:
## Note on `TestJobListener` rewrite (commit 8e7227aa)

The original implementation used `synchronized`/`wait()`/`notify()` which has a classic race condition: if the Quartz job fires *before* the test thread enters `wait()`, the `notify()` signal is lost and the test hangs until timeout — causing flaky CI failures.

Replaced with `CountDownLatch` which is the idiomatic Java solution for "wait until N events happen." A latch never loses signals — if the job fires before `await()` is called, the count is already zero and `await()` returns immediately.

Other cleanup in the same commit:
- Removed redundant `volatile` on `AtomicInteger` (already internally thread-safe)
- Made fields `final` (immutable after construction)
- Removed `synchronized` blocks (unnecessary with latch coordination)

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerCreateScheduleAndNofifyScalingEngineTest.java
```diff
@@ -47,7 +47,9 @@
 import org.springframework.transaction.annotation.Transactional;
 import org.springframework.web.context.WebApplicationContext;
 
-@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT)
+@SpringBootTest(
+    webEnvironment = WebEnvironment.RANDOM_PORT,
+    properties = "spring.quartz.properties.org.quartz.jobStore.isClustered=false")
```
bonzofenix: can you explain why we did this ?

bonzofenix:
## Note on `@DirtiesContext` removal and `HealthExporter` change

### What changed

1. **Removed `@DirtiesContext(BEFORE_CLASS)`** from `ScheduleRestControllerCreateScheduleAndNofifyScalingEngineTest`
2. **Made `HealthExporter.init()` idempotent** — catches `IllegalArgumentException` on duplicate collector registration, clears the registry, and re-registers.

### Why

The Prometheus `CollectorRegistry.defaultRegistry` is a JVM-global static singleton that survives Spring context destruction. On `main`, all tests share the same Spring context cache key, so the collector is only registered once. 

Our test adds `properties = "spring.quartz.properties.org.quartz.jobStore.isClustered=false"`, which gives it a **unique context cache key**. Combined with `@DirtiesContext(BEFORE_CLASS)`, this caused:

1. `@DirtiesContext` destroys the previously-cached default context
2. Our test creates a new context (unique key) → `register()` succeeds
3. Next test class needing the default context recreates it → `register()` fails with "already in use"
4. Cascading `Failed to load ApplicationContext` across all subsequent test classes

### Why this is safe

- `isClustered=false` already isolates our test in its own Spring context — `@DirtiesContext` was unnecessary
- The `HealthExporter` defensive registration only triggers during test context recreation; in production there's always exactly one context

bonzofenix:
### Difference from `main`

On `main`, this test has `@DirtiesContext(BEFORE_CLASS)` and **no** custom properties — so it shares the same context cache key as other tests. The destroy/recreate cycle works fine because the recreated context uses the same key and gets cached for reuse. The collector is only registered once per JVM.

On this branch, we added `isClustered=false` as a property override, which changes the context cache key. That made `@DirtiesContext` destructive: it nukes the shared context, and when other tests try to rebuild it, they hit the already-registered collector in the global registry.

Removing `@DirtiesContext` is the correct fix because `isClustered=false` already provides isolation (separate context) without needing to destroy the shared one.

dbtasks/src/main/resources/bin/apply-changelog.sh
```diff
@@ -3,12 +3,13 @@
 
 set -euo pipefail
 
+JAVA_BIN=${JAVA_BIN:-"${JAVA_HOME:+$JAVA_HOME/bin/java}"}
 JAVA_BIN=${JAVA_BIN:-"/home/vcap/app/.java-buildpack/open_jdk_jre/bin/java"}
```
joergdw: Why do we have two times a definition of `JAVA_BIN`?.

dbtasks/src/main/resources/bin/apply-changelog.sh
joergdw, line 1: Although it works here as written, could we please use `${}`-syntax? It is easier to spot where the name of the variable finishes. (And in some cases it avoids errors.)

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
joergdw, line 19: Just for my understanding. Why do we have this setter instead of just setting it in the constructor? This function only gets called in [HealthConfig.java:26](<https://github.com/cloudfoundry/app-autoscaler/blob/APPAUTOSCALER-1047-java-buildpack-5x/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/HealthConfig.java#L26>)

And this if and only if we anyway create a fresh `HealthExporter`.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
```diff
@@ -12,6 +20,11 @@ public void setDbStatusCollector(DbStatusCollector dbStatusCollector) {
 
   public void init() {
     DefaultExports.initialize();
+    Collector previous = registeredCollector.getAndSet(dbStatusCollector);
+    if (previous != null) {
+      logger.debug("Re-registering DbStatusCollector after context refresh");
```
joergdw, line 25: This is structurally always false, see comment above.

joergdw (commented):
This looks good to me, however I have a few questions.

scheduler/src/test/java/org/cloudfoundry/autoscaler/scheduler/rest/ScheduleRestControllerCreateScheduleAndNofifyScalingEngineTest.java
bonzofenix: The `isClustered=false` property was added to prevent Quartz from attempting cluster coordination during tests with `@DirtiesContext`, which caused a "Scheduler already registered" error on context reload. That approach was later superseded — the re-registration guard in `HealthExporter` (using a static `AtomicReference`) handles the duplicate-registration problem directly, so the property is no longer needed and has since been removed from the test annotation.

dbtasks/src/main/resources/bin/apply-changelog.sh
bonzofenix: Good catch. The intent was: prefer `JAVA_HOME` if set, else fall back to the buildpack hardcoded path — but expressing it as two reassignments of the same variable was confusing. Fixed in ea88aed: now uses an intermediate `JAVA_BIN_FROM_HOME` variable so the three-level priority (`JAVA_BIN` > `JAVA_HOME` > buildpack default) is explicit in a single assignment.

dbtasks/src/main/resources/bin/apply-changelog.sh
bonzofenix: Done — applied `${}` quoting throughout the script in the same commit (ea88aed).

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
bonzofenix: The setter exists because Spring's `@Bean(initMethod = "init")` pattern wires the bean in two steps: construct, then inject dependencies, then call the init method. Using a constructor here would require either a constructor-arg bean definition or `@Autowired` on the constructor — both work, but the setter approach matches the existing Spring XML/JavaConfig style in this codebase. Happy to convert to constructor injection if you prefer — it would let us make `dbStatusCollector` `final`.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
bonzofenix: This is actually intentionally reachable. `registeredCollector` is `static`, so it survives Spring context reloads. Tests annotated with `@DirtiesContext` force a full context refresh — Spring creates a new `HealthExporter` bean instance and calls `init()` again, but the static field still holds the `Collector` registered by the previous context. Without the unregister, Prometheus throws a `IllegalArgumentException: Collector already registered` on the second `register()` call. The guard is what prevents that failure — it's not dead code.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
joergdw: Thank you for your explanation. So the tooling (LSP) does not catch up here with the injector-technique.

I personally prefer the constructor-approach because it is a bit more explicit and makes the attribute `final` but it is neither required for this PR nor does it help that much in the sense that LSPs then understand the code better.

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/health/HealthExporter.java
joergdw: This looks like pollution to me. We have here code that is there only for the tests to make them work. But this code should focus rather on the production-environment (unless that makes testing too difficult). And there it is always false.

Unfortunately I don't know how to do it better. Can we at least add a comment?
```suggestion
    boolean withRefreshedContextInTestSuite = previous != null;
    if (withRefreshedContextInTestSuite) {
      logger.debug("Re-registering DbStatusCollector after context refresh");
```

joergdw (approved):

PR #1184, opened 2026-05-13
fix: adjust acceptance test thresholds for cgroup v2 (Noble stemcell)

bonzofenix:
## Summary

Fix acceptance test failures caused by the cf-deployment v55→v56 upgrade (Jammy→Noble stemcell, cgroup v1→v2 migration).

### Changes
- Increase `WaitForNInstancesRunning` timeout from 5min to 8min (extracted as `scaleTimeout` constant)
- Increase `StartDiskUsage` HTTP curl timeout from 10s to 60s
- Reduce disk write from 800MB to 550MB (prevents container crash on Noble)
- Lower disk thresholds from 600→500 MiB and diskutil from 60→45% to account for decimal→binary conversion
- Adjust memoryused thresholds to explicit values (25/55 MiB) based on actual cgroup v2 reporting
- Adjust memoryutil thresholds from 58/63 to 30/50% to match actual ~56% utilization on cgroup v2
- Increase `minimalMemoryUsage` from 17→35 to account for kernel memory in container limit
- Extract all magic numbers into named constants with explanatory comments
- Reuse http.Client across ticker calls to avoid leaking TCP connections

## Root Cause

**cf-deployment v56 migrated from Jammy (cgroup v1) to Noble stemcell (cgroup v2).** Four issues:

### 1. Memory tests OOM (memoryused, memoryutil)

cgroup v2 includes kernel memory (pagetables, slab, stack) in the container's `memory.max` limit. cgroup v1 tracked kernel memory separately via `memory.kmem.limit_in_bytes`.

The test allocated `NodeMemoryLimit - 17MB = 111MB` in a 128MB container. On cgroup v2, the total exceeds 128MB → OOM kill → container restarts with low memory → scale-out threshold never reached.

**Fix:** Increase `minimalMemoryUsage` from 17→35 and cap heap allocation at 80MB.

### 2. Memory metric values differ on cgroup v2

On cgroup v2, 80 decimal MB heap → ~72 MiB reported (not 80). This is because:
- Go allocates `80 × 1,000,000 = 76.3 MiB` (decimal→binary)
- cgroup v2 reports Anon memory which closely matches RSS

The original formula `0.9 * heapToUse` produced a threshold of 72 — right at the boundary. And `memoryutil` was 56% actual vs 63% threshold.

**Fix:** Use explicit thresholds based on observed cgroup v2 values:
- memoryused: scale-in=25, scale-out=55 (actual value ~72 MiB)
- memoryutil: scale-in=30, scale-out=50 (actual value ~56%)

### 3. Disk tests: decimal MB vs binary MiB conversion

The test app writes `N × 1,000,000` bytes (decimal MB) but CF reports disk in binary MiB (`bytes ÷ 1,048,576`). This ~4.9% gap meant thresholds were never reached:

- Write 550 "MB" = 550,000,000 bytes → CF reports 524 MiB
- Old threshold 600 MiB → 524 < 600 → scale-out never triggered
- Old diskutil 60% → 524/1024 = 51% → never triggered

**Fix:** Lower thresholds to account for conversion: disk 600→500, diskutil 60→45%.

See #1185 for a follow-up to fix the test app to write binary MiB instead.

### 4. Disk tests crash at 800MB

Writing 800MB via `crypto/rand` on a 1GB disk quota crashed containers on Noble (higher OS overhead).

**Fix:** Reduce disk write to 550MB.

### 5. General metric latency

Guardian's [`calculateTotalUsage`](https://github.com/cloudfoundry/guardian/blob/main/rundmc/runrunc/stats.go#L104) can return 0 while `InactiveFile > File + Anon + SwapCached` during cgroup warmup. This dilutes averages during the first evaluation window but is a non-issue since memory is allocated after app startup.

**Fix:** Increase wait timeouts from 5→8 minutes.

## Verification

Confirmed via live debugging on Noble stemcell:
- `cf ssh` into container: `memory.stat` shows `anon`, `file`, `inactive_file` fields (cgroup v2)
- Perl 100MB allocation test: log-cache correctly reports ~115MB via guardian formula
- memoryutil metric dump shows 56% consistently (matching 72/128 MiB)
- CI app suite passes with corrected thresholds

## Test plan

- [x] memoryused test passes
- [x] memoryutil test passes
- [x] disk test passes
- [x] diskutil test passes
- [x] service-key test passes
- [x] responsetime/throughput/cpu tests unaffected
- [x] All 22 CI checks green

### DIFF
--- a/acceptance/app/dynamic_policy_test.go
+++ b/acceptance/app/dynamic_policy_test.go
@@ -26,9 +26,44 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 		maxHeapLimitMb int
 	)
 
-	// cgroup v2 (Noble stemcell) counts kernel memory against the container limit,
-	// unlike cgroup v1 which tracked it separately. This requires more headroom.
-	const minimalMemoryUsage = 35
+	const (
+		// cgroup v2 (Noble stemcell) counts kernel memory against the container limit,
+		// unlike cgroup v1 which tracked it separately. This requires more headroom.
+		minimalMemoryUsage = 35
+
+		// Maximum heap the test app should allocate (MB). Must leave room for
+		// app baseline + kernel overhead within the container memory limit.
+		maxSafeHeapAllocationMb = 80
+
+		// Thresholds are set below actual values to avoid flaking on metric jitter.
+		thresholdSafetyFactor = 0.9
+
+		// How long the test app holds resource usage before releasing (minutes).
+		holdMinutes = 5
+
+		// responsetime test: app sleeps 100ms per request, threshold at 50ms triggers scale-out.
+		responseTimeSlowDelayMs   = 100
+		responseTimeScaleOutMs    = 50
+		responseTimeScaleInLowMs  = 50
+		responseTimeScaleInHighMs = 150
+
+		// throughput test: generate 20 rps, scale-out at 15 rps per instance.
+		throughputRps              = 20
+		throughputScaleOutPerInstance = 15
+		throughputScaleInLow       = 5
+		throughputScaleInHigh      = 15
+
+		// disk tests: write 550MB on a 1GB disk quota.
+		diskUsageMb      = 550
+		diskScaleInMb    = 300
+		diskScaleOutMb   = 600
+		diskUtilScaleIn  = 30
+		diskUtilScaleOut = 60
+
+		// memoryutil thresholds (percentage of memory quota).
+		memoryUtilScaleIn  = 50
+		memoryUtilScaleOut = 65
+	)
 
 	When("an ordinary service-binding is used", func() {
 		JustBeforeEach(func() {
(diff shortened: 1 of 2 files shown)


acceptance/app/dynamic_policy_test.go
```diff
@@ -53,49 +90,50 @@ var _ = Describe("AutoScaler dynamic policy", func() {
...
 				})
 
 				It("should scale out and then back in.", func() {
 					By(fmt.Sprintf("Use heap %d MB of heap on app", int64(heapToUse)))
-					helpers.CurlAppInstance(cfg, appToScaleName, 0, fmt.Sprintf("/memory/%d/5", int64(heapToUse)))
+					helpers.CurlAppInstance(cfg, appToScaleName, 0, fmt.Sprintf("/memory/%d/%d", int64(heapToUse), holdMinutes))
 
 					By("wait for scale to 2")
-					helpers.WaitForNInstancesRunning(appToScaleGUID, 2, 5*time.Minute)
+					helpers.WaitForNInstancesRunning(appToScaleGUID, 2, 8*time.Minute)
```
joergdw, line 110: What about reducing a bit the magic numbers? Here as well as subsequently?
```suggestion
					reactionTimeBuffer := 3
					helpers.WaitForNInstancesRunning(appToScaleGUID, 2, (holdMinutes+reactionTimeBuffer)*time.Minute)
```

(Btw., did we really need 3 minutes reactionTimeBuffer?)

acceptance/app/dynamic_policy_test.go
```diff
@@ -26,7 +26,44 @@ var _ = Describe("AutoScaler dynamic policy", func() {
...
 
-	const minimalMemoryUsage = 17 // observed minimal memory usage by the test app
+	const (
+		// cgroup v2 (Noble stemcell) counts kernel memory against the container limit,
+		// unlike cgroup v1 which tracked it separately. This requires more headroom.
+		minimalMemoryUsage = 35
+
+		// Maximum heap the test app should allocate (MB). Must leave room for
+		// app baseline + kernel overhead within the container memory limit.
+		maxSafeHeapAllocationMb = 80
```
joergdw, line 36: Let's spend some comment how we get to this particular number.

My personal favourite: Lets calculate it. E.g. `(config.NodeMemoryLimit - minimalMemoryUsage) * oomSafetyFactor` where the latter one could be `0.9`.

E.g.: `(128MB - 35MB) * 0.9 = 83.7MB`

acceptance/app/dynamic_policy_test.go
```diff
@@ -117,12 +155,12 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 			Context("when responsetime is greater than scaling out threshold", func() {
 
 				BeforeEach(func() {
-					policy = helpers.GenerateDynamicScaleOutPolicy(1, 2, "responsetime", 50)
+					policy = helpers.GenerateDynamicScaleOutPolicy(1, 2, "responsetime", responseTimeScaleOutMs)
```
joergdw, line 162: I like how this PR already reduces the magic numbers.

joergdw (approved):
This looks fine to me. I have found one small room for improvement and another one to consider.

bonzofenix:
.

PR #1197, opened 2026-05-19
fix: make memoryutil/memoryused thresholds dynamic based on NodeMemoryLimit

bonzofenix:
## Summary

PR #1184 fixed acceptance test failures caused by the cgroup v1→v2 migration (Noble stemcell). However it introduced a regression for environments that run the tests with larger container sizes (e.g. 512MB):

- Hardcoded `maxSafeHeapAllocationMb = 80` caps heap regardless of container size
- Hardcoded `memoryUtilScaleOut = 50%` was calibrated for 80MB on a 128MB container (~56% util)
- On a 512MB container: 80MB heap = ~16% utilisation → never reaches 50% threshold → test times out

**Root cause in numbers:**
| Container | Heap | Actual util | Threshold (post-1184) | Result |
|-----------|------|-------------|----------------------|--------|
| 128MB     | 80MB | ~56%        | 50%                  | ✓ pass |
| 512MB     | 80MB | ~16%        | 50%                  | ✗ timeout |

## How This Fix Works

Instead of hardcoded constants, both the heap allocation and scale-out threshold are computed at runtime from `cfg.NodeMemoryLimit`:

**Heap allocation:**
```go
// Allocate 80% of the container limit, leaving room for cgroup v2 kernel overhead
maxHeapLimitMb = int(float64(cfg.NodeMemoryLimit) * heapFractionOfLimit) - minimalMemoryUsage
```

**memoryutil scale-out threshold:**
```go
// cgroup v2 reports binary MiB (Anon memory); Go allocates decimal MB.
// Actual reported util ≈ (heapMB × 1e6/1024²) / NodeMemoryLimit.
// Set threshold to 90% of expected util — safely below actual, above baseline.
expectedUtilPct := (float64(maxHeapLimitMb) * (1_000_000.0 / 1_048_576.0)) / float64(cfg.NodeMemoryLimit) * 100
memoryUtilScaleOut = int(expectedUtilPct * 0.90)
```

**memoryused thresholds** are similarly derived from the expected reported MiB value:
```go
reportedMiB := heapToUse * (1_000_000.0 / 1_048_576.0)
scaleInThreshold  := int64(reportedMiB/4) + 10   // between baseline and post-scale avg
scaleOutThreshold := int64(reportedMiB * 0.85)    // below single-instance value
```

**Result:**
| Container | Heap  | Expected util | Threshold (this PR) | Result |
|-----------|-------|---------------|---------------------|--------|
| 128MB     | 67MB  | ~50%          | ~44%                | ✓ pass |
| 512MB     | 374MB | ~70%          | ~62%                | ✓ pass |

## Preserved from #1184

- `minimalMemoryUsage = 35` (cgroup v2 kernel memory overhead headroom)
- `WaitForNInstancesRunning` timeout of 8 minutes
- Decimal MB → binary MiB conversion factor in threshold calculation

## Test plan

- [ ] memoryutil test passes on 128MB container (default OSS CI config)
- [ ] memoryutil test passes on 512MB container
- [ ] memoryused test passes on both container sizes
- [ ] All other dynamic policy tests unaffected

### DIFF
--- a/acceptance/app/dynamic_policy_test.go
+++ b/acceptance/app/dynamic_policy_test.go
@@ -24,7 +24,7 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 		doneAcceptChan     chan bool
 		ticker             *time.Ticker
 		maxHeapLimitMb     int
-		memoryUtilScaleOut int
+		memoryUtilScaleOut int64
 		reportedMiB        float64
 	)
 
@@ -87,7 +87,7 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 		// On cgroup v2, Go allocates decimal MB but CF reports binary MiB (Anon memory).
 		// Scale-out threshold set to memoryUtilSafetyFactor of expected util — safely below actual, above baseline.
 		expectedUtilPct := reportedMiB / float64(cfg.NodeMemoryLimit) * 100
-		memoryUtilScaleOut = int(expectedUtilPct * memoryUtilSafetyFactor)
+		memoryUtilScaleOut = int64(expectedUtilPct * memoryUtilSafetyFactor)
 	})
 	When("an ordinary service-binding is used", func() {
 		JustBeforeEach(func() {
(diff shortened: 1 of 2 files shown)


joergdw:
The compilation fails for the scheduler which has not been touched in this PR.

joergdw (approved):

PR #1206, opened 2026-05-22
perf(acceptance): speed up app suite ~15 min via droplet reuse, parallelism, and tunable timeouts

bonzofenix:
## What and why

Acceptance tests for the app suite were taking ~30 min on 4 nodes. This PR cuts that to ~15 min through three complementary changes: eliminating redundant `cf push` operations per test, doubling parallelism, and reducing over-padded wait times. The timing constants are now configurable so slow foundations can dial them up without code changes.

## Changes

**Droplet reuse — biggest win (~10 min)** — a single app droplet is built once in `BeforeSuite` and reused via `CreateTestAppFromDroplet` across all ~20 tests. Previously each test did a full `cf push` (~3 min); a droplet-based push takes ~20 s. Unique seed app names prevent parallel conflicts.

**NODES 4→8** — doubles the test parallelism in CI. Wall-clock time for the remaining work halves.

**Reduced hold and wait times** — `holdMinutes` (how long the test app sustains artificial CPU/disk/memory load) reduced 5→4: still comfortably above the `aggregate_interval (120s) + breach_duration (60s)` floor. `WaitForNInstancesRunning` timeout normalized to 8 min across all tests (was 10 min for CPU scale-in). Custom metrics polling tightened 15s→5s (direct HTTP — no aggregate window needed). `responsetime`/`throughput` scale waits halved to 4 min.

**Configurable timing via acceptance config** — `min_load_duration_minutes` and `scale_event_timeout_minutes` are now JSON config fields. Defaults are conservative (5 min / 10 min, matching original values) so existing deployments are unaffected. `build-extension-file.sh` explicitly sets the tuned values (4 / 8) for our CI environment.

**CF task GC race fixed** — `show_final` in `run-mta-acceptance-tests.sh` was re-querying the CF API after tasks completed, racing against CF's GC of `SUCCEEDED` tasks. Task state is now cached when the polling loop detects completion.

**SonarCloud duplication** — `BeforeEach` setup and `waitFor*` helpers extracted to remove flagged duplication across test files.

## Test plan

- [x] All resource-hold tests (memory, cpu, disk, throughput, responsetime) pass with 4-min hold
- [x] Custom metric scale out/in passes with 5s polling
- [x] `show_final` correctly reports results even when CF has GC'd completed tasks
- [ ] Watch app suite CI wall-clock drop from ~30 min baseline to ~15 min

### DIFF
--- a/acceptance/app/dynamic_policy_test.go
+++ b/acceptance/app/dynamic_policy_test.go
@@ -100,17 +100,7 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 				})
 
 				It("should scale out and then back in.", func() {
-					By(fmt.Sprintf("Use heap %d MB of heap on app", maxHeapLimitMb))
-					helpers.CurlAppInstance(cfg, appToScaleName, 0, fmt.Sprintf("/memory/%d/%d", maxHeapLimitMb, holdMinutes))
-
-					By("wait for scale to 2")
-					helpers.WaitForNInstancesRunning(appToScaleGUID, 2, cfg.ScaleEventTimeout())
-
-					By("Drop memory used by app")
-					helpers.CurlAppInstance(cfg, appToScaleName, 0, "/memory/close")
-
-					By("Wait for scale to minimum instances")
-					helpers.WaitForNInstancesRunning(appToScaleGUID, 1, cfg.ScaleEventTimeout())
+					waitForMemoryScaling(appToScaleName, appToScaleGUID, maxHeapLimitMb, holdMinutes)
 				})
 			})
 		})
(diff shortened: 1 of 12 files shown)


acceptance/app/dynamic_policy_test.go
```diff
@@ -65,14 +63,15 @@ var _ = Describe("AutoScaler dynamic policy", func() {
 		baselineMemoryMiB        = 10
 	)
 	BeforeEach(func() {
+		holdMinutes = cfg.MinLoadDuration()
```
joergdw: We could do this as well at once at suite-setup.

joergdw (approved):
LGTM

acceptance/app/dynamic_policy_test.go
bonzofenix: Done in e4805dd13. Moved `holdMinutes`, `maxHeapLimitMb`, `reportedMiB`, and `memoryUtilScaleOut` to package-level vars computed once in [`BeforeSuite`](https://github.com/cloudfoundry/app-autoscaler/blob/worktree-acceptance-test-speedup/acceptance/app/app_suite_test.go#L57-L83). Removed the `BeforeEach` entirely.

PR #1232, opened 2026-06-01
Test default-handling of basic-authentication

joergdw:
This PR adds a test for the default-handling of basic-authentication for custom-metrics which has been proposed in <https://github.com/cloudfoundry/app-autoscaler/pull/1168#discussion_r3330986351> on the review of #1168.

It furthermore makes the "off"-case reachable for the metricsforwarder and adds a test accordingly.

### DIFF
--- a/metricsforwarder/config/parser.go
+++ b/metricsforwarder/config/parser.go
@@ -56,6 +56,8 @@ func toConfig(rawConfig rawConfig) (Config, error) {
 		}
 	case "default", "":
 		result.CredentialHelperConfig = models.BasicAuthHandlingNative{}
+	case "disabled":
+		result.CredentialHelperConfig = nil
 	}
 
 	return result, nil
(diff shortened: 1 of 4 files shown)


metricsforwarder/config/parser.go
```diff
@@ -207,10 +212,10 @@ func loadVcapConfig(conf *rawConfig, vcapReader configutil.VCAPConfigurationRead
 
 	var basicAuthImplCfg *models.BasicAuthHandlingImplConfig
 	if conf.CredHelperImpl != "" {
```
asalan316, line 214: so, this if branch addresses `conf.CredHelperImpl != "disabled"`  case ?

asalan316: The commit 5421c872f76b91c5b9bd0c2404a1c938c9acec9f answers it :)

asalan316 (approved):

PR #1251, opened 2026-06-11
ci: replace credhub generate with local openssl rand

bonzofenix:
## What and why

Deploy pipelines previously required CredHub **write** access to generate 7 ephemeral passwords (health endpoints + service broker). These passwords are only used within a single deploy cycle — all components restart together, so there is no need to persist them in CredHub across deploys. Removing the writes makes CredHub usage read-only. 

## Changes

**Local password generation** — `generate_deployment_secrets()` now uses `openssl rand -base64 12` instead of `credhub generate --no-overwrite`. The 7 passwords are exported as shell variables and passed directly to the MTA extension file, bypassing CredHub storage entirely.

**CredHub interpolation template trimmed** — the 7 `((/bosh-autoscaler/${DEPLOYMENT_NAME}/...))` references are removed from the heredoc in `build-extension-file.sh`. The remaining references are infrastructure secrets managed by BOSH/CF (postgres creds, TLS certs, UAA client secrets, CF admin password) that must still come from CredHub.

**`load_secrets()` cleaned up** — the corresponding `export VAR=value` extractions for the generated passwords are removed since the values are now already in the environment before `load_secrets()` is called.

## Test plan

- [x] `shellcheck` passes on `build-extension-file.sh`
- [x] `generate_deployment_secrets` exports all 7 vars via `openssl rand`
- [x] `load_secrets` no longer references the removed CredHub paths
- [x] Infrastructure secrets (postgres, TLS, UAA, CF admin) still interpolated via CredHub


scripts/build-extension-file.sh
```diff
@@ -33,29 +33,25 @@ export SYSTEM_DOMAIN="autoscaler.app-runtime-interfaces.ci.cloudfoundry.org"
...
-  credhub generate --no-overwrite -n "${prefix}/autoscaler_scalingengine_health_password"    --length 16 -t password
-  credhub generate --no-overwrite -n "${prefix}/service_broker_password_blue"                --length 16 -t password
-  credhub generate --no-overwrite -n "${prefix}/service_broker_password"                     --length 16 -t password
-  return
+  METRICSFORWARDER_HEALTH_PASSWORD="$(openssl rand -base64 12)"
+  METRICSGATEWAY_HEALTH_PASSWORD="$(openssl rand -base64 12)"
+  OPERATOR_HEALTH_PASSWORD="$(openssl rand -base64 12)"
+  EVENTGENERATOR_HEALTH_PASSWORD="$(openssl rand -base64 12)"
+  SCALINGENGINE_HEALTH_PASSWORD="$(openssl rand -base64 12)"
+  SERVICE_BROKER_PASSWORD_BLUE="$(openssl rand -base64 12)"
```
asalan316, line 41: is 12 is same as  --length 16 in credhub ? Does not length matters here

asalan316 (approved):

PR #1258, opened 2026-06-15
fix(tls): switch from internal to external service defaults

silvestre:
# Issue

As discussed in the CF Slack thread [1], `code.cloudfoundry.org/tlsconfig`'s
`WithInternalServiceDefaults` does not allow ECC certificates.

The use of `WithInternalServiceDefaults` is a holdover from BOSH-era
deployments where microservices communicated directly. In the CF deployment
model, traffic flows through CF ingress (Gorouter, load balancers, reverse
proxies), so `WithExternalServiceDefaults` is the appropriate choice.

[1] https://cloudfoundry.slack.com/archives/C0RUU7GBZ/p1781197261596289?thread_ts=1781181057.598879&cid=C0RUU7GBZ

# Fix

Switch `WithInternalServiceDefaults` to `WithExternalServiceDefaults` in
`models/security.go` for both client and server TLS configs.

### DIFF
--- a/models/security.go
+++ b/models/security.go
@@ -25,13 +25,13 @@ func (t *TLSCerts) CreateClientConfig() (*tls.Config, error) {
 	}
 
 	if t.CACertFile != "" && t.CertFile == "" && t.KeyFile == "" {
-		clientTls := tlsconfig.Build(tlsconfig.WithInternalServiceDefaults())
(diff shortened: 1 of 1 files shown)


bonzofenix (commented):
lgtm

bonzofenix (approved):
lgtm

PR #1259, opened 2026-06-15
chore: pin devbox package versions for faster installs

bonzofenix:
## What and why

Six devbox packages were pinned to `latest`, forcing Nix to re-resolve versions on every `devbox install`. This adds unnecessary evaluation time both locally and in CI — especially for `postgresql` and `apple-sdk` which are large derivations.

## Changes

**Pin 6 packages from `latest` to their current resolved versions** — eliminates version re-evaluation and ensures consistent Nix binary cache hits across machines and CI runners:

| Package | Before | After |
|---------|--------|-------|
| apple-sdk | latest | 14.4 |
| coreutils | latest | 9.11 |
| gnused | latest | 4.9 |
| gnutar | latest | 1.35 |
| postgresql | latest | 17.10 |
| which | latest | 2.23 |

These are the exact versions that `latest` was already resolving to — no functional change, just deterministic pinning.

## CI context

`Install devbox` runs in every CI job. Measured on recent `main` runs:

| Workflow | Jobs | Devbox time/job |
|----------|------|-----------------|
| Acceptance Tests - MTA | 5 | 4–5 min |
| Build with Postgres | 4 | ~2 min |

~31 min of CI time per PR spent on `devbox install`. Pinning helps the devbox Nix cache (`enable-cache: true`) hit more reliably.

## Test plan

- [x] `devbox install` succeeds with pinned versions
- [x] `devbox list` shows same packages/versions as before

### DIFF
--- a/devbox.json
+++ b/devbox.json
@@ -10,7 +10,7 @@
 
     "act": "0.2.88",
     "apple-sdk": {
-      "version": "latest",
+      "version": "14.4",
       "platforms": [
         "aarch64-darwin"
       ]
@@ -19,7 +19,7 @@
     "bash-language-server": "5.6.0",
     "bosh-cli":             "7.3.1",
     "cloudfoundry-cli":     "8.18.3",
-    "coreutils":            "latest",
+    "coreutils":            "9.11",
     "credhub-cli":          "2.9.54",
     "curl":                 "8.17.0",
     "delve":                "1.26.3",
@@ -27,8 +27,8 @@
     "gh":                   "2.92.0",
     "ginkgo":               "2.28.1",
     "gnumake":              "4.4.1",
-    "gnused":               "latest",
-    "gnutar":               "latest",
+    "gnused":               "4.9",
+    "gnutar":               "1.35",
     "go-tools":             "2025.1.1",
     "golangci-lint":        "2.12.2",
     "gopls":                "0.22.0",
@@ -41,13 +41,13 @@
     "mysql":                "10.6.12",
     "oha":                  "1.14.0",
     "openssl":              "3.6.0",
-    "postgresql":           "latest",
+    "postgresql":           "17.10",
     "python":               "3.14.4",
     "redocly":              "2.17.0",
     "ripgrep":              "15.1.0",
     "shellcheck":           "0.11.0",
     "temurin-bin-21":       "21.0.3",
-    "which":                "latest",
+    "which":                "2.23",
     "xq-xml":               "1.4.0",
     "yq-go":                "4.53.2",
     "go":                   "1.25.5"
--- a/devbox.lock
+++ b/devbox.lock
@@ -97,7 +97,7 @@
         }
       }
     },
-    "apple-sdk@latest": {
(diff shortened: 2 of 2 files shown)


nix/packages/app-autoscaler-cli-plugin/default.nix
```diff
@@ -0,0 +1,49 @@
...
+    install -D -m755 $src $out/bin/app-autoscaler-cli-plugin
+  '';
+
+  meta = with lib; {
+    description = "App-AutoScaler CLI plugin for Cloud Foundry";
+    homepage = "https://github.com/cloudfoundry/app-autoscaler-cli-plugin";
+    license = licenses.asl20;
+    mainProgram = "app-autoscaler-cli-plugin";
+    platforms = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+  };
```
joergdw: Please follow [this rule](<https://github.com/NixOS/nixpkgs/blob/59c584622f03baf96e0cea39236d5c0f34a7c6a0/pkgs/README.md?plain=1#L500>) to let it be filtered out by security-aware people.

joergdw (commented):
I would prefer to follow here the usual Nix-conventions on packaging which are focussed on reproducibility and secure (more) against supply-chain-attacks (than other approaches). For speed, there is the concept of a cache called "substituter". In detail:
1. The Nix-way is to bake the precise toolchain into the package-recipe. This enhances protection against supply-chain-attacks because it is more difficult to introduce something into the Nix build pipeline. See <https://github.com/NixOS/nixpkgs/blob/59c584622f03baf96e0cea39236d5c0f34a7c6a0/pkgs/README.md?plain=1#L73-L74>. This is relevant as [history shows](<https://www.cve.org/CVERecord?id=CVE-2024-3094>). By doing so, consumers gain trust on two points: stronger assurance against supply‑chain tampering of the artefacts, and a fully reproducible build they can verify themselves.
2. The proper way to speed up package-downloads is to introduce a substituter, see <https://nix.dev/manual/nix/latest/command-ref/conf-file#conf-substituters>.

I would prefer to write those packages the ordinary way. Although it requires more setup, introducing a substituter would bring us more benefit than per-package shortcuts. This also fits better with a strategy where we have Nix package definitions for many of the software products we are responsible for.

bonzofenix:
Hey @joergdw 👋

Simplified this PR — now it only pins the 6 `latest` packages in `devbox.json` to their currently resolved versions. No nix/packages.nix changes, no fetchurl conversion.

**What changed:** `latest` → specific version for apple-sdk, coreutils, gnused, gnutar, postgresql, which. Same versions that were already resolving — just pinned explicitly so Nix can hit binary cache without re-evaluating.

**Why:** `devbox install` was taking ~24 min locally and 2–5 min per CI job partly because `latest` forces Nix to re-resolve each time. Pinning eliminates that.

Could you take another look? Should be a quick review — just version strings in devbox.json + the resulting devbox.lock update.

joergdw (approved):

PR #1149, opened 2026-04-13
feat: run PR acceptance tests with OrgManager role in shared CF org

bonzofenix:
## What and why

PR CI pipelines required admin credentials to deploy and run acceptance tests, blocking use in non-admin environments. This PR removes that dependency: PRs deploy into `SAP_autoscaler_tests_OSS` using OrgManager credentials; main keeps admin unchanged.

## Changes

**`cf_deployment_login`** — picks admin (main) or OrgManager (PR) based on `PR_NUMBER`. Fails fast with a clear error if `AUTOSCALER_ORG_MANAGER_PASSWORD` is unset. All deploy/test/cleanup scripts use this instead of hardcoded `cf_login`.

**Password grant for components** — `eventgenerator`, `scalingengine`, `operator` use `grant_type: password` with OrgManager credentials on PRs. Replaces `client_credentials` which required admin. Extension template wired accordingly.

**Space-scoped broker** — PRs register the service broker `--space-scoped` (OrgManager-compatible). Main uses global registration unchanged.

**`cf_target` simplified** — `find_or_create_{org,space}` removed; org and space are pre-provisioned by the workflow. `cf_target` is now a plain `cf target -o … -s …`.

**Space creation owned by workflow** — the "Create space and assign SpaceDeveloper" step (PR-only, gated with `if: pull_request`) creates the space as org-manager-user so it retains the creator's SpaceDeveloper role. The test runner re-grants SpaceDeveloper at startup to cover the race where another workflow recreates the space between jobs.

**org-manager-user never deleted** — `cleanup-acceptance.sh` and its Makefile targets removed. `cleanup_test_user` removed from `common.sh`. The shared org-manager user is managed out-of-band by the one-off `create-org-manager-user` workflow.

**Security groups moved to infra pipeline** — postgres and metricsgateway security groups bound org-wide once by `deploy-postgres.sh` / `deploy-metricsgateway.sh`. Per-PR security group setup removed from `metricsforwarder/Makefile` and `metricsgateway/Makefile`.

**Shared metricsgateway** — PRs point metricsforwarder at `system-production-metricsgateway` via `METRICSFORWARDER_METRICS_GATEWAY_URL` repo variable, fixing XFCC validation (cert org guid must match the gateway's org).

**`BoolOrInt` in broker tests** — CF returns `plan_updateable` as `bool` for admin, `int` for OrgManager. Unmarshaler now tries bool first, then int.

[...]

PR #1265, opened 2026-06-17
feat: add OAuth2 password grant support for event generator log-cache access

bonzofenix:
## Summary

Adds OAuth2 resource owner password grant support to the CF client and event generator, enabling log-cache access via user credentials instead of client credentials.

### Changes
- `cf/config.go`: Add password grant fields to CF client configuration
- `cf/cfclient_wrapper.go`: Support password grant in CF client wrapper
- `models/uaa_creds.go`: New UaaCreds model for password grant config
- `eventgenerator/metric/cf_oauth2_client.go`: New OAuth2 token client with thundering-herd prevention (sync.Once)
- `eventgenerator/metric/fetcher_factory.go`: Wire password grant into fetcher factory

### Testing
- Full unit test coverage for cf_oauth2_client (490 lines)
- Updated fetcher_factory tests
- Updated CF config tests

---
**Part 1 of 4** — Split from #1149 for easier review.
1. **This PR** — OAuth2 password grant (code)
2. #1266 — Acceptance test OrgManager support (test framework)
3. #1267 — Deployment scripts: shared org + space-scoped broker


eventgenerator/metric/cf_oauth2_client.go
```diff
@@ -0,0 +1,178 @@
...
+// forceRefreshToken refreshes the token after a 401 response.
+// The rejectedToken parameter is the token that was rejected by the server.
+// If another goroutine has already refreshed the token (i.e., the cached token
+// differs from the rejected one), the new token is returned without re-fetching.
+func (c *CFOauth2HTTPClient) forceRefreshToken(rejectedToken string) (string, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// If the token has changed since we got the 401, another goroutine already refreshed it
+	if c.token != rejectedToken {
```
asalan316, line 132: The comment correctly says 

In the race condition path where another goroutine has already refreshed the token, 
we are returning `c.token` without validating it. This could cause issues if:

1. The token is an empty string (would send "Authorization: Bearer ")
2. The token is expired (would cause another 401, potential infinite loop)
3. Another goroutine's refresh failed but mutated c.token to empty

How about ensuring  that do not return an invalid cached token in the race condition scenario

cf/config.go
```diff
@@ -43,9 +50,31 @@ func (conf *Config) Validate() error {
...
+		conf.ClientID = "cf"
+	}
+	return nil
+}
+
+func (conf *Config) validateClientCredentials() error {
+	if conf.GrantType == "" {
+		conf.GrantType = GrantTypeClientCredentials
+	}
 	if conf.ClientID == "" {
```
asalan316, line 76: is empty client secret is allowed here ?

models/uaa_creds.go
```diff
@@ -1,12 +1,21 @@
...
 	ClientID          string `yaml:"client_id" json:"clientID"`
 	ClientSecret      string `yaml:"client_secret" json:"clientSecret"`
+	GrantType         string `yaml:"grant_type" json:"grantType"`
+	Username          string `yaml:"username" json:"username"`
+	Password          string `yaml:"password" json:"password"`
 	SkipSSLValidation bool   `yaml:"skip_ssl_validation" json:"skipSSLValidation"`
 }
 
 func (c UAACreds) IsNotEmpty() bool {
 	return c.URL != ""
```
asalan316, line 16: Is checking url enough ?

How about validation other fields e.g ClientID required, ClientSecret required, Username/Password  ?

they seems to be required here https://github.com/cloudfoundry/app-autoscaler/pull/1265/changes#diff-ddbe55f48c292903fb81e776d43f3fc7c7c1f70f3821626bc6f85c4cffbf4b59R36

eventgenerator/metric/cf_oauth2_client.go
```diff
@@ -0,0 +1,178 @@
...
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("failed to request token: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return "", fmt.Errorf("token request failed with status %d: %s", resp.StatusCode, string(body))
```
asalan316: exposing sensitive information in application logs - uaa response contains access tokens, client id etc

models/uaa_creds.go
```diff
@@ -1,12 +1,21 @@
 package models
 
+const GrantTypePassword = "password"
+
 type UAACreds struct {
 	URL               string `yaml:"url" json:"url"`
 	ClientID          string `yaml:"client_id" json:"clientID"`
 	ClientSecret      string `yaml:"client_secret" json:"clientSecret"`
+	GrantType         string `yaml:"grant_type" json:"grantType"`
```
asalan316, line 9: hard to find new configuration without reading code . Should these new fields

```
 (grant_type, username, password
```

be placed in the default_config.json files e.g

api/default_config.json
eventgenerator/default_config.json
operator/default_config.json
scalingengine/default_config.json

models/uaa_creds.go
```diff
@@ -1,12 +1,21 @@
 package models
 
+const GrantTypePassword = "password"
```
asalan316, line 3: Should we document the usage of 
password grant in evengenerator or appropiate place e.g example configs showing password grant for eventgenerator ?

eventgenerator/metric/cf_oauth2_client.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	var tokenResp tokenResponse
+	if err := json.NewDecoder(resp.Body).Decode(&tokenResp); err != nil {
+		return "", fmt.Errorf("failed to decode token response: %w", err)
+	}
+
+	c.token = tokenResp.AccessToken
+	// Calculate expiration time with 30 second buffer for clock skew
+	c.expiresAt = time.Now().Add(time.Duration(tokenResp.ExpiresIn-30) * time.Second)
```
asalan316: Subtracts 30 seconds without checking if ExpiresIn > 30. 

Short-Lived Tokens (0-29 seconds)

E.g ExpiresIn values 0-29 seconds , this creates negative duration -> token is "already expired"->immediate fetch again→ infinite refresh loop. Negative ExpiresIn

with mutiple concurrent request

- Second 1: 100 requests → 100 see "expired" token → 100 refresh attempts

- Second 2: 100 requests → 100 see "expired" token → 100 refresh attempts 

=> overwhelming UAA, Rate limiting etc

eventgenerator/metric/cf_oauth2_client.go
bonzofenix: Done in [`d31012cbf`](https://github.com/cloudfoundry/app-autoscaler/commit/d31012cbfb0f36bd4392c34fc5a925e2efb27e34). Now validates the cached token before returning in the race condition path — if empty or expired, refreshes instead of returning stale token.

[`forceRefreshToken`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/eventgenerator/metric/cf_oauth2_client.go#L122-L128)

cf/config.go
bonzofenix: Intentionally allowed. CF's built-in `cf` UAA client uses an empty secret by design — UAA expects `cf:` (client_id with no secret) for the password grant flow. This matches `cf login` behavior.

See [CF UAA docs on default clients](https://docs.cloudfoundry.org/uaa/uaa-user-management.html) — the `cf` client is a public client with no secret.

models/uaa_creds.go
bonzofenix: Keeping this as-is. `IsNotEmpty()` is a presence check ("are UAA creds configured at all?"), not a validity check. Callers use it to branch between "UAA configured" vs "skip UAA".

Full field validation (ClientID, Username, Password, etc.) happens in [`cf/config.go#validatePasswordGrant()`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/cf/config.go#L55-L67) and [`validateClientCredentials()`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/cf/config.go#L69-L80) — separate concern from the zero-value check here.

eventgenerator/metric/cf_oauth2_client.go
bonzofenix: Done in [`56fd1c77b`](https://github.com/cloudfoundry/app-autoscaler/commit/56fd1c77b88eb470404ad03b26bfee9ad65947bb). Removed response body from error message — now only logs the HTTP status code.

[`doRefreshToken`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/eventgenerator/metric/cf_oauth2_client.go#L168-L170)

models/uaa_creds.go
bonzofenix: Done in [`d0051363e`](https://github.com/cloudfoundry/app-autoscaler/commit/d0051363e715bbba9f1faad658d8cf8f4186879e). Added UAA section to [`eventgenerator/default_config.json`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/eventgenerator/default_config.json#L44-L52) with all password grant fields.

Only added to eventgenerator since that's the only component using password grant. The other services (api, operator, scalingengine) use client_credentials and their existing configs already cover that.

eventgenerator/metric/cf_oauth2_client.go
bonzofenix: Done in [`d9c74d08c`](https://github.com/cloudfoundry/app-autoscaler/commit/d9c74d08c855b01901d70f12878fe633cb01d04e). Now uses `ExpiresIn / 2` as buffer when `ExpiresIn <= 30`, preventing negative durations and tight refresh loops.

[`doRefreshToken`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/eventgenerator/metric/cf_oauth2_client.go#L177-L185)

models/uaa_creds.go
bonzofenix: Done in [`ae6d25cab`](https://github.com/cloudfoundry/app-autoscaler/commit/ae6d25cab85af24e99dad16f57f1fbf41b82199a). Created [`eventgenerator/README.md`](https://github.com/cloudfoundry/app-autoscaler/blob/split/oauth2-password-grant/eventgenerator/README.md) documenting both authentication modes (client_credentials and password grant) with example YAML configs, required/optional field descriptions, and how the token refresh mechanism works.

asalan316: Glad that the docs related to service is kept in the service folder itself (close to service)

asalan316 (approved):

PR #1266, opened 2026-06-17
feat: add OrgManager user support to acceptance test framework

bonzofenix:
## Summary

Introduces org-manager user concept to acceptance tests, enabling tests to run with limited permissions (OrgManager instead of admin).

### Changes
- `acceptance/config/config.go`: Add OrgManager user config fields
- `acceptance/helpers/cleanup.go`: Per-node teardown to prevent space deletion races
- `acceptance/helpers/org.go`: Org helper for shared org environments
- `acceptance/api/api_test.go`: Accept 403 in deletePolicy for org-manager
- `acceptance/broker/broker_test.go`: Adapt broker tests for org-manager
- `acceptance/assets/app/go_app/internal/app/disk_test.go`: Disk write timing assertion

### Testing
- Tests updated to handle OrgManager permission boundaries
- Per-node teardown prevents parallel test races

---
**Part 2 of 4** — Split from #1149 for easier review.
1. #1265 — OAuth2 password grant (code)
2. **This PR** — Acceptance test OrgManager support (test framework)
3. #1267 — Deployment scripts: shared org + space-scoped broker


acceptance/Makefile
```diff
@@ -20,8 +20,7 @@ export GOWORK := off
 
 .PHONY: clean
 clean:
-	rm --recursive --force './build' './vendor' './ginkgo_v2' './results'
```
asalan316, line 23: Why other folders are skip ?

bonzofenix: Honestly I do not remember why this was changed — looks unintentional. Restored the original clean target in https://github.com/cloudfoundry/app-autoscaler/commit/b822d760d.

acceptance/helpers/spaces.go
```diff
@@ -25,7 +25,11 @@ type Space struct {
 
 func GetSpaceGuid(cfg *config.Config, orgGuid string) string {
 	testSpace := GetTestSpaces(orgGuid, cfg)[0]
-	spaceGuidByte := cf.Cf("space", testSpace, "--guid").Wait(cfg.DefaultTimeoutDuration())
+	return getSpaceGuidByName(testSpace, cfg.DefaultTimeoutDuration())
+}
+
+func getSpaceGuidByName(spaceName string, timeout time.Duration) string {
+	spaceGuidByte := cf.Cf("space", spaceName, "--guid").Wait(timeout)
```
asalan316:
```suggestion
	spaceGuidByte := cf.Cf("space", spaceName, "--guid").Wait(timeout)
	 Expect(spaceGuidCmd).To(Exit(0), fmt.Sprintf("failed to get GUID for space %s: %s", 
            spaceName, string(spaceGuidCmd.Err.Contents())))
    
```

acceptance/helpers/cleanup.go
```diff
@@ -11,27 +11,35 @@ import (
...
-			deleteAllApps(cfg, orgGuid, setup.GetOrganizationName(), GetSpaceGuid(cfg, orgGuid), spaceName)
-			DeleteSpaces(cfg.ExistingOrganization, GetTestSpaces(orgGuid, cfg), cfg.DefaultTimeoutDuration())
+		targetOrgWithSpace(setup.GetOrganizationName(), "", cfg.DefaultTimeoutDuration())
+		orgGuid := GetOrgGuid(cfg, cfg.ExistingOrganization)
+		spaceNames := GetTestSpaces(orgGuid, cfg)
+		if len(spaceNames) == 0 {
+			return
 		}
+
+		for _, spaceName := range spaceNames {
```
asalan316: Observation: 


Potential issue when Ginkgo nodes are greater than 1 - 


Space GUID lookup inside loop could fail (Space not found) if spaces are already deleted concurrently (by another ginkgo node) between name lookup and GUID lookup i.e, multiple test nodes could attempt to delete the same space

acceptance/helpers/spaces.go
bonzofenix: Done in 7bff9ab.

Applied the suggestion: renamed `spaceGuidByte` → `spaceGuidCmd` and added [`Expect(spaceGuidCmd).To(Exit(0), ...)`](https://github.com/cloudfoundry/app-autoscaler/blob/split/acceptance-orgmanager/acceptance/helpers/spaces.go#L32-L34) to fail fast if the CF command errors.

acceptance/helpers/cleanup.go
bonzofenix: Good catch — fixed in 7bff9ab.

Root cause: `GetTestSpaces` returns names only, so the loop called `getSpaceGuidByName` per space — a second CF lookup that could race with another Ginkgo node that already deleted the space.

Fix: replaced the two-step lookup with [`GetRawSpaces`](https://github.com/cloudfoundry/app-autoscaler/blob/split/acceptance-orgmanager/acceptance/helpers/spaces.go#L57-L65) which returns `[]Space{Guid, Name}` from a single `/v3/spaces` curl. The new [`filterTestSpaces`](https://github.com/cloudfoundry/app-autoscaler/blob/split/acceptance-orgmanager/acceptance/helpers/spaces.go#L47-L55) helper filters by name prefix, and [`CleanupInExistingOrg`](https://github.com/cloudfoundry/app-autoscaler/blob/split/acceptance-orgmanager/acceptance/helpers/cleanup.go#L27-L44) now iterates over the already-resolved `[]Space`, passing `space.Guid` directly — no second GUID lookup.

asalan316 (approved):

PR #1267, opened 2026-06-17
feat: support shared org and space-scoped broker for PR deployments

bonzofenix:
## Summary

Enables PR deployments to use a shared existing CF org with space-scoped service brokers, instead of creating a new org per deployment.

### Changes
- `scripts/common.sh`: Add deployment helper functions (cf_target, space-scoped broker logic)
- `scripts/vars.source.sh`: Add EXISTING_ORGANIZATION variable support
- `scripts/build-extension-file.sh`: Pass valid_org_guid for metricsgateway XFCC
- `scripts/mta-deploy.sh`: Target org/space before space-scoped broker creation
- `scripts/enable-service-access.sh`: New — enable service access for space-scoped broker
- `scripts/create-org-manager-user.sh`: Simplified for shared org workflow
- `scripts/extension-file.tpl.yaml`: Add metricsgateway org-guid config
- Various script fixes for shared-org compatibility

### Testing
- Tested via PR acceptance test pipeline

---
**Part 3 of 4** — Split from #1149 for easier review.
1. #1265 — OAuth2 password grant (code)
2. #1266 — Acceptance test OrgManager support (test framework)
3. **This PR** — Deployment scripts: shared org + space-scoped broker


scripts/common.sh
```diff
@@ -115,16 +241,7 @@ function cleanup_apps(){
...
-	local deploy_service_url="https://deploy-service.${system_domain}"
-	local mtas_response
-
-	# Fetch MTAs from deploy-service (--insecure for self-signed certs)
-	if mtas_response="$(curl --silent --fail --insecure --header "Authorization: $(cf oauth-token)" "${deploy_service_url}/api/v2/spaces/${space_guid}/mtas" 2>/dev/null)"; then
-		mtar_app="$(jq -r '.[] | .metadata.id' <<< "${mtas_response}" 2>/dev/null)" || true
-	else
-		echo "Warning: Failed to fetch MTAs from deploy-service, skipping MTA cleanup"
-	fi
+	mtar_app="$(curl --header "Authorization: $(cf oauth-token)" "deploy-service.${system_domain}/api/v2/spaces/${space_guid}/mtas"  | jq ". | .[] | .metadata | .id" -r)"
```
asalan316:
```suggestion
	mtar_app="$(curl --silent --fail --insecure --header "Authorization: $(cf oauth-token)" "https://deploy-service.${system_domain}/api/v2/spaces/${space_guid}/mtas"  | jq ". | .[] | .metadata | .id" -r)"
```

scripts/common.sh
```diff
@@ -40,13 +45,25 @@ function bbl_login() {
...
 }
 
+# Login to CF with appropriate credentials for deployment operations
+# Uses admin on main branch, org-manager on PR branches
+function cf_deployment_login(){
+	step 'login to cf for deployment operations'
+	if [[ "${PR_NUMBER:-main}" == "main" ]]; then
+		cf_admin_login
+	else
+		cf api "https://api.${system_domain}" --skip-ssl-validation
```
asalan316, line 63:
```suggestion
		step "login to cf as ${AUTOSCALER_ORG_MANAGER_USER}"
		cf api "https://api.${system_domain}" --skip-ssl-validation
```

scripts/common.sh
```diff
@@ -71,6 +88,106 @@ function cleanup_service_broker(){
 	fi
 }
 
+# Returns GUIDs of autoscaler service instances (one per line)
+function get_autoscaler_service_instance_guids(){
+	cf curl "/v3/service_instances" | jq -r \
+		'.resources[] | select(.relationships.service_plan.data != null) |
+		select(.name | contains("autoscaler")) | .guid' || true
```
asalan316: Only return instances for a particular PR deployment (not other PRs)

```suggestion
		select(.name | startswith("'"${DEPLOYMENT_NAME}"'")) | .guid' || true
```

scripts/common.sh
```diff
@@ -40,13 +45,25 @@ function bbl_login() {
 	eval "$("${script_dir}/bbl-print-env.sh" "${bbl_state_path}")"
 }
 
-function cf_login(){
-	step 'login to cf'
+function cf_admin_login(){
```
asalan316: The renaming of function will break https://github.com/cloudfoundry/app-autoscaler/blob/main/scripts/cleanup-autoscaler-deployments.sh#L18

scripts/enable-service-access.sh
```diff
@@ -0,0 +1,53 @@
...
+fi
+
+SERVICE_NAME="$1"
+
+echo "${SEPARATOR}"
+echo "Enabling service access (admin)"
+echo "${SEPARATOR}"
+echo "Service: ${SERVICE_NAME}"
+echo ""
+
```
asalan316, line 43: Fail early before actually using the service broker

```suggestion
# Validate service exists in marketplace
echo "Validating service '${SERVICE_NAME}' exists..."
if ! cf service-brokers | grep -q "${SERVICE_NAME}"; then
  echo "ERROR: Service '${SERVICE_NAME}' not found in registered brokers" >&2
  echo "       Available brokers:" >&2
  cf service-brokers >&2
  exit 1
fi

```

scripts/common.sh
```diff
@@ -71,6 +88,106 @@ function cleanup_service_broker(){
...
+		echo " - ${remaining} instances remaining, waiting..."
+		sleep "${poll_interval}"
+	done
+
+	echo " - timeout waiting for service instance deletion"
+	return 1
+}
+
+# Deletes all autoscaler service instances and their bindings
+function delete_autoscaler_service_instances(){
```
asalan316: How about using the common function to remove the autoscaler service instances removal here as well:

https://github.com/cloudfoundry/app-autoscaler/blob/aa05b2dab3e5cfc0ae6e848c97a927071778984a/scripts/common.sh#L254-L270

scripts/common.sh
bonzofenix: you are right this we should not do

scripts/common.sh
bonzofenix: i like this thanks

scripts/common.sh
bonzofenix: Done in [f933e2cf6](https://github.com/cloudfoundry/app-autoscaler/commit/f933e2cf6) — renamed back to `cf_login` across all 6 scripts.

scripts/common.sh
bonzofenix: Removed the unused functions in [889f49a50](https://github.com/cloudfoundry/app-autoscaler/commit/889f49a50) — they were never wired into any caller. The existing inline purge loop in `cleanup_apps` stays as-is.

scripts/common.sh
bonzofenix: The function was removed entirely in [889f49a50](https://github.com/cloudfoundry/app-autoscaler/commit/889f49a50) — it was never wired into any caller so we dropped it rather than fixing the filter.

Makefile
```diff
@@ -382,6 +382,10 @@ cf-login:
 		  'The necessary changes to the environment get lost when make exits its process.'
 	@${MAKEFILE_DIR}/scripts/os-infrastructure-login.sh
 
+.PHONY: register-broker
```
asalan316, line 385: Do we still need ?

asalan316 (approved):
Approved with some minor findings

PR #1277, opened 2026-06-23
chore(deps): bump Spring Boot to 4.1 

silvestre:
# Issue
The used version of [Spring Boot (3.5) goes out of maintenance in a week](https://endoflife.date/spring-boot).

# Fix
Bump Spring Boot to latest 4.1 and adapt codebase to the required change. Change was mainly done by Moderne OpenRewrite and Claude.

### DIFF
--- a/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/CloudFoundryConfigurationProcessor.java
+++ b/scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/CloudFoundryConfigurationProcessor.java
@@ -1,7 +1,5 @@
 package org.cloudfoundry.autoscaler.scheduler.conf;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import java.util.List;
 import java.util.Map;
 import org.slf4j.Logger;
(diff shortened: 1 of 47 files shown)


scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/service/ScheduleManager.java
```diff
@@ -446,39 +445,33 @@ private ApplicationSchedules parseSchedulesFromPolicy(PolicyJsonEntity policyJso
     ObjectMapper mapper = new ObjectMapper();
     String policyJson = policyJsonEntity.getPolicyJson();
     ApplicationSchedules applicationSchedules = null;
-    mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
silvestre, line 449: Removed as it's now the default: https://github.com/FasterXML/jackson/wiki/Jackson-Release-3.0#config-default-changes-1

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/CloudFoundryConfigurationProcessor.java
joergdw, line 8: Unrelated to this PR but we tackle already similar issues here: I just have noted that this has been deprecated. We could try to use the substituter:
```suggestion
import org.springframework.boot.EnvironmentPostProcessor;
```

joergdw (approved):
LGTM

scheduler/src/main/java/org/cloudfoundry/autoscaler/scheduler/conf/CloudFoundryConfigurationProcessor.java
silvestre: Thank you for the suggestion! I'll create a followup PR to tackle deprecations but merge this one as is to unblock renovate :)

PR #1268, opened 2026-06-17
ci: wire shared org deployment into CI/CD workflows

bonzofenix:
## Summary

Wires the shared org / space-scoped broker deployment model into GitHub Actions and Concourse pipelines.

### Changes
- `.github/workflows/acceptance_tests_reusable.yaml`: Add AUTOSCALER_ORG and EXISTING_ORGANIZATION inputs
- `.github/workflows/acceptance_tests_mta.yaml`: Pass org config to reusable workflow
- `.github/workflows/create-org-manager-user.yaml`: Update for shared org
- `ci/infrastructure/pipeline.yml`: Update Concourse pipeline
- `ci/infrastructure/scripts/deploy-metricsgateway.sh`: Add security group binding
- `ci/infrastructure/security-groups/`: Move security group JSON here
- Remove obsolete `metricsforwarder/Makefile` and `metricsgateway/Makefile`

### Testing
- CI pipeline integration test via PR deployment

---
**Part 4 of 4** — Split from #1149 for easier review.
1. #1265 — OAuth2 password grant (code)
2. #1266 — Acceptance test OrgManager support (test framework)
3. #1267 — Deployment scripts: shared org + space-scoped broker
4. **This PR** — CI/CD workflows wiring

PR #1295, opened 2026-06-25
fix: prevent cleanup from deleting shared org

bonzofenix:
## Problem

Multiple PRs deploy into the shared org `SAP_autoscaler_tests_OSS`. The cleanup step in `scripts/common.sh` unconditionally deletes whatever org `AUTOSCALER_ORG` points to — if one PR's cleanup runs, it destroys the shared org and all other running PR deployments lose their space and service broker registrations.

## Root Cause

Traced via CI logs: the `main` branch run's Cleanup step (run 28087413541, June 24 09:40–09:46) deleted the previous org. Then `split/ci-workflows` PR #1268's Deploy step (run 28091226510, 10:15:58) recreated it via `find_or_create_org()`. Any subsequent PR cleanup would delete it again.

## Fix

Guard org deletion with a name check: only delete the org when `AUTOSCALER_ORG == DEPLOYMENT_NAME` (meaning the org was purpose-built for this deployment). When using a shared org, the names differ (`SAP_autoscaler_tests_OSS` vs `autoscaler-mta-{PR}`) and cleanup skips the deletion.

## Verification

| Scenario | DEPLOYMENT_NAME | AUTOSCALER_ORG | Match? | Result |
|----------|----------------|----------------|--------|--------|
| main push | autoscaler-mta-main | autoscaler-mta-main | ✅ | Deletes (correct) |
| PR with shared org | autoscaler-mta-1149 | SAP_autoscaler_tests_OSS | ❌ | Skips (correct) |
| PR without shared org (old) | autoscaler-mta-1266 | autoscaler-mta-1266 | ✅ | Deletes (correct) |


asalan316 (approved):
approving but there are some sonar qube finding

PR #1320, opened 2026-07-20
Update a few dependencies;

joergdw:
Bumps a few dependencies to more recent versions.


silvestre (approved):

.github/workflows/acceptance_tests_reusable.yaml
joergdw, line 1: Unrelated – but tests currently have a bad success-rate from ca. 20%.

PR #1322, opened 2026-07-21
fix(mta): make custom builders self-sufficient for standalone mbt build

silvestre:
# Issue

The MTA descriptor's custom builders relied on prerequisites that only run when mbt is invoked at the end of `make mta-build` (go-mod-vendor-mta, vendor-changelogs). Invoking mbt directly (e.g. from a CI that calls `mbt build` on its own) skips those, producing an MTAR that fails to deploy.

The acceptance-tests module builds via `make acceptance-release`, which assumed a devbox-provisioned environment. Two build-time assumptions prevented it from running in a minimal build image (e.g. the mbt CI image) that has the go toolchain but not ginkgo or the cf CLI.

# Fix

Fold those prerequisites into the builders themselves so `mbt build` is self-sufficient regardless of caller, and drop them from the mta-build / mta-release make targets (they now run per-module inside mbt).

Also give the Java modules (dbtasks, scheduler) their own submodule path so `mbt sbom-gen` aggregates each pom alone instead of the root reactor twice, which otherwise merges two identical BOMs and fails CycloneDX uniqueItems validation.

compile-acceptance-tests.sh invoked a pre-installed `ginkgo` on PATH: Build a host-native ginkgo from the version pinned in acceptance/go.mod instead, then use it to cross-compile each suite via GOOS/GOARCH. Needs only `go`.

AUTOSCALER_PLUGIN_VERSION was derived from `cf plugins --checksum`, requiring a running cf with the plugin installed: Resolve it from nix/packages.nix — the same single source of truth the devbox env builds the plugin from — using awk.

Both keep the existing devbox behavior unchanged while letting the acceptance release build in any environment with the go toolchain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,16 @@ DEST ?= /tmp/build
 TARGET_DIR ?= ./build
 MTAR_FILENAME ?= app-autoscaler-release-v$(VERSION).mtar
 ACCEPTANCE_TESTS_FILE ?= ${DEST}/app-autoscaler-acceptance-tests-v$(VERSION).tgz
-AUTOSCALER_PLUGIN_VERSION ?= $(shell cf plugins --checksum 2>/dev/null | awk '/^AutoScaler/ {print $$2}')
+# Resolve the AutoScaler CLI plugin version from its single source of truth,
+# nix/packages.nix (the same pin the devbox env builds from), so the acceptance
+# release can be built in any environment with just awk — no running `cf` or a
+# pre-installed plugin required. Override by exporting AUTOSCALER_PLUGIN_VERSION.
+AUTOSCALER_PLUGIN_VERSION ?= $(shell awk '\
+	/app-autoscaler-cli-plugin = buildGoModule/{b=1} \
+	b && /major *=/{gsub(/[^0-9]/,"");maj=$$0} \
+	b && /minor *=/{gsub(/[^0-9]/,"");min=$$0} \
+	b && /patch *=/{gsub(/[^0-9]/,"");pat=$$0;print maj"."min"."pat;exit}' \
+	$(MAKEFILE_DIR)/nix/packages.nix)
 CI ?= false
 
 DEBUG := false
@@ -309,7 +318,10 @@ mta-logs:
 # the other dependencies. Instead we should rely on the dependencies being complete and updated
 # as needed.
 .PHONY: mta-build
-mta-build: mta-build-clean go-mod-vendor-mta vendor-changelogs
+# go-mod-vendor-mta and vendor-changelogs are invoked per-module by the custom
+# builders in mta.tpl.yaml, so `mbt build` (via mta-build.sh) is self-sufficient
+# and we don't run them here as well.
+mta-build: mta-build-clean
 	@$(MAKEFILE_DIR)/scripts/mta-build.sh
 
 .PHONY: mta-build-clean
@@ -366,7 +378,7 @@ acceptance-release: clean-acceptance generate-openapi-generated-clients-and-serv
 
 
 .PHONY: mta-release
-mta-release: generate-fakes mta-build
+mta-release: mta-build
 	@echo " - building mtar release '${VERSION}' to dir: '${DEST}' "
 
 .PHONY: clean-acceptance
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,15 @@ DEST ?= /tmp/build
 TARGET_DIR ?= ./build
 MTAR_FILENAME ?= app-autoscaler-release-v$(VERSION).mtar
 ACCEPTANCE_TESTS_FILE ?= ${DEST}/app-autoscaler-acceptance-tests-v$(VERSION).tgz
+
 # Resolve the AutoScaler CLI plugin version from its single source of truth,
 # nix/packages.nix (the same pin the devbox env builds from), so the acceptance
 # release can be built in any environment with just awk — no running `cf` or a
 # pre-installed plugin required. Override by exporting AUTOSCALER_PLUGIN_VERSION.
+#
+# In the future we will replace it with something simpler:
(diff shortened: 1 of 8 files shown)


scripts/compile-acceptance-tests.sh
```diff
@@ -13,12 +13,31 @@ read -ra SUITES <<< "$SUITES"
...
+
+# Build a host-native ginkgo tool from the version pinned in acceptance/go.mod
+# into the given path. Using `go build` (rather than a pre-installed `ginkgo` on
+# PATH) lets this run in any build image that has the go toolchain — e.g. the mbt
+# CI image — without ginkgo installed separately. The tool must be built for the
+# HOST platform (no GOOS/GOARCH override) so it can execute here; it cross-compiles
+# each suite for the target platform via the GOOS/GOARCH env passed to it.
+build_host_ginkgo() {
+  local out="${1}"
+  go build -C acceptance -o "${out}" github.com/onsi/ginkgo/v2/ginkgo
```
silvestre: Reuse ginkgo build from line 49

silvestre: Done in 5d6143b4 — extracted a shared `build_ginkgo` helper that both the host-native build and `compile_ginkgo` now call, so the `go build ... ginkgo` invocation lives in one place.

Makefile
```diff
@@ -8,7 +8,16 @@ DEST ?= /tmp/build
...
+# Resolve the AutoScaler CLI plugin version from its single source of truth,
+# nix/packages.nix (the same pin the devbox env builds from), so the acceptance
+# release can be built in any environment with just awk — no running `cf` or a
+# pre-installed plugin required. Override by exporting AUTOSCALER_PLUGIN_VERSION.
+AUTOSCALER_PLUGIN_VERSION ?= $(shell awk '\
+	/app-autoscaler-cli-plugin = buildGoModule/{b=1} \
+	b && /major *=/{gsub(/[^0-9]/,"");maj=$$0} \
+	b && /minor *=/{gsub(/[^0-9]/,"");min=$$0} \
+	b && /patch *=/{gsub(/[^0-9]/,"");pat=$$0;print maj"."min"."pat;exit}' \
+	$(MAKEFILE_DIR)/nix/packages.nix)
```
joergdw, line 25: We could ask that nix directly:

```suggestion
AUTOSCALER_PLUGIN_VERSION ?= $(shell \
	nix eval --raw '$(MAKEFILE_DIR)#app-autoscaler-cli-plugin.version')
```

scripts/compile-acceptance-tests.sh
```diff
@@ -32,14 +50,21 @@ compile_ginkgo() {
...
     done
   done
 }
 
 main() {
-  compile_suites
+  # Build the ginkgo tool once for the host platform into a temp dir, then use it
+  # to cross-compile the suites. Absolute path so it works after `cd acceptance`.
+  local ginkgo_bin
+  GINKGO_TMPDIR="$(mktemp -d)"
```
joergdw: I suggest to generally use long-forms of options in scripts, where speed matters less than readability. Sometimes macOS primeval times tools make this a bit annoying but in this case devbox is our saviour and already contains the modern version.
```suggestion
  GINKGO_TMPDIR="$(mktemp --directory)"
```

scripts/compile-acceptance-tests.sh
```diff
@@ -32,14 +50,21 @@ compile_ginkgo() {
...
+  ginkgo_bin="${GINKGO_TMPDIR}/ginkgo"
+  build_ginkgo "${ginkgo_bin}"
```
joergdw: Sometimes this as well gets called in a devbox-environment e.g. locally: `VERSION=1.1.1 make clean-build create-assets`. Shouldn't we avoid it in these cases and guard this ginkgo-building by checking first if ginkgo already has been provided in the environment?

mta.tpl.yaml
joergdw, line 1: I am unsure if this repository even needs these additional instructions. By design, we use make in the top-most layer (in the sense of [tiers](<https://en.wikipedia.org/wiki/Multitier_architecture>) to build everything. Once we leave it, we never want to come back.

I would propose to rather eliminate everything that is not needed for mbt to produce the valid mtar.

Makefile
joergdw, line 343: As this PR is about getting the mtar self-contained, I propose to segregate the mta.tpl.yaml and the Makefile. Currently it looks as both are responsible for mtar-building. I wonder if then it isn't a source of confusion. What do you think about:

```suggestion
mta-build:
	@echo -e \
		'🚸 Responsible for the build of mtars is not GNUmake but mbt.' "\n" \
		'mbt calls make to prepare its stage.' "\n" \
		'Calling mbt via this Makefile is not intended and discouraged to keep responsibilities separated.' "\n" \
		'You can invoke mbt by executing:' "\n" \
		"\t" 'cp mta.tpl.yaml mta.yaml' "\n" \ 
		"\t" 'sed --in-place "s/MTA_VERSION/${VERSION}/g" mta.yaml' "\n" \
		"\t" 'mbt build --target="<target-folder>" --mtar="<target-file-name>"'
```

Apart from removing `mta-release` a few lines below, this would of course require at least the following file to be adapted as well: ./README.md and .github/workflows/acceptance_tests_reusable.yaml

joergdw (approved):
This PR looks fine and functional.

I propose one alternative that follows a different approach. Instead it again keeps the Makefile as the single entry-point:
 + PR #1330

scripts/compile-acceptance-tests.sh
joergdw: Outside this PR, in the future, we can use ginkgo from pkgsStatic for Linux and then it can be provided by Nix for inclusion into the mtar.

silvestre (commented):
LGTM

mta.tpl.yaml
joergdw: We have agreed to do that in a dedicated story.

PR #1323, opened 2026-07-22
chore(CI): remove scale-to-one-az ops file to resolve smoke test failures on CF v58

asalan316:
## Problem

- Smoke tests consistently failed starting with CF v58.0.0
- Tests worked correctly on v57.4.0 and earlier
- Issue caused by breaking changes in CF v58 major version update

## Solution
Remove `operations/cf/scale-to-one-az.yml` from the OPS_FILES list in the CI pipeline configuration

## Testing
- Smoke tests pass on CF v58.0.0+
- Deployment pipeline completes successfully

### Possible Follow up:  
Investigate if scaling down other VMS e.g nats, router etc can lead to successful cf deployment


silvestre (approved):
Expensive, but a stop gap measure until we can retire this foundation.

PR #1354, opened 2026-08-04
fix(acceptance): set autoscaling-api endpoint before querying in DebugInfo

silvestre:
# Issue

Occasionally we instead of debug output on app suite failures we saw `No AutoScaler api endpoint set.` instead, see #1355.

`DebugInfo` launched all 15 `cf` commands concurrently; `cf autoscaling-api <endpoint>` writes the CF plugin's local config file while `autoscaling-history` and `autoscaling-metrics` commands were already running and reading it creating a race condition.

# Fix
Run commands sequentially.

Fixes: #1355

### DIFF
--- a/acceptance/helpers/debug.go
+++ b/acceptance/helpers/debug.go
@@ -14,56 +14,51 @@ import (
 )
 
 func DebugInfo(cfg *config.Config, setup *workflowhelpers.ReproducibleTestSuiteSetup, anApp string) {
-	if os.Getenv("DEBUG") != "" && cfg.ASApiEndpoint != "" {
-		if os.Getenv("CF_PLUGIN_HOME") == "" {
-			_ = os.Setenv("CF_PLUGIN_HOME", os.Getenv("HOME"))
-		}
-		output := new(strings.Builder)
-		_, _ = fmt.Fprintf(output, "\n=============== DEBUG ===============\n")
-
-		// autoscaling-api writes the plugin's local config file that the other
-		// autoscaling-* commands read, so it must complete before they start.
-		waitAndPrint(startCommand("cf", "autoscaling-api", cfg.ASApiEndpoint), output)
-
-		var commands []*Session
-		commands = append(commands, startCommand("cf", "app", anApp))
-		commands = append(commands, startCommand("cf", "events", anApp))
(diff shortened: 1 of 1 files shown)


acceptance/helpers/debug.go
```diff
@@ -34,19 +40,23 @@ func DebugInfo(cfg *config.Config, setup *workflowhelpers.ReproducibleTestSuiteS
 		commands = append(commands, command("cf", "autoscaling-metrics", anApp, "disk"))
 		commands = append(commands, command("cf", "autoscaling-metrics", anApp, "diskutil"))
 		commands = append(commands, command("cf", "autoscaling-metrics", anApp, "test_metric"))
-		output := new(strings.Builder)
-		_, _ = fmt.Fprintf(output, "\n=============== DEBUG ===============\n")
 		for _, command := range commands {
-			command.Wait(30 * time.Second)
```
geigerj0, line 40: > DebugInfo launched all 15 cf commands concurrently

how come all these commands were executed concurrently even though `command.Wait(30 * time.Second)` waits until the command exists and then proceeds with the next one? Am I missing something?

silvestre: Yes, does 3be7d857d7ec6ed89f97f6b324d2099506cd9337 help? 😅

geigerj0: I see, thank you, way better! however, why are we executing this all at once? 

is this really necessary? linear-control flows are way easier to understand 🤓 

if concurrency is not really required, how about making everything linear?

silvestre: Hmm, it is called from https://github.com/cloudfoundry/app-autoscaler/blob/67864c1cb03636f4a6a816ad03a55c1d6015d362/acceptance/app/app_suite_test.go#L85-L100 which is registered in AfterEach, whenever there's an app in the acceptance tests.

I'll put a commit on this PR removing the parallelism - let's check if becomes measurably slower.

silvestre: 3238eb1dfa7ebce7f0c8b8d21f88bce782e084fc had no statistically significant difference in runtime.

geigerj0 (approved):

PR #1395, opened 2026-08-21
chore(renovate): increase priority for Spring (Boot) updates

silvestre:
Spring updates often resolve many CVEs in the transitive dependency tree. 

This `packageRule` currently would match `org.springframework.boot` as the parent POM for the scheduler and increase the PR prio, see also https://docs.renovatebot.com/configuration-options/#packagerulesprpriority

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -89,6 +89,12 @@
       ]
     },
     {
+      "description": "Prioritize Spring updates over others",
+      "matchManagers": ["maven"],
+      "matchPackageNames": ["/springframework/"],
+      "prPriority": 10
+    },
+    {
       "description": "Disable Liquibase updates due to Functional Source License (FSL)",
       "groupName": "Liquibase",
       "matchManagers": [


renovate.json
```diff
@@ -88,6 +88,13 @@
         "/loggregator-agent-release/"
       ]
     },
+    {
+      "description": "Remove concurrent PR cap for Spring Boot updates",
+      "matchManagers": ["maven"],
+      "matchPackageNames": ["/springframework/"],
+      "prPriority": 10,
```
geigerj0: isn't the concern of the PR rather "always do spring updates first, regardless of the concurrent PR limit" than just "remove concurrent PR limit for spring updates"?

renovate.json
```diff
@@ -88,6 +88,13 @@
...
+      "prConcurrentLimit": 0
```
geigerj0: i only find `prConcurrentLimit ` on repo-config-level but not for `packageRules`-level: https://docs.renovatebot.com/configuration-options/#packagerules.

does setting `prConcurrentLimit` here work?

silvestre: No, it seems not to work, I trusted the validator https://github.com/cloudfoundry/app-autoscaler/actions/runs/32464441372/job/96717962253?pr=1395 but it probably just ignores extra fields.

Thanks for catching this!

I suppose just increasing the prio is enough for now.

renovate.json
silvestre: I reworded the description in the file and the PR.

renovate.json
geigerj0: even though this validator seems to be recommended to use over the JSON-schema: https://docs.renovatebot.com/json-schema/#limitations

I still doubt that it covers everything. for example, `prPriority` is completely missing but `allowedVersions` is present: https://github.com/renovatebot/renovate/blob/main/lib/config/validation.ts

geigerj0 (approved):

PR #1401, opened 2026-08-28
refactor(api): replace go-chi with stdlib http.ServeMux

silvestre:
# Issue

go-chi was used, even though stdlib `ServeMux` is a drop-in replacement. The broker router is only ever consumed as a plain `http.Handler` and uses no chi-specific features (URL params, middleware chains, sub-routers).

# Fix

Replaced go-chi with stdlib.


geigerj0 (approved):
good catch finding this!

PR #1423, opened 2026-09-02
🚀 Speedup by grouping dependency-updates into fewer PRs

joergdw:
This change groups all dependencies together that are unlikely to have an impact on the outcome of the acceptance-tests. So stable clis, ide-tools and the relatively stable linter shellcheck.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -106,7 +106,7 @@
       "enabled": false
     },
     {
-      "description": "Group together clis that are likely to have no impact on the result of the acceptance-tests. This is for speed-increase and to reduce the traffic-jam in the renovate-dependecy-dashboard <https://github.com/cloudfoundry/app-autoscaler/issues/832>.",
+      "description": "Group together dependencies that are likely to have no impact on the result of the acceptance-tests. This is for speed-increase and to reduce the traffic-jam in the renovate-dependecy-dashboard <https://github.com/cloudfoundry/app-autoscaler/issues/832>.",
       "groupName": "Low risk cli-updates",
       "matchManagers": ["devbox"],
       "prPriority": -1,
@@ -122,9 +122,11 @@
         "gnused",
         "gnutar",
         "gopls",
+        "gum",
         "jdt-language-server",
         "jq",
         "maven",
+        "oha",
         "openssl",
         "python",
         "ripgrep",
--- a/renovate.json
+++ b/renovate.json
@@ -104,6 +104,35 @@
         "/^org\\.liquibase:.*/"
       ],
       "enabled": false
+    },
+    {
+      "description": "Group together clis that are likely to have no impact on the result of the acceptance-tests. This is for speed-increase and to reduce the traffic-jam in the renovate-dependecy-dashboard <https://github.com/cloudfoundry/app-autoscaler/issues/832>.",
+      "groupName": "Low risk cli-updates",
+      "matchManagers": ["devbox"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -104,6 +104,37 @@
         "/^org\\.liquibase:.*/"
       ],
       "enabled": false
+    },
```
geigerj0, line 107: what does this mean? 🤷‍♂️ 

Image

renovate.json
```diff
@@ -104,6 +104,37 @@
...
+    {
+      "description": "Group together dependencies that are likely to have no impact on the result of the acceptance-tests. This is for speed-increase and to reduce the traffic-jam in the renovate-dependecy-dashboard <https://github.com/cloudfoundry/app-autoscaler/issues/832>.",
+      "groupName": "Low risk cli-updates",
```
geigerj0, line 110: i really like the idea of grouping updates to avoid traffic jam. however, I am asking myself the question if this grouping will really fix our problem of "traffic jam" in the dependency updates. these tools don't receive that much updates so I am wondering if it is worth it making the renovate config more complex.
here you can see that there are not much updates coming in throughout the past months for these tools: https://github.com/cloudfoundry/app-autoscaler/commits/main/devbox.json

renovate.json
joergdw: I have removed that. This came in by pure accident: The IDE adds hints into git-commmit-message-comments for the author. By removing accidentally a line-break, git was not able anymore to recognise them as comments that should not be part of the commit-message.

renovate.json
joergdw: If we have a look at the dashboard
renovate-Dashboard
we see with "oha" and "gum" currently two entries. Additionally we see 
Open renovate-dep-PRs
some PRs for other clis that have already been opened.

I personally would consider this as a gain even if those situations appear once every few months or something like that. Given how much PRs get stuck due to how bad acceptance-tests run.

geigerj0: yes, because so many other dep bumps have piled up. but if you look at devbox.json, there is not much flowing in general.

it looks to me like another list that needs to be kept "manually" in sync (devbox.json <-> renovate config). 

maybe it is more worth it to _somehow_ tackle the root cause of all this: renovate rate limit

if you still think that it is worth it, feel free to merge, up to you

geigerj0 (approved):

renovate.json
joergdw: @silvestre What do you think?

Personally I don't think that his is much complication compared to what we already have. On the other side it is clearly a workaround (instead of a true problem solving) and another thing to maintain.

silvestre: I agree with @geigerj0's points, and it is a workaround, but it's low risk and easy to rollback. Worst case in case it's not maintained is that it becomes less effective. Let's try it.

geigerj0: > tackle the root cause of all this: renovate rate limit

an idea just sparked in my head: how about running renovate "self-hosted" via GHA https://github.com/renovatebot/github-action

have we every tried this out? this way, we could get around the rate-limiting issue

silvestre: The ratelimit is configured by us to not overwhelm our CI environment. If costs were no issue we could just remove the rate limit and scale up the CI environment to cope with the load.

geigerj0: got it

PR #1466, opened 2026-09-15
Remove `alex`-support

geigerj0:
# Problem
Stumbled over a configuration file for a tool ([alex](https://github.com/get-alex/alex#configuration)) that nobody actually uses.

# Solution
* Let's keep the amount of config-files small and remove it from the repo.
* Remove linter workflow with custom alex-config


silvestre (commented):
I would rather remove the hardcoded `alex` conf from the linter workflow: https://github.com/cloudfoundry/app-autoscaler/blob/a557fcd0fa602058035b1f0a20685fbefbf65b33/.github/workflows/linters.yaml#L40-L55

geigerj0:
> I would rather remove the hardcoded `alex` conf from the linter workflow:
> 
> https://github.com/cloudfoundry/app-autoscaler/blob/a557fcd0fa602058035b1f0a20685fbefbf65b33/.github/workflows/linters.yaml#L40-L55

good point! did not know that there is another linter with another out-of-sync config file 🙈, thanks!

.github/workflows/linters.yaml
```diff
@@ -35,29 +35,6 @@ jobs:
         uses: reviewdog/action-actionlint@8b682e1e7e512151d73a6b9449cbff173846226e # v1
         with:
           reporter: github-pr-review
-      - name: Create alex config
```
geigerj0, line 38: green linter run: https://github.com/cloudfoundry/app-autoscaler/actions/runs/34946381261/job/104306664726?pr=1466

silvestre (approved):

